PR #1347, opened 2022-11-30
Bug fix with `-Upgrade` in `./Add_Single_SRD.ps1`

callummole:
The code was unable to find an appropriately named and existing VM.

This is because `"00-00-" -match "\d-\d-"` returns false due to only looking for one digit.

Amended the VM string matching to use `"\d{2}-\d{2}-\d{10}"`.


jemrobinson (approved):
LGTM

PR #1348, opened 2022-12-01
Add script for automatically adding email authentication

callummole:

Azure AD Connect adds phone numbers to authentication methods, but not emails. We do not always have phone numbers for users. The script `SHM_Add_AAD_User_Emaile_Authentication.ps1` adds an email authentication method for any users that: 1) have OnPremisesSyncEnabled, 2) have an email, 3) do not already have an email authentication method.
z

### :microscope: Tests

The script has been successfully used to add email authentication methods..

### DIFF
--- a/deployment/administration/SHM_Add_AAD_User_Email_Authentication.ps1
+++ b/deployment/administration/SHM_Add_AAD_User_Email_Authentication.ps1
@@ -27,7 +27,7 @@ if (Get-MgContext) {
 # Find all users without with OnPremisesSyncEnabled=True, with Mail entries, who have not already got an authenticationemailmethod.
 # This indicates that they originated from a local AD.
 # ----------------------------------------------------------------------------------
-$users = Get-MgUser -Property Mail,DisplayName,OnPremisesSyncEnabled,UserPrincipalName | where { $_.OnPremisesSyncEnabled } | where { $_.Mail }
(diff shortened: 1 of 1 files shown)


JimMadge:
> Azure AD Connect adds phone numbers to authentication methods, but not emails.

I think this is intentional. Control of both an email address and phone number gives confidence about who owns an account during registration. We don't, in general, have a way to verify who owns an account (like some in person registration or ID check).

@jemrobinson ?

jemrobinson:
This is a tough one. There's a problem for this DSG because we don't (yet) have phone numbers for most of the participants. 

However, @callummole and I talked yesterday about the idea that we *do* want users to demonstrate ownership of *both* an email address and a phone number before they set up their accounts.

Maybe the solution is to a) make this clear in our documentation b) for this DSG only, run this script to allow users to set up their accounts with email only? This would imply not merging this PR.

Thoughts?

martintoreilly:
I think I'm with @JimMadge on this one. Currently we require a user to have access to both their email account and phone number and it to me it feels like a reduction in security / assurance to reduce that to just the email account.

As I understand it, we were prompted to consider this because a cohort of DSG users for the December DSG have not supplied phone numbers. I strongly think we should not let that drive / influence our decision on this issue.

PR #1375, opened 2023-02-01
Add firewall to Pulumi

### LINKED ISSUE #1374: Pulumi: SHM firewall
Add SHM firewall to pulumi

### LINKED ISSUE #1376: Pulumi: SHM route table
Add Pulumi SHM route table

### LINKED ISSUE #1384: Pulumi: Missing domain admin password from keyvault
The domain admin password is not currently added to the keyvault, so it isn't possible to log-in to the domain controller and set up Azure AD Connect.

jemrobinson:

- [x] You are targeting the ~**develop**~ **python-migration** branch.
- [x] Your branch is up-to-date with the **target branch** (you probably started your branch from the target but it may have changed since then).

- Add Azure firewall and associated rules to Pulumi
- Add route table to redirect traffic to firewall
- Add private DNS zones to allow access to automation account

Closes #1374 . Closes #1376. Closes #1384.

### :microscope: Tests

Tested on a local deployment


data_safe_haven/pulumi/components/shm_firewall.py
```diff
@@ -0,0 +1,1128 @@
...
+            sku=network.PublicIPAddressSkuArgs(name="Standard"),
+            opts=opts,
+        )
+
+        # Get firewall subnet
+        snet_firewall = network.get_subnet_output(
+            subnet_name=props.subnet_firewall_name,
+            resource_group_name=props.resource_group_name,
+            virtual_network_name=props.subnet_firewall_virtual_network_name,
+        )
```
JimMadge: This is saying, 'find me a subnet with this name, belonging to this virtual network, in this resource group'.

I assume this is a resource in the same stack?

We could pass the Python object about, but maybe this way keeps the code tidier when there are many resources?

data_safe_haven/pulumi/components/shm_firewall.py
```diff
@@ -0,0 +1,1128 @@
...
+                                    port=443,
+                                    protocol_type="Https",
+                                )
+                            ],
+                            source_addresses=[
+                                "*"
+                                # "{{network.vnetRepositoriesTier2.subnets.mirrorsExternal.cidr}}",
+                                # "{{network.vnetRepositoriesTier2.subnets.proxies.cidr}}",
+                                # "{{network.vnetRepositoriesTier3.subnets.mirrorsExternal.cidr}}",
+                                # "{{network.vnetRepositoriesTier3.subnets.proxies.cidr}}"
```
JimMadge, line 972: These will exist eventually but not yet?

data_safe_haven/pulumi/components/shm_firewall.py
```diff
@@ -0,0 +1,1128 @@
...
+        )
+
+        # Route all connected traffic through the firewall
+        route = network.Route(
+            f"{self._name}_via_firewall",
+            address_prefix="0.0.0.0/0",
+            # next_hop_ip_address=(firewall.ip_configurations)[0].private_ip_address,
+            next_hop_ip_address=firewall.ip_configurations.apply(
+                lambda cfgs: cfgs[0].private_ip_address
+            ),
```
JimMadge:
```suggestion
            next_hop_ip_address=firewall.ip_configurations.apply(
                lambda cfgs: cfgs[0].private_ip_address
            ),
```

data_safe_haven/pulumi/components/shm_firewall.py
jemrobinson: Yes - this is so we don't forget to add them once they exist :)

data_safe_haven/pulumi/components/shm_firewall.py
jemrobinson: Yes. You can either create a `Subnet` as its own resource or as a subresource inside a `VirtualNetwork`. In order to stop the subnets being destroyed and recreated on a re-run we currently do the second one, which means there isn't a `Subnet` object to pass around. I've thought of a way we might be able to do the first one though - if it works, I'll update this logic.

JimMadge: I see, I think I might have come across this before too.

If I remember correctly the vnet object has an attribute containing its subnets (although that might be Terraform).

JimMadge: I do that [here](https://github.com/alan-turing-institute/bureau/blob/main/build/bureaubuild/infrastructure.py)

```python
    vnet = network.VirtualNetwork(
        'vnet',
        virtual_network_name='vnet',
        resource_group_name=resource_group.name,
        address_space=network.AddressSpaceArgs(
            address_prefixes=['10.0.0.0/16'],
        ),
        subnets=[network.SubnetArgs(
            name='default',
            address_prefix='10.0.0.0/24'
        )]
    )
```

and later

```python
        'focal': BuildVMArgs(
            resource_group=resource_group,
            image_reference=compute.ImageReferenceArgs(
                offer='0001-com-ubuntu-server-focal',
                publisher='canonical',
                sku='20_04-lts-gen2',
                version='latest',
            ),
            subnet=vnet.subnets[0],
            private_key=private_key
        ),
```

then in `BuildVM.init`

```python
        nic = network.NetworkInterface(
            f'{name}_nic',
            resource_group_name=args.resource_group.name,
            location=args.resource_group.location,
            ip_configurations=[
                network.NetworkInterfaceIPConfigurationArgs(
                    name='buildvmipconfig',
                    subnet=network.SubnetArgs(id=args.subnet.id),
                    public_ip_address=network.PublicIPAddressArgs(
                        id=public_ip.id
                    )
                )
            ],
            opts=child_opts,
        )
```

So it looks like `vnet.subnets` is a list of `pulumi_azure_native.network.subnet` rather than `SubnetArgs`.

jemrobinson: Yes, but they're ordered rather than named. I suppose we could match up the ordering the named return values?

JimMadge: Oh, yes that does make it a bit tricker.

Could do something like

```python
def get_subnet(subnets: [subnet], name: str) -> Optional(subnet):
    for subnet in subnets:
        if subnet.name == name:
            return subnet
    return None
```

jemrobinson: Done by running `get_subnet` and adding the `GetSubnetResult` as an output from the component.

jemrobinson:
I think I've addressed all of @JimMadge's comments. I've checked that creation/teardown of SHM and SRE work as expected. This involved adding a few additional cross-referenced dependencies to ensure that resources are created in the correct order. There are also some general type-hinting fixes that I added while debugging.

data_safe_haven/commands/deploy_shm_command.py
```diff
@@ -115,7 +121,7 @@ def handle(self) -> None:
     def add_missing_values(self, config: Config) -> None:
         """Request any missing config values and add them to the config"""
         # Request FQDN if not provided
-        fqdn = self.option("fqdn")
+        fqdn = cast(str, self.option("fqdn"))
```
JimMadge, line 124: Does this just trick the type checker that the result of `self.option("fqdn")` is always a string?

What if it is empty?
What if it isn't a string?

data_safe_haven/external/api/graph_api.py
```diff
@@ -535,7 +540,7 @@ def delete_application(
                 f"Could not delete application '{application_name}'.\n{str(exc)}"
             ) from exc
 
-    def get_application_by_name(self, application_name: str) -> JSONType:
+    def get_application_by_name(self, application_name: str) -> Dict[str, Any] | None:
```
JimMadge, line 543:
```suggestion
    def get_application_by_name(self, application_name: str) -> Optional[Dict[str, Any]]:
```

data_safe_haven/external/api/graph_api.py
```diff
@@ -555,13 +562,16 @@ def get_service_principal_by_name(self, service_principal_name: str) -> JSONType
         except (DataSafeHavenMicrosoftGraphException, IndexError):
             return None
 
-    def get_id_from_application_name(self, application_name: str) -> str:
+    def get_id_from_application_name(self, application_name: str) -> str | None:
```
JimMadge, line 565:
```suggestion
    def get_id_from_application_name(self, application_name: str) -> Optional[str]:
```

data_safe_haven/external/api/graph_api.py
```diff
@@ -555,13 +562,16 @@ def get_service_principal_by_name(self, service_principal_name: str) -> JSONType
...
-            return self.get_application_by_name(application_name)["appId"]
+            application = self.get_application_by_name(application_name)
+            if not application:
+                return None
+            return application["appId"]
         except DataSafeHavenMicrosoftGraphException:
             return None
 
-    def get_id_from_groupname(self, group_name: str) -> str:
+    def get_id_from_groupname(self, group_name: str) -> str | None:
```
JimMadge, line 574:
```suggestion
    def get_id_from_groupname(self, group_name: str) -> Optional[str]:
```

data_safe_haven/external/api/graph_api.py
```diff
@@ -571,7 +581,7 @@ def get_id_from_groupname(self, group_name: str) -> str:
         except (DataSafeHavenMicrosoftGraphException, IndexError):
             return None
 
-    def get_id_from_username(self, username: str) -> str:
+    def get_id_from_username(self, username: str) -> str | None:
```
JimMadge, line 584:
```suggestion
    def get_id_from_username(self, username: str) -> Optional[str]:
```

data_safe_haven/external/api/graph_api.py
```diff
@@ -664,7 +686,7 @@ def http_post(self, url: str, **kwargs: Any) -> requests.Response:
                 f"Could not execute POST request.\n{str(exc)}"
             ) from exc
 
-    def read_applications(self) -> JSONType:
+    def read_applications(self) -> Sequence[Dict[str, Any]]:
```
JimMadge, line 689: Do we really not know what concrete collection is returned here?

data_safe_haven/external/interface/azure_ipv4_range.py
```diff
@@ -1,24 +1,56 @@
...
+        ip_address_first = self[0]
+        while True:
+            ip_address_last = ip_address_first + int(number_of_addresses - 1)
+            with suppress(DataSafeHavenIPRangeException):
+                candidate = AzureIPv4Range(ip_address_first, ip_address_last)
+                if not any(subnet.overlaps(candidate) for subnet in self.subnets):
+                    self.subnets.append(candidate)
+                    break
+            ip_address_first = ip_address_first + number_of_addresses
+        return candidate
```
JimMadge, line 56: This feels like an Advent of Code problem :smile:.

data_safe_haven/pulumi/components/virtual_machine.py
```diff
@@ -36,15 +36,17 @@ def __init__(
...
         self.vm_name = vm_name
-        self.vm_name_underscored = vm_name.replace("-", "_")
+        self.vm_name_underscored = Output.from_input(vm_name).apply(
+            lambda n: n.replace("-", "_")
+        )
         self.vm_size = vm_size
 
     @property
-    def os_profile(self) -> compute.OSProfileArgs:
+    def os_profile(self) -> compute.OSProfileArgs | None:
```
JimMadge, line 45:
```suggestion
    def os_profile(self) -> Optional[compute.OSProfileArgs]:
```

data_safe_haven/pulumi/components/virtual_machine.py
```diff
@@ -36,15 +36,17 @@ def __init__(
...
         return self.os_profile_args
 
     @property
-    def image_reference(self) -> compute.ImageReferenceArgs:
+    def image_reference(self) -> compute.ImageReferenceArgs | None:
```
JimMadge, line 49:
```suggestion
    def image_reference(self) -> Optional[compute.ImageReferenceArgs]:
```

JimMadge (changes requested):
This already got merged :open_mouth:.

Routing the Pulumi Inputs/Outputs and dealing with promises adds code, but I'm pretty confident it is a better solution overall. Should be more robust and it offloads handling state to Pulumi.

I don't like all the places that `typing.cast` has been introduced. It feels like sidestepping validating/sanitising input. Which will be necessary because we can't guarantee what a user or API will give to the program.

data_safe_haven/external/interface/azure_ipv4_range.py
jemrobinson: Yeah, sorry about that! If you think of a better alternative feel free to suggest one :)

data_safe_haven/commands/deploy_shm_command.py
jemrobinson: This is a limitation in the type signature of `Command.option`. It returns `dict[Unknown, Unknown] | Unknown | Any | Literal[False]` which are all of the possibilities that `cleo` can coerce the command-line arguments into.

However, this isn't very useful for type checking. I'll have another think about this.

PR #1386, opened 2023-02-06
Pulumi: Fix user list retrieval

### LINKED ISSUE #1384: Pulumi: Missing domain admin password from keyvault

### LINKED ISSUE #1385: Pulumi: Cannot add users to SREs
`list_users.ps1` tests whether the `$SREName` input parameter is $null, but`$SREName` is set to a `[string]`, so it becomes an empty string rather than null. This meant subsequent checks whether `$SREName` is not null always return `True`, and the script would frequently return nothing rather than correctly picking out the users from the domain controller to be added to SREs etc.

craddm:

Fixes Powershell script `list_users.ps1` so that querying the user list correctly returns users from the domain controller, allowing them to be added to SREs etc.

Also adds domain controller password to backend vault

Closes #1385
Also closes #1384

### DIFF
--- a/data_safe_haven/commands/deploy_shm_command.py
+++ b/data_safe_haven/commands/deploy_shm_command.py
@@ -97,6 +97,10 @@ class DeploySHMCommand(LoggingMixin, Command):
                 config.shm.domain_controllers["ldap_searcher_password_secret"],
                 stack.secret("password-domain-ldap-searcher"),
             )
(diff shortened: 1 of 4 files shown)


jemrobinson (approved):
LGTM

PR #1387, opened 2023-02-14
Policy for software package requests

### LINKED ISSUE #622: Process for adding python/R/Julia/Ubuntu/etc. packages
We have two ways in which software packages can be made available for users inside an SRE:

1. Baked into our "batteries included" VM image
- Ubuntu packages
- Julia packages
- python packages
- R packages

2. Available from our mirrors of external repositories
- python packages
- R packages

We need to come up with a process for deciding how to approve what goes into (1) and what goes into (2). Specifically, under what circumstances would we say no to a request for a package and how do we decide whether this request should go into (1) or (2)? Where does the liability belong if a malicious package is included?

@thobson88 is going to take a look at this (initially in the context of the request for ~100 new R packages on #615).

### Stages
- [x] Design a process for deciding whether to include any particular package (done in #671)
- [ ] Automatically generate a nice 1-page PDF summary for each proposed package which highlights any CVEs/safety concerns and lists eg. download stats etc.

jemrobinson:

Add a policy for approving software package requests. Note that this is a policy for adding packages to the main repository, and not a policy for any specific deployment.

Much of the text comes from #622 which was removed during the private-public transition but I think is still relevant here.

Closes #622

### :microscope: Tests

Tested that automated package generation works as expected.

NB. CI test failure is due to the fact that we're adding a new `.md` file and the `Edit this page` link refers back to the non-existent file in GitHub.

### DIFF
--- a/.github/ISSUE_TEMPLATE/software_package_request.md
+++ b/.github/ISSUE_TEMPLATE/software_package_request.md
@@ -1,0 +1,37 @@
+---
+name: Software package request
+about: Request that a new package is added to the Tier 3 allow list
+title: "<short description of issue>"
+labels: "type: enhancement"
+assignees: ""
+---
+
+## :white_check_mark: Checklist
+
+<!--
+Before reporting a problem please check the following. Replace the empty checkboxes [ ] below with checked ones [x] accordingly.
+-->
+
+- [ ] I have searched open and closed issues for duplicates.
+- [ ] This is a request for a new software package to be added to the Data Safe Haven
+- [ ] The package is still missing in the [latest version](https://github.com/alan-turing-institute/data-safe-haven/releases).
+
+## :gift: Package details
+
+<!--
+Provide details about the package you would like to see added:
+
+- Package name
+- Package version (if different from latest)
+- Package repository (e.g. CRAN, PyPI)
+-->
+
+## :steam_locomotive: Why is this needed?
+
+<!--
+To support this request, please respond to the following questions:
+
+- What will you be able to do with this package that you can't currently do?
+- Is this the most widely supported package for the intended purpose? What alternatives are there?
+- What risks to data integrity/security might arise from including this package or its dependencies?
+-->
--- a/.github/ISSUE_TEMPLATE/software_package_request.md
+++ b/.github/ISSUE_TEMPLATE/software_package_request.md
@@ -22,8 +22,13 @@ Before reporting a problem please check the following. Replace the empty checkbo
 Provide details about the package you would like to see added:
 
 - Package name
+- Target audience: "core" (broadly useful for any project) or "extra" (domain specific)
 - Package version (if different from latest)
 - Package repository (e.g. CRAN, PyPI)
+- Number of authors/contributors to the package codebase
+- Any existing versions that should not be used (linking to publicly-accessible CVE databases if relevant)
+- Download statistics (recent and longer-term, for both current and previous versions)
+- List of packages that this package depends on
 -->
 
 ## :steam_locomotive: Why is this needed?
--- a/docs/processes/software_package_approval.md
+++ b/docs/processes/software_package_approval.md
@@ -5,8 +5,15 @@ To maximise the usability of the secure research environments, we pre-install ce
 Other software packages which are only of interest to a subset of users can be made available for download from remote package repositories.
 Currently, we support PyPI (Python) and CRAN (R) as remote repositories.
 
-Only a subset of packages that have been vetted to mitigate the risk of introducing malicious or unsound software into the secure environment are made available in this way.
-This page sets out the policy for adding software packages to the approved download list and/or software to the pre-installed software list.
+For higher {ref}`sensitivity tiers <policy_classification_sensitivity_tiers>` ({ref}`policy_tier_3` and above), only a subset of packages are made available in this way.
+This subset of packages constitutes an "allowlist" of packages that have been vetted to mitigate the risk of introducing malicious or unsound software into the secure environment.
+
+```{warning}
+The Data Safe Haven team manages a default allowlist, but individual deployments may be using their own versions.
+Check with your {ref}`role_system_manager` what is being used in your case
+```
+
+This page sets out the policy for adding software packages to the default allowlist and/or software to the pre-installed software list.
 It also describes the procedure that users of the secure environment should follow to request new additions to the allowlist.
 
 ## Background
--- a/docs/processes/software_package_approval.md
+++ b/docs/processes/software_package_approval.md
@@ -1,0 +1,70 @@
+# Software approval policy
+
+To maximise the usability of the secure research environments, we pre-install certain software packages that are deemed broadly useful to a cross section of researchers, thus making them available to all users.
+
+Other software packages which are only of interest to a subset of users can be made available for download from remote package repositories.
+Currently, we support PyPI (Python) and CRAN (R) as remote repositories.
+
+Only a subset of packages that have been vetted to mitigate the risk of introducing malicious or unsound software into the secure environment are made available in this way.
+This page sets out the policy for adding software packages to the approved download list and/or software to the pre-installed software list.
+It also describes the procedure that users of the secure environment should follow to request new additions to the allowlist.
+
+## Background
+
+Given the safeguards afforded by the safe haven environment, and the separation of responsibilities between its constituent resources, the level of risk associated with the introduction of third party software packages is considered low.
+Moreover, access to the environment is carefully controlled and there is a presumption of trust in the individual researchers to whom access is granted.
+
+Nevertheless, the introduction of any software into the safe haven must be considered against the potential risks of:
+
+- approved users having access to data to which they shouldn't (e.g. from data mixing)
+- unapproved users having access to data (e.g. from a data breach)
+- poisoning of data and/or outputs
+- resource misuse (allocation of computational resources for unintended or wasteful purposes).
+
+Such risks may originate unwittingly, from a user who wants to "just get the job done", or from a user, network team member or administrator acting maliciously.
+
+Specific risks which this policy aims to mitigate include:
+
+- package name squatting (allowlisting a similarly-named package instead of the intended one)
+- privilege escalation attacks (enabling a user to gain elevated access permissions).
+
+(package_inclusion_policy)=
+
+## Policy
+
+- For each supported programming language, three package lists will be maintained:
+    - a core allowlist of broadly useful packages that should be pre-installed in each environment
+    - an extra allowlist of packages that may be useful for specific projects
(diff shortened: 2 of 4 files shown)


docs/processes/software_package_approval.md
```diff
@@ -0,0 +1,81 @@
...
+
+1. A user requests a package by opening a `Software package request` issue on the Data Safe Haven GitHub repository, including responses to the following questions:
+
+- Is this package the mostly widely supported for the intended purpose?
+- What will you be able to do with this package that you can't currently do? What alternatives are there?
+- What risks to data integrity/security might arise from including this package or its dependencies?
+
+2. A member of the project team reviews the request according to the terms of the {ref}`package_inclusion_policy`.
+
+3. The reviewer adds their decision (accept/reject) to the issue and notifies the user who made the request.
```
edwardchalstrey1: If I was a reviewer, how would I know how to make this decision? Is it possible to expand some guidance here - for example, in Daniel from Edon's case, would the reviewer just google "R Arrow" and check it's a real package and somehow figure out if it's something without known security vulnerabilities?

jemrobinson: The issue template asks them to answer the following questions:

- Package name
- Package version (if different from latest)
- Package repository (e.g. CRAN, PyPI)
- Number of authors/contributors to the package codebase
- Any existing versions that should not be used (linking to publicly-accessible CVE databases if relevant)
- Download statistics (recent and longer-term, for both current and previous versions)
- List of packages that this package depends on
- What will you be able to do with this package that you can't currently do?
- Is this the most widely supported package for the intended purpose? What alternatives are there?
- What risks to data integrity/security might arise from including this package or its dependencies?

edwardchalstrey1: Makes sense, so the burden of checking risks falls to the requester

.github/ISSUE_TEMPLATE/software_package_request.md
```diff
@@ -0,0 +1,37 @@
+---
+name: Software package request
+about: Request that a new package is added to the Tier 3 allow list
+title: "<short description of issue>"
+labels: "type: enhancement"
+assignees: ""
+---
```
JimMadge, line 7: How much do we really want to support this at the project level?

It feels to me that the essential things are,

- Have a _minimal_ collection of essential/important software pre-installed (editors, compilers, interpreters, build tools, data science tools)
- Allow users as much flexibility to install their own dependencies as we can while maintaining security
- Allow instance managers to create allow lists for Tier 3

I think it might be a good idea to not distribute a 'default' Tier 3 allow list at all and leave that to particular deployments.

docs/processes/software_package_approval.md
```diff
@@ -0,0 +1,75 @@
+# Software approval policy
+
+To maximise the usability of the secure research environments, we pre-install certain software packages that are deemed broadly useful to a cross section of researchers, thus making them available to all users.
+
+Other software packages which are only of interest to a subset of users can be made available for download from remote package repositories.
+Currently, we support PyPI (Python) and CRAN (R) as remote repositories.
+
+Only a subset of packages that have been vetted to mitigate the risk of introducing malicious or unsound software into the secure environment are made available in this way.
```
JimMadge: Doesn't this only apply to Tier 3 +

docs/processes/software_package_approval.md
```diff
@@ -0,0 +1,75 @@
...
+- unapproved users having access to data (e.g. from a data breach)
+- poisoning of data and/or outputs
+- resource misuse (allocation of computational resources for unintended or wasteful purposes).
+
+Such risks may originate unwittingly, from a user who wants to "just get the job done", or from a user, network team member or administrator acting maliciously.
+
+Specific risks which this policy aims to mitigate include:
+
+- package name squatting (allowlisting a similarly-named package instead of the intended one)
+- privilege escalation attacks (enabling a user to gain elevated access permissions).
```
JimMadge:
```suggestion
- privilege escalation attacks (enabling a user to gain elevated access permissions)
```

To be consistent.

docs/processes/software_package_approval.md
```diff
@@ -0,0 +1,75 @@
...
+
+(package_inclusion_policy)=
+
+## Policy
+
+- For each supported programming language, three package lists will be maintained:
```
JimMadge:
```suggestion
- For each supported repository, three package lists will be maintained:
```

docs/processes/software_package_approval.md
jemrobinson: The burden of explaining why the risks are manageable. We still don't have a well-defined way to make a decision based on these answers though :)

docs/processes/software_package_approval.md
```diff
@@ -0,0 +1,75 @@
...
+  - a core allowlist of broadly useful packages that should be pre-installed in each environment
+  - an extra allowlist of packages that may be useful for specific projects
+  - an expanded list to be made available from the package repositories consisting of the core and extra packages plus their dependencies
+- Users may request to add packages to these allowlists via the {ref}`package request procedure <package_request_procedure>`.
+  - In the interests of improving researcher productivity the aim will be to accommodate such requests, provided there are no outweighing security concerns associated with the package or its dependencies.
+- Requests will be reviewed by the project team using the information provided by the user when making the request
+- If approved, a requested package will be added to either the core or extra allowlist (as appropriate)
```
JimMadge: Again, do we really want to do this? It sounds like we are taking on quite a bit of burden in reviewing and adding packages.

Does this work for the end users? It would be faster for the admins of their instance to update the allowlists than to go through review, merge, release, update.

docs/processes/software_package_approval.md
```diff
@@ -0,0 +1,70 @@
...
+- unapproved users having access to data (e.g. from a data breach)
+- poisoning of data and/or outputs
+- resource misuse (allocation of computational resources for unintended or wasteful purposes).
+
+Such risks may originate unwittingly, from a user who wants to "just get the job done", or from a user, network team member or administrator acting maliciously.
+
+Specific risks which this policy aims to mitigate include:
+
+- package name squatting (allowlisting a similarly-named package instead of the intended one)
+- privilege escalation attacks (enabling a user to gain elevated access permissions).
```
JimMadge: It might be worth adding a point here about data ingress. In particular with PyPI as there is no review it would be fairly easy to make a package just containing (a large amount of) data.

JimMadge (commented):
I think I'm not convinced that we _should_ be doing this.

I really like the explanation of how we are confident about the security of the SRE, but that allowing data from outside presents risks to data mixing or other misuse.

The criteria for what we would consider adding seem quite broad.
I think this has caused us problems in the past, particularly with Python packages are resolving all dependencies across many packages is difficult (or impossible).

If that is what we really want to do, we should probably look at using lmod (or another module system) to isolate those packages.

docs/processes/software_package_approval.md
```diff
@@ -0,0 +1,70 @@
...
+
+Only software that is considered broadly useful to a cross section of researchers should be included in core.
+
+To meet this condition, a package should:
+
+- implement at least one generic (i.e. not domain-specific) statistical algorithm or method, or
+- provide support for a cross-cutting analysis technique (e.g. geospatial data analysis, NLP), or
+- facilitate data science or software development best practices (e.g. for robustness, correctness, reproducibility), or
+- enhance the presentational features of the programming language (e.g. for producing plots, notebooks, articles, websites), or
+- enhance the usability of the programming language or development environment (e.g. RStudio, PyCharm).
```
JimMadge:
```suggestion
- enhance the usability of the programming language or development environment (e.g. RStudio, PyCharm)
```

.github/ISSUE_TEMPLATE/software_package_request.md
jemrobinson: There was some discussion in the RSE Slack about setting up a centrally-curated vetted list of packages. Maybe we could try to resurrect that idea?

The idea of having a 'default' Tier 3 list is that we (DSH project) are saying "we've had some experience with a few hundred end-users and here are the things they generally want and that we think are sensible YMMV". What do you think?

docs/processes/software_package_approval.md
jemrobinson: Done

docs/processes/software_package_approval.md
jemrobinson: Done

docs/processes/software_package_approval.md
jemrobinson: Tier-3 allowlists are currently SHM-wide. Allowing all projects access to a particular package is a reasonably big decision. Any package that convinces the DSH administrators that it should be allowed to all projects is presumably a reasonable candidate to be in our "default" allowlist.

Running a default allowlist is also a usability feature - we're not saying "work out what's allowed" to each deployment, we're saying "here's a default, but use your own if you prefer".

I agree, it would definitely be faster to manually update the allowlist, but that means that the work that's been put into evaluating the package doesn't come back to the upstream project.

.github/ISSUE_TEMPLATE/software_package_request.md
JimMadge: That would be interesting. I like the idea of these being a well supported, external (even if we are quite involved with it) project which curates 'safe' packages for research/data science.

That removes some burden from us on doing it by ourselves. It also adds to this idea that you can make a union of existing, curated allow lists to combine sensible defaults and your own personal requirements.

docs/processes/software_package_approval.md
JimMadge: Oh, I had forgotten it was per SHM. Have we thought about when we might want to have different allow lists per project?

In that case, anything that gets approved is much closer to being one of the pre-installed packages than I was imagining.

> "here's a default, but use your own if you prefer"

I suppose I'm a little nervous about the risk of us getting it wrong. It feels a touch more like a guarantee than other bits of policy where we just say "this is what _we_ do, but you need to think about your own situation".

jemrobinson: That's a good point but we *do* have an existing Tier-3 list which people who run an out-of-the-box deployment will be using. We might not be comfortable in curated a set of packages but the fact is that we already are. Maybe our focus should be on switching to a community-developed list instead?

JimMadge: Sounds good. In terms of actions, shall we

1. Add this to the project board as an item for RSE-TR-WG?
2. See if this fits into any part of SATRE?

JimMadge (approved):

docs/processes/software_package_approval.md
jemrobinson: Definitely (1) and we can ask about (2) but that might be better covered by one of the other DARE-funded projects.

PR #1389, opened 2023-02-16
Update user management guide to explain adding users to security group and changing a phone number

edwardchalstrey1:

- ~~If-and-only-if you have changed any Powershell code, you have run the code formatter. You can do this with `./tests/AutoFormat_Powershell.ps1 -TargetPath <path to file or directory>`.~~

Update user management guide (for system managers) to explain adding users to security group and changing a phone number

Closes https://github.com/alan-turing-institute/trusted-research/issues/146

### :microscope: Tests

Test by running `make html` in `docs`

### TODO

- [x] Update to include editing a person's name:
    - process 1) Delete the user in DC1 2) adsync 3) rerun the createuserscript with correct name and username 4) adsync
- [x] Remove Portal steps (apart from at the end to check something worked, which may take a few mins)

### DIFF
--- a/docs/deployment/snippets/user_csv_format.partial.md
+++ b/docs/deployment/snippets/user_csv_format.partial.md
@@ -44,7 +44,7 @@
       This is **not** uploaded to their Data Safe Haven user account but is needed when sending account activation messages.
       ```
 
-    - `GroupName`: [Optional] The name of the Active Directory security group(s) that the users should be added to (eg. `SG SANDBOX Research Users` ). You can also set this manually later on (see {ref}`adding_users_manually`).
+    - `GroupName`: The name of the Active Directory security group(s) that the users should be added to (eg. `SG SANDBOX Research Users` ). You can also change this manually later on (see {ref}`adding_users_manually`).
 
       ```{tip}
       If the user needs to be added to multiple groups, separate them with a pipe-character ( `|` ).
--- a/docs/deployment/snippets/user_csv_format.partial.md
+++ b/docs/deployment/snippets/user_csv_format.partial.md
@@ -44,7 +44,7 @@
       This is **not** uploaded to their Data Safe Haven user account but is needed when sending account activation messages.
       ```
 
-    - `GroupName`: The name of the Active Directory security group(s) that the users should be added to (eg. `SG SANDBOX Research Users` ). You can also change this manually later on (see {ref}`adding_users_manually`).
+    - `GroupName`: The name of the Active Directory security group(s) that the users should be added to (eg. `SG SANDBOX Research Users` ). Users will need to be added to the relevant security group before they can access an SRE. You can also change this manually later on (see {ref}`adding_users_manually`).
 
       ```{tip}
       If the user needs to be added to multiple groups, separate them with a pipe-character ( `|` ).
--- a/docs/deployment/snippets/user_csv_format.partial.md
+++ b/docs/deployment/snippets/user_csv_format.partial.md
@@ -44,7 +44,7 @@
       This is **not** uploaded to their Data Safe Haven user account but is needed when sending account activation messages.
       ```
 
-    - `GroupName`: The name of the Active Directory security group(s) that the users should be added to (eg. `SG SANDBOX Research Users` ). Users will need to be added to the relevant security group before they can access an SRE. You can also change this manually later on (see {ref}`adding_users_manually`).
+    - `GroupName`: The name of the {ref}`security_groups` that the users should be added to. Users will need to be added to the relevant security group before they can access an SRE. You can also change this manually later on (see {ref}`adding_users_manually`).
 
       ```{tip}
       If the user needs to be added to multiple groups, separate them with a pipe-character ( `|` ).
--- a/docs/deployment/snippets/user_csv_format.partial.md
+++ b/docs/deployment/snippets/user_csv_format.partial.md
@@ -44,7 +44,7 @@
       This is **not** uploaded to their Data Safe Haven user account but is needed when sending account activation messages.
       ```
 
-    - `GroupName`: [Optional] The name of the Active Directory security group(s) that the users should be added to (eg. `SG SANDBOX Research Users` ).
+    - `GroupName`: [Optional] The name of the Active Directory security group(s) that the users should be added to (eg. `SG SANDBOX Research Users` ). You can also set this manually later on (see {ref}`adding_users_manually`).
 
       ```{tip}
       If the user needs to be added to multiple groups, separate them with a pipe-character ( `|` ).
--- a/docs/roles/system_manager/manage_users.md
+++ b/docs/roles/system_manager/manage_users.md
@@ -64,6 +64,24 @@ Once you're certain that you're adding a new user, make sure that the following 
 - Be particularly careful never to use them to log in to any user-accessible VMs (such as the SRDs)
 ```
 
+### {{woman}} {{man}} Adding users to security group(s)
+
+Users that didn't have a `GroupName` specified in the `user_details_template.csv`, will need to be added manually, before they can access an SRE.
+
+- Log into the **SHM primary domain controller** (`DC1-SHM-<SHM ID>`) VM using the login credentials {ref}`stored in Azure Key Vault <roles_system_deployer_shm_remote_desktop>`
+- Open `Active Directory Users and Computers`
+- Click on `Safe Haven Security Groups`
+- Find the group that the user needs to be added to. For each SRE there are the following groups:
+    - `SG <SRE ID> Data Administrators` (enables database privileges, not write access to `/data`)
+    - `SG <SRE ID> Research Users`
+    - `SG <SRE ID> System Administrators`
+- Right click on the group and click `Properties`
+- Click the `Members` tab and click `Add...`
+- Enter a part of the user's name and click `Check Names`
+- Select the correct user and click `OK`, then click `OK` again until the window closes
+- Open a `Powershell` command window with elevated privileges
+- Run `C:\Installation\Run_ADSync.ps1`
+
 ### {{iphone}} Changing a users phone number
 
 Sometimes if a user has provided an incorrect phone number, or has a new phone number that needs to be added, the phone number associated with their account will need to be updated.
--- a/docs/roles/system_manager/manage_users.md
+++ b/docs/roles/system_manager/manage_users.md
@@ -93,7 +93,7 @@ Users will need to be added to the relevant security group before they can acces
 
 ### {{iphone}} Edit user details
 
-The `DC1` is the source of truth for user details. If these details need to be changed, the user should be deleted in the Azure Portal and re-created from the `DC1`.
+The `DC1` is the source of truth for user details. If these details need to be changed, they should be changed in the `DC1` and then synchronised to Azure.
 
 - Log into the **SHM primary domain controller** (`DC1-SHM-<SHM ID>`) VM using the login credentials {ref}`stored in Azure Key Vault <roles_system_deployer_shm_remote_desktop>`
 - In Server Manager click `Tools > Active Directory Users and Computers`
@@ -105,6 +105,10 @@ The `DC1` is the source of truth for user details. If these details need to be c
 - Click `OK` after any changes are made
 - Open a `Powershell` command window with elevated privileges
 - Run `C:\Installation\Run_ADSync.ps1`
+- You can check the changes you made were successful by logging into the Azure Portal as the AAD admin
+    - Open `Azure Active Directory`
+    - Click on `Users` under `Manage` and search for the user
+    - Click on the user and then `Edit properties` and confirm your changes propagated to Azure
 
 (deleting_users)=
 
--- a/docs/roles/system_manager/manage_users.md
+++ b/docs/roles/system_manager/manage_users.md
@@ -86,7 +86,7 @@ Users that didn't have a `GroupName` specified in the `user_details_template.csv
 
 If a user has provided an incorrect phone number, or has a new phone number, the phone number associated with their account will need to be updated.
 
-- Login into the Azure Portal and connect to the correct AAD
+- Login into the Azure Portal and navigate to the correct AAD
 - Open `Azure Active Directory`
 - Click `Users` under `Manage`
 - Search for the user and check the box next to their name, then click `Delete` (this could take a few minutes)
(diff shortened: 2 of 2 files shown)


docs/roles/system_manager/manage_users.md
```diff
@@ -64,6 +64,40 @@ Once you're certain that you're adding a new user, make sure that the following
 - Be particularly careful never to use them to log in to any user-accessible VMs (such as the SRDs)
 ```
 
+### {{woman}} {{man}} Adding users to security group(s)
+
+Users that didn't have a `GroupName` specified in the `user_details_template.csv`, will need to be added manually, before they can access an SRE.
```
JimMadge: This is a useful tip, but I think the instructions here are more general than fixing this one problem.

Maybe something like this would be better as a warning further up the page where editing the csv is discussed? At the moment, it says this is optional, which is true, but it sounds like users are unable to work until they are added to an appropriate group.

docs/roles/system_manager/manage_users.md
```diff
@@ -64,6 +64,40 @@ Once you're certain that you're adding a new user, make sure that the following
 - Be particularly careful never to use them to log in to any user-accessible VMs (such as the SRDs)
 ```
 
+### {{woman}} {{man}} Adding users to security group(s)
```
JimMadge: Should we also explain removing users from groups (or modifying groups)?

docs/roles/system_manager/manage_users.md
```diff
@@ -64,6 +64,40 @@ Once you're certain that you're adding a new user, make sure that the following
...
+- Right click on the group and click `Properties`
+- Click the `Members` tab and click `Add...`
+- Enter a part of the user's name and click `Check Names`
+- Select the correct user and click `OK`, then click `OK` again until the window closes
+- Open a `Powershell` command window with elevated privileges
+- Run `C:\Installation\Run_ADSync.ps1`
+
+### {{iphone}} Changing a users phone number
+
+Sometimes if a user has provided an incorrect phone number, or has a new phone number that needs to be added, the phone number associated with their account will need to be updated.
```
JimMadge:
```suggestion
If a user has provided an incorrect phone number, or has a new phone number, the phone number associated with their account will need to be updated.
```

docs/roles/system_manager/manage_users.md
```diff
@@ -64,6 +64,40 @@ Once you're certain that you're adding a new user, make sure that the following
...
+
+- Login into the Azure Portal and connect to the correct AAD
```
JimMadge:
```suggestion
- Login into the Azure Portal and navigate to the correct AAD
```

docs/roles/system_manager/manage_users.md
```diff
@@ -64,6 +64,40 @@ Once you're certain that you're adding a new user, make sure that the following
...
+- Open `Azure Active Directory`
+- Click `Users` under `Manage`
+- Search for the user and check the box next to their name, then click `Delete` (this could take a few minutes)
```
JimMadge: If I recall correctly, this is because we can't change the phone numbers on AAD so we essentially recreate the AAD user from the local AD.

Is that right?
It might be worth adding some text to that effect to explain why this process looks a little odd.

JimMadge (changes requested):
Looks good, just a few comments/suggestions.

docs/roles/system_manager/manage_users.md
edwardchalstrey1: Perhaps "Managing security group(s)" which can include addition and deletion

docs/roles/system_manager/manage_users.md
edwardchalstrey1: Sure, yes that what @jemrobinson suggested, that the DC1 is the source of truth for user info

docs/roles/system_manager/manage_users.md
JimMadge: :+1:

If it starts to get very large it might be worth breaking it down into addition, deletion, _etc._

edwardchalstrey1:
@JimMadge thanks for you comments, I've separated into 3 sections for Adding to security groups, editing users and deleting them, all of which are slightly different but overlapping processes

docs/roles/system_manager/manage_users.md
jemrobinson: 1. the user shouldn't need to be deleted in AzureAD, in fact doing this will require the user to re-setup MFA
2. email is the one that can't be changed in AzureAD (once set it's used as a possible UUID for the user)

edwardchalstrey1: Ah ok, so we don't want to delete in AzureAD that when changing details.

Shall we say that if email is the thing that needs to be changed, the user should be deleted entirely (and that setting up MFA again will be required)?

edwardchalstrey1: See my edit

docs/roles/system_manager/manage_users.md
```diff
@@ -64,9 +66,11 @@ Once you're certain that you're adding a new user, make sure that the following
 - Be particularly careful never to use them to log in to any user-accessible VMs (such as the SRDs)
 ```
 
+(adding_users_manually)=
+
 ### {{woman}} {{man}} Adding users to security group(s)
 
-Users that didn't have a `GroupName` specified in the `user_details_template.csv`, will need to be added manually, before they can access an SRE.
+Users will need to be added to the relevant security group before they can access an SRE. This may have already been done if the `GroupName` was specified in the `user_details_template.csv` (see {ref}`generate_user_csv`), but otherwise can be achieved by the following:
```
JimMadge: Again, this feels like it is in the wrong place.

This section is about adding a user to a group, for any purpose. I think the warning about users not being able to access the SRE if they are in no group would be better placed with the creating users/users csv. That way we might help people avoid that mistake. If you are reading about it here, you have probably already messed up.

docs/roles/system_manager/manage_users.md
```diff
@@ -82,9 +86,9 @@ Users that didn't have a `GroupName` specified in the `user_details_template.csv
 - Open a `Powershell` command window with elevated privileges
 - Run `C:\Installation\Run_ADSync.ps1`
 
-### {{iphone}} Changing a users phone number
+### {{iphone}} Edit user details
 
-If a user has provided an incorrect phone number, or has a new phone number, the phone number associated with their account will need to be updated.
+The `DC1` is the source of truth for user details. If these details need to be changed, the user should be deleted in the Azure Portal and re-created from the `DC1`.
```
JimMadge: Will delegate to @jemrobinson here.

I _think_ user details should sync between AAD and the local AD, so deleting and recreating the user shouldn't often be necessary. I recall one entry being fixed, which it seems was email address and not phone number. That might be the only case where you would make a new user.

JimMadge (changes requested):

docs/roles/system_manager/manage_users.md
jemrobinson: Deleting the user on the DC should also delete them on AzureAD - you should (hopefully) not need to edit users on AzureAD at all.

docs/roles/system_manager/manage_users.md
jemrobinson: Yes, the users in AzureAD are synchronised from the local AD on the DC. So all changes (apart from resetting MFA) should be done on the local DC.

JimMadge: An that is unidirectional? _I.e._ no sync from AAD to local AD?

jemrobinson: There is password hash synchronisation, but I think that's just that when a password is reset the hash is sent to both AzureAD and the local AD, I don't think the password hash is synchronised from AzureAD to the local AD.

The only thing I can think of that's synchronised from AzureAD to the local AD is the AzureAD anchor attribute which identifies which local AD user maps to which AzureAD user.

edwardchalstrey1:
@JimMadge I've restructured this again after you and @jemrobinson comments above - also separated email/username changes from phone no.

docs/deployment/snippets/user_csv_format.partial.md
```diff
@@ -44,7 +44,7 @@
       This is **not** uploaded to their Data Safe Haven user account but is needed when sending account activation messages.
       ```
 
-    - `GroupName`: [Optional] The name of the Active Directory security group(s) that the users should be added to (eg. `SG SANDBOX Research Users` ).
+    - `GroupName`: The name of the Active Directory security group(s) that the users should be added to (eg. `SG SANDBOX Research Users` ). You can also change this manually later on (see {ref}`adding_users_manually`).
```
JimMadge:
```suggestion
    - `GroupName`: The name of the Active Directory security group(s) that the users should be added to (eg. `SG SANDBOX Research Users` ). Users will need to be added to the relevant security group before they can access an SRE. You can also change this manually later on (see {ref}`adding_users_manually`).
```

docs/roles/system_manager/manage_users.md
```diff
@@ -64,6 +68,69 @@ Once you're certain that you're adding a new user, make sure that the following
 - Be particularly careful never to use them to log in to any user-accessible VMs (such as the SRDs)
 ```
 
+(adding_users_manually)=
+
+### {{woman}} {{man}} Modifying user SRE access
+
+Users will need to be added to the relevant security group before they can access an SRE. This will have been done by setting the `GroupName` in the `user_details_template.csv` (see {ref}`generate_user_csv`), but can me modified later by the following:
```
JimMadge:
```suggestion
Users may have been added to one or more security groups through setting the `GroupName` field in the `user_details_template.csv` (see {ref}`generate_user_csv`). Security Group assignments can also be manually modified```

File: docs/roles/system_manager/manage_users.md [hunk 10]
```diff
@@ -64,6 +68,69 @@ Once you're certain that you're adding a new user, make sure that the following
...
+
+- Log into the **SHM primary domain controller** (`DC1-SHM-<SHM ID>`) VM using the login credentials {ref}`stored in Azure Key Vault <roles_system_deployer_shm_remote_desktop>`
+- In Server Manager click `Tools > Active Directory Users and Computers`
+- Click on `Safe Haven Security Groups`
+- Find the group that the user needs to be added to. For each SRE there are the following groups:
+    - `SG <SRE ID> Data Administrators` (enables database privileges, not write access to `/data`)
+    - `SG <SRE ID> Research Users`
+    - `SG <SRE ID> System Administrators`
```
JimMadge: This is good, is there an explanation in the documents for these groups and their meaning? That feels like a missing piece of the puzzle here.

It might be best to put those definitions in the documentation as a reference and link to it here (and in the section about creating users with the csv).

docs/roles/system_manager/manage_users.md
```diff
@@ -64,6 +68,69 @@ Once you're certain that you're adding a new user, make sure that the following
...
+    - Enter a part of the user's name and click `Check Names`
+    - Select the correct user and click `OK`, then click `OK` again until the window closes
+- To remove a user click on the username of the person and then `Remove`
+    - Click `Yes` if you're sure this user should no longer have access to this SRE, then click `OK` again until the window closes
+- Open a `Powershell` command window with elevated privileges
+- Run `C:\Installation\Run_ADSync.ps1`
+
+### {{iphone}} Edit user details
+
+The `DC1` is the source of truth for user details. If these details need to be changed, they should be changed in the `DC1` and then synchronised to Azure.
```
JimMadge:
```suggestion
The `DC1` is the source of truth for user details. If these details need to be changed, they should be changed in the `DC1` and then synchronised to Azure AD.
```

docs/roles/system_manager/manage_users.md
```diff
@@ -64,6 +68,69 @@ Once you're certain that you're adding a new user, make sure that the following
...
+    - Click `OK`
+    - Open a `Powershell` command window with elevated privileges
+    - Run `C:\Installation\Run_ADSync.ps1`
+    - Create a new csv (or edit an existing) one with the correct user details (see {ref}`create_new_users`)
+    - Run `C:\Installation\CreateUsers.ps1 <path_to_user_details_file>`
+    - Run `C:\Installation\Run_ADSync.ps1`
+- You can check the changes you made were successful by logging into the Azure Portal as the AAD admin
+    - Open `Azure Active Directory`
+    - Click on `Users` under `Manage` and search for the user
+    - Click on the user and then `Edit properties` and confirm your changes propagated to Azure
```
JimMadge:
```suggestion
    - Click on the user and then `Edit properties` and confirm your changes propagated to Azure AD
```

JimMadge (changes requested):
Looks good.

I think I wasn't explaining what I meant very well so I've suggested an edit.

It would be good to get the security groups listed and explained in one place then reference it throughout this page.

docs/roles/system_manager/manage_users.md
edwardchalstrey1: It doesn't look like there's a good explanation for these groups anywhere. Do we think that based on the "Roles" it should be: 

- Researcher: `SG <SRE ID> Research Users`
- Investigator: `SG <SRE ID> Data Administrators`
- System Manager: `SG <SRE ID> System Administrators`

Screenshot 2023-02-21 at 14 09 33

I'm unsure why for example researchers don't have database creation privileges? (assuming I'm correct you need to be in `SG <SRE ID> Data Administrators` for that

JimMadge: I think all of those security groups apply to Researchers and it comes from principle of least privilege.

- Research Users - Probably most Researchers, not special permissions
- Data Administrators - Researchers who can create/modify/delete database ~~tables~~ schemas. Given to a smaller number of Researchers to prevent accidents.
- System Administrators - Researchers with elevated privileges through `sudo`. Rarely used but could be useful in Tier0/1 to let groups manage their own packages.

@jemrobinson, @martintoreilly is that correct?

jemrobinson: The reason for normal users not to be able to create/delete arbitrary schemas is because some SREs have their input data in database form (and we don't want users to be able to change that). All users should be able to create tables in the `public` schema though.

edwardchalstrey1: Ok, I will update this info - I think it should still go on the `System Manager -> Managing Data Safe Haven Users` page, but near the top and referenced later

JimMadge (approved):

PR #1390, opened 2023-02-20
Add Python type-hinting throughout Pulumi codebase

jemrobinson:

- [x] You are targeting the **~develop~ python-migration branch**.
- [x] Your branch is up-to-date with the **~develop~ python-migration branch** (you probably started your branch from `develop` but it may have changed since then).

- Add type hints to codebase
- Add typing stubs where needed

Fix unresolved issues from #1375

### :microscope: Tests

Tested by running `mypy` locally.


jemrobinson:
@JimMadge : this is mostly ready to review (just two files that aren't passing `mypy: strict` at the moment). I've removed the `cast`s that you thought were ugly/unnecessary and added type-hinting throughout. This has also uncovered a couple of type-related bugs so these are fixed here too.

data_safe_haven/cli.py
```diff
@@ -18,6 +18,6 @@
 application.add(UsersCommand())
 
 
-def main():
+def main() -> int:
```
JimMadge, line 21: Seems very old fashioned for Python :smile:

data_safe_haven/commands/deploy_sre_command.py
```diff
@@ -108,14 +118,16 @@ def handle(self) -> None:
...
 
         except DataSafeHavenException as exc:
             for (
                 line
-            ) in f"Could not deploy Secure Research Environment {self.argument('name')}.\n{str(exc)}".split(
+            ) in f"Could not deploy Secure Research Environment {self.sre_name}.\n{str(exc)}".split(
                 "\n"
             ):
                 self.error(line)
+        return 1
```
JimMadge, line 130: Does this automatically get interpreted as an exit code?

data_safe_haven/pulumi/transformations.py
```diff
@@ -0,0 +1,46 @@
+"""Common transformations needed when manipulating Pulumi resources"""
```
JimMadge, line 1: Good idea to keep the code DRY

data_safe_haven/pulumi/components/sre_research_desktop.py
```diff
@@ -40,21 +44,21 @@ def __init__(
...
-        ).apply(lambda s: s.name)
+        ).apply(get_name_from_subnet)
         self.virtual_network_resource_group_name = Output.from_input(
             virtual_network_resource_group
-        ).apply(lambda rg: rg.name)
+        ).apply(get_name_from_rg)
         self.vm_ip_addresses = Output.all(subnet_research_desktops, vm_details).apply(
             lambda args: self.get_ip_addresses(
-                cast(network.GetSubnetResult, args[0]), len(cast(list, args[1]))
+                cast(network.GetSubnetResult, args[0]), len(cast(List[str], args[1]))
```
JimMadge: Is this one avoidable?

data_safe_haven/pulumi/transformations.py
```diff
@@ -0,0 +1,46 @@
...
+# Third party imports
+from pulumi_azure_native import network, resources
+
+# Local imports
+from data_safe_haven.exceptions import DataSafeHavenPulumiException
+from data_safe_haven.external.interface import AzureIPv4Range
+
+
+def get_available_ips_from_subnet(subnet: network.GetSubnetResult) -> List[str]:
+    if address_prefix := subnet.address_prefix:
```
JimMadge, line 15: Walrus in the wild :smile:

JimMadge (commented):
I think that is looking better. Painful to add now but I hope in the future this will,

- Help avoid errors/bugs due to types mismatching
- Ensure that input is properly parsed/sanitised

Are the `.pyi` files necessary? They are for modules we import?

data_safe_haven/commands/deploy_sre_command.py
jemrobinson: I think so, but I haven't tested it.

data_safe_haven/pulumi/components/sre_research_desktop.py
jemrobinson: It's difficult. The only type that `mypy` can infer for `args` here is `List[network.GetSubnetResult | List[str]]`. This means that we can't pass `args[0]` and `args[1]` to `get_ip_addresses` since this expects one `network.GetSubnetResult` and one `List[str]`.

The only rewrite I can think of is to change `get_ip_addresses` to accept `List[network.GetSubnetResult | List[str]]` directly and do the type checking inside that function. It's a bit messy though and means that the conversion will be done in a different place than the original obfuscation.

Maybe you can think of another option?

data_safe_haven/commands/deploy_sre_command.py
JimMadge: Getting increasingly suspicious of `cleo`. That seems very unpythonic compared to raising an exception.

Shouldn't the Python interpreter decide the return code, not the code being executed?

jemrobinson:
> Are the `.pyi` files necessary? They are for modules we import?

These are effectively header files for libraries we import that don't have native type hinting (or don't advertise that they do). For common libraries you can normally install a `library-name-types` package but these don't always exist. Additionally, for some of the `azure` libraries I've changed the `.pyi` file to match the return value the function *actually* gives, not the one advertised in the docs.

data_safe_haven/commands/deploy_sre_command.py
jemrobinson: It's not too late to dump cleo and use one of the other CLI builders instead. IMO it tries to do to much (and be a bit too clever) which is why it's also got a load of IO and formatting functions in there too.

data_safe_haven/pulumi/components/sre_research_desktop.py
JimMadge: I suppose `Output.all()` is required to combine the two outputs into a single new output? Otherwise Pulumi can't ensure both promises have values?

In that case, `get_ip_addresses` could take arguments of the union type, but that seems bad because neither input is really a union. Feels like a case where it is appropriate to use Any and rely on duck typing.

I have opened a PR on your fork with a suggestion https://github.com/jemrobinson/data-safe-haven/pull/2
(Not sure if there is a more elegant want to do this. I can't push straight to your fork)

data_safe_haven/commands/deploy_sre_command.py
JimMadge: Shall we open a discussion and look at some options? Probably less painful to do it now than later.

JimMadge:
> These are effectively header files for libraries we import that don't have native type hinting (or don't advertise that they do). For common libraries you can normally install a library-name-types package but these don't always exist. Additionally, for some of the azure libraries I've changed the .pyi file to match the return value the function actually gives, not the one advertised in the docs.

I hope the starting point was automatically generated somehow :scream:.

Sorry, I asked the wrong question. I think what I'm really trying to ask is,

- Does mypy fail without these?
- If it does,
  - Do we really want to maintain these ourselves?
  - Isn't that an upstream bug?
  - Can we use `#  type: ignore` instead?

Also, did you get `mypy` to play nicely with Pulumi? I remember having problems which I think boiled down to Pulumi being so big it was impractical to do a thorough type check.

jemrobinson:
>     * Does mypy fail without these?

It complains about missing stubs (eg. `Skipping analyzing "azure.mgmt.dns": module is installed, but missing library stubs or py.typed marker`)
 
>     * If it does,
>       * Do we really want to maintain these ourselves?

Not really.

>       * Isn't that an upstream bug?

Yes, but it's not a bug since type-hinting is still very optional.

>       * Can we use `#  type: ignore` instead?

Yes.
 
> Also, did you get `mypy` to play nicely with Pulumi? I remember having problems which I think boiled down to Pulumi being so big it was impractical to do a thorough type check.

Yes, the trick is actually to use a `.pyi` stub that replaces `import everything` with `import the things we need`.

I now have everything working with the stub files. I can try removing them if you prefer this?

data_safe_haven/commands/deploy_sre_command.py
JimMadge: Skeleton discussion [here](https://github.com/alan-turing-institute/data-safe-haven/discussions/1394)

JimMadge:
> I now have everything working with the stub files. I can try removing them if you prefer this?

I think we should try and make that work, especially as we agree there is no desire to maintain these stub files.

As type hinting is, and presumably will remain, optional I think it makes most sense to not type untyped libraries ourselves.

jemrobinson:
@JimMadge : Now ready for review with type-hinting stubs removed and `# type: ignore` used instead.

JimMadge (approved):
LGTM :+1:

I think the Pulumi stub file is a good idea to help type checking finish in a reasonable time without overwriting Pulumi's own typing hints.

PR #1397, opened 2023-02-22
Allow sudo for members of SRE System Administrators security group

### LINKED ISSUE #1396: Update System Manager docs on how to give tier 0/1 users sudo privileges

On the [users management page](https://alan-turing-institute.github.io/data-safe-haven/develop/roles/system_manager/manage_users.html) after #1389 is merged it would be good to also specify that the `SG <SRE ID> System Administrators` security group is not sufficient to grant sudo privileges.

## :memo: Required documentation changes

The process that did work:

- Log into Serial Console of the compute VM with the admin username and pass stored in the key vault
- `sudo usermod -aG sudo firstname.lastname@prod4.turingsafehaven.ac.uk`
- Check it worked with `getent group sudo` and also trying a sudo cmd as that user in the SRE

JimMadge:

- [ ] If-and-only-if you have changed any Powershell code, you have run the code formatter. You can do this with `./tests/AutoFormat_Powershell.ps1 -TargetPath <path to file or directory>`.

- Add a `sudoers.d` file to allow members of an SREs System Administrators security group to use sudo.

Closes #1396

### DIFF
--- a/deployment/secure_research_environment/cloud_init/cloud-init-srd.mustache.yaml
+++ b/deployment/secure_research_environment/cloud_init/cloud-init-srd.mustache.yaml
@@ -221,6 +221,11 @@ write_files:
     content: |
       xfce4-session
 
+  - path: "/etc/sudoers.d/10-ad-sg-system-administrators"
+    permissions: "0440"
+    content: |
+      %sg\ {{sre.id}}\ system\ administrators ALL=(ALL:ALL) ALL
+
   - path: "/etc/systemd/system/clamav-clamonacc.service"
     permissions: "0644"
     content: |


JimMadge:
@jemrobinson is `sre.id` the right config variable? It should be the sre name as it appears in the security group names _e.g._ sandbox.

JimMadge:
@edwardchalstrey1 could you test this `in place` by adding the equivalent file to an environment where you are a member of the system administrators group?

edwardchalstrey1:
> @edwardchalstrey1 could you test this `in place` by adding the equivalent file to an environment where you are a member of the system administrators group?

@JimMadge how can I do this? should I edit the file locally and re-run one of the deployment scripts?

JimMadge:
@edwardchalstrey1 No need to redeploy. If you have an SRD running in an SRE where you are in the system administrators group, create a file at that path (with those permissions) and with those contents (expanding the `{sre.id}`). You will need to do that as root with the VM admin account.

After that your user (and anyone in the group) should be able to use `sudo`. I think you will have to restart after adding the `sudoers.d` file.

edwardchalstrey1:
@JimMadge 
1. I created this file as root in the serial console oof Sandbox SRD VM
    Screenshot 2023-02-24 at 10 12 02
2. then ran:
    ```
     sresandboxadmin:~$ sudo chmod 0440 /etc/sudoers.d/10-ad-sg-system-administrators
    ```
4. then restarted the VM
5. Then logged into the SRE on that SRD and tried a sudo cmd, but still got:
    ```
     ed.chalstrey1 is not allowed to run sudo on SRE-SANDBOX-180.  This incident will be reported.
    ```

I checked and `ed.chalstrey1` is definitely in the System Administrators SG for Sandbox

JimMadge:
You need to replace `{{sre.id}}` with the SRE name, so that the sudoers line gives the name of the administrators group.

I assume in this case it is "sandbox", so `%sg\ sandbox\ system\ administrators ALL=(ALL:ALL) ALL`

edwardchalstrey1:
@JimMadge I've done the same again with the correct SRE ID but it still doesn't work unfortunately

like: `%sg\ {{sandbox}}\ system\ administrators ALL=(ALL:ALL) ALL`

JimMadge:
Could you try without the braces, like this: `%sg\ sandbox\ system\ administrators ALL=(ALL:ALL) ALL`.

edwardchalstrey1:
ah of course, because `sre.id` was a variable but sandbox I'm hardcoding - hopefully doesn't need quotes! will try now

edwardchalstrey1:
@JimMadge ok yes that worked

edwardchalstrey1 (approved):

JimMadge:
@jemrobinson @martintoreilly This would be a fairly significant change to make. I think we should check we are all happy (and double check the documentation is consistent).

jemrobinson:
Who's currently added to "System Administrators" group by default?
- If it's no-one (or just global sysadmins) then adding this code is only allowing for the possibility of giving people sudo access which I think is fine
- If anyone is currently added by default, we should consider removing them if we want to merge this.

JimMadge:
I assume (but might be wrong) that no-one is added to any of the security groups by default.

The docs currently explain the SG's [here](https://alan-turing-institute.github.io/data-safe-haven/develop/roles/system_manager/manage_users.html#sre-security-groups)

> SG <SRE ID> System Administrators - Researchers with elevated privileges through sudo. Rarely used but could be useful in Tier 0/1 SREs to let groups manage their own packages.

And these definitions are linked to in the SHM deployment guide where users are created with the csv file.

This might be concerning

https://github.com/alan-turing-institute/data-safe-haven/blob/bd5e9fee50377b36764de25fa665d3c7d63e4bd1/deployment/secure_research_environment/cloud_init/resources/guacamole_db_update.mustache.sql#L104-L135

https://github.com/alan-turing-institute/data-safe-haven/blob/bd5e9fee50377b36764de25fa665d3c7d63e4bd1/deployment/secure_research_environment/cloud_init/postgres_create_postgres_triggers.mustache.sql#L1

Also passed to [Lockdown_Sql_Server.ps1](https://github.com/alan-turing-institute/data-safe-haven/blob/bd5e9fee50377b36764de25fa665d3c7d63e4bd1/deployment/secure_research_environment/remote/create_databases/scripts/Lockdown_Sql_Server.ps1) as `SysAdminGroup`.

So, it looks like this group was intended to be database and guacamole admins?

jemrobinson:
:warning: @edwardchalstrey1 if you have given anyone membership of the `SG System Administrators` group, please remove it now :warning:

@JimMadge : Yes, I think this makes sense.
- System Administrators: admin rights on the Guacamole user console (although this shouldn't ever be needed and we might want to reconsider) and probably *not* on local VMs
- Data Administrators: consider allowing local sudo rights - thoughts on this?

The problem here is that local `sudo` opens up a *lot* of possible attacks (eg. password snooping of other users). If we allow researchers local `sudo` we should also enforce that System Administrators *never* log on to the research VMs.

edwardchalstrey1:
@jemrobinson @JimMadge I have given one of the DSG PIs sudo rights (because tier 0 so was suggested this is ok), which I'm hesitant to take away mid-project. However, I used the method in top comment of #1396 to do this, so I guess he would keep privileges if we took him out the group anyway?

Looks like there aren't any others - what would you like me to do?

edwardchalstrey1:
> Data Administrators: consider allowing local sudo rights - thoughts on this?

An intermediate privilege tier seems sensible - here is the request that the DSG PI made as an example:

> can you try and install python3 pyqt5 using following command in the TRE?
> sudo apt-get install python3-pyqt5
> I think we might be able to reduce compute by calling qgis from python but I still haven't had any luck getting qgis to be successfully imported. The problem seems to be getting pyqt5 available and I'm hoping that having a different install approach might work (edited) 

So actually in this case it probably would have been fine to make sure that I myself could run sudo commands, but my approach was to give him them too in case he wanted to do anything else similar and not wait for me to respond

JimMadge:
>     * System Administrators: admin rights on the Guacamole user console (although this shouldn't ever be needed and we might want to reconsider) and probably _not_ on local VMs
> 
>     * Data Administrators: consider allowing local sudo rights - thoughts on this?

If System Administrator isn't necessary, then I think we should get rid of it (or at least repurpose it as sudo users).
I think it would be better not to give Data Administrators two purposes. A user can be a member of multiple groups afterall.

> The problem here is that local `sudo` opens up a _lot_ of possible attacks (eg. password snooping of other users). If we allow researchers local `sudo` we should also enforce that System Administrators _never_ log on to the research VMs.

I don't think I understand what the attack is here.
I was seeing this as a Tier0/1 feature where there is already open communication to the internet, so the worst a user could do is break their SRD (although, maybe we shouldn't invite that either...).
Could go down the route of allowing the group to only execute certain commands as root `apt` for example.
Although, then you are kind of shifting the responsibility to those programs to not have arbitrary code execution exploits or a way to fork a new process as root.

jemrobinson:
> I don't think I understand what the attack is here.

Since `/home` is stored on Azure Files, there is a lot of network traffic involved in syncing files. This includes all command history (which might contain passwords). Anyone who can snoop this traffic (which our security audit confirmed requires root access) can potentially see the passwords of other users of the VM. If a user with System Administrator rights gets their password exposed in this way then local root can be used to gain System Administrator privileges.

jemrobinson:
> @jemrobinson @JimMadge I have given one of the DSG PIs sudo rights (because tier 0 so was suggested this is ok), which I'm hesitant to take away mid-project. However, I used the method in top comment of #1396 to do this, so I guess he would keep privileges if we took him out the group anyway?

Leaving sudo rights (but removing them from the group) is probably OK in this instance. We should bear in mind that this user can potentially read the passwords of other users of the SRE (including any with System Administrator rights if you've logged in with such a user). They can also read passwords for accounts that are able to:
- register new VMs in the local AD
- read user details for accounts on the local AD
- access the data ingress and data egress volumes (ie. they will be able to read these volumes from outside the SRE)

They can also make destructive changes like:
- pointing all PyPI requests to their own malicious server
- install malware (eg. cryptomining or password sniffing software)

This is definitely not something that we should be doing "because the user wants to" or "because it's easier" and definitely not in a live production environment without previous testing.

@JimMadge: I think it's possible to set up permissions that are able to eg. `apt-install` without being able to access files that are only readable by root. Is that correct?

JimMadge:
> This includes all command history (which might contain passwords).

Are there examples of that? I can't think of why you would put your user password in a command.
You _could_ but it feels a bit like saying you could write your password in a plain text file (although probably more likely, by mistake in particular).

If root on an SRD can really see plain text credentials of so many users then I don't think we can allow it in any case (this would also give away to compromise accounts with SHM or cross SRE permissions?). But also I think that makes the security of the whole system pretty poor. We will always be one privilege escalation exploit away from complete compromise.

> @JimMadge: I think it's possible to set up permissions that are able to eg. apt-install without being able to access files that are only readable by root. Is that correct?

More or less. Like I said above you can restrict what commands a user can run (and as who).
_e.g._ `%group ALL=(root:root) apt` would allow members of `group` to run `apt` as `root`.
The catch is, if the command you allow gives a way to fork a new process (either intensionally or through and exploit) then it is still a way to run more (or any) commands as root.

edwardchalstrey1:
Based on the comments above, I suggest that we do one of the following:

1. Merge the change and have a documented policy that only System Managers are ever added to this SG, and state in the documentation that this enables them to run sudo whilst in the SRE
2. Don't merge the change and instead add documentation that says something like: "To make any changes to the SRE that require root privileges, the System Manager must log in to the Serial Console of the compute VM in Azure Portal..."

In both cases should there be a disclaimer in the System Manager docs like: "Installing things as root after deployment is not generally recommended except for known edge cases/problems such as X, Y and Z"

JimMadge:
@edwardchalstrey1 I would say go with option 2 (and revert the change to the docs where we said the System Administrators group gives sudo). Long term, this could be a good feature, but we will need some time (and to be careful and thorough) to understand the permissions this group has and if sudo in a low-sensitivity environment could compromise other environments or the SHM.

edwardchalstrey1:
@JimMadge see #1407

PR #1412, opened 2023-03-03
Add automated updates to Pulumi

### LINKED ISSUE #1405: Pulumi: SHM update automation

### LINKED ISSUE #1413: Pulumi: SHM log analytics

jemrobinson:

- Add SHM LogAnalytics workspace
- Add VM extension agent to Windows and Linux VMs
- Add update automation for Windows and Linux VMs
- Add schedules for Windows (SHM only) and Linux (one for SHM and one for each SRE)

Closes #1405. Closes #1413 

### :microscope: Tests

Tested on a fresh deployment


data_safe_haven/pulumi/components/shm_monitoring.py
```diff
@@ -101,32 +122,290 @@ def __init__(
...
-            else []
+        # Add a private DNS record for each automation custom DNS config
+        automation_account_private_dns_zone_group = network.PrivateDnsZoneGroup(
+            f"{self._name}_automation_account_private_dns_zone_group",
+            private_dns_zone_configs=[
+                network.PrivateDnsZoneConfigArgs(
+                    name=replace_separators(f"{stack_name}-aa-to-{dns_zone_name}", "-"),
+                    # private_dns_zone_id=Output.from_input(props.dns_resource_group_name).apply(
+                    #     lambda rg_name: network.get_private_zone(private_zone_name=f"privatelink.{dns_zone_name}", resource_group_name=rg_name).id
+                    # ),
```
JimMadge:
```suggestion
```

data_safe_haven/pulumi/components/shm_monitoring.py
```diff
@@ -101,32 +122,290 @@ def __init__(
...
+        subscription_resource_id = resource_group.id.apply(
+            lambda id_: id_.split("/resourceGroups/")[0]
+        )
+        # Create Windows VM virus definitions update schedule: daily at 01:01
+        schedule_windows_definitions = automation.SoftwareUpdateConfigurationByName(
+            f"{self._name}_schedule_windows_definitions",
+            automation_account_name=automation_account.name,
+            resource_group_name=resource_group.name,
+            schedule_info=automation.SUCSchedulePropertiesArgs(
+                expiry_time="9999-12-31T23:59:00+00:00",
```
JimMadge, line 274: The Y10K bug! 🐛

data_safe_haven/pulumi/components/sre_research_desktop.py
```diff
@@ -119,7 +123,9 @@ def __init__(
             vm_names=props.vm_names,
             vm_sizes=props.vm_sizes,
         )
-        # Note that deploying inside an apply is advised against but not forbidden
+        # Note that creating resources inside an .apply() is discouraged but not
+        # forbidden. This is the one way to create one resource for each entry in
+        # an Output[Sequence]. See https://github.com/pulumi/pulumi/issues/3849.
```
JimMadge, line 128: Given this, I think it is worth asking. Can we apply any 'above' specifications and then use the 'real' values of `vm_details` to create the SRDs in a different Pulumi project?

Why is `vm_details` `Output[T]`?

JimMadge (changes requested):
Looks good. Just one suggestion and a question.

data_safe_haven/pulumi/components/sre_research_desktop.py
jemrobinson: Yes, we could create the SRD VMs on a separate stack using the outputs of the SRE stack. This might be worth considering.

> Why is vm_details Output[T]?

Because it's the result of an `Output.all` which is necessary because we want to combine three `Output[T]`s into a single `Output` which is necessary because we don't know the appropriate IP addresses until we've deployed the VNet (in principle we could hardcode them but this would also be fragile).

JimMadge: > Yes, we could create the SRD VMs on a separate stack using the outputs of the SRE stack. This might be worth considering.

Hopefully not starting a semantic argument, I think we would need a separate project (which I suppose would by definition also be a different stack).

> Because it's the result of an Output.all which is necessary because we want to combine three Output[T]s into a single Output which is necessary because we don't know the appropriate IP addresses until we've deployed the VNet (in principle we could hardcode them but this would also be fragile).

Yes I don't think hardcoding is the way to go. It also looses a lot of the benefit IAC affords us.

Sorry, I think this is retredding old ground but I have forgotten. Is there a reason we can't represent the properties we need to create the SRDs as one or more `Collection[Output[T]]`s rather than (what I'm assuming are something like) `Output[List[T]]`? That way you could iterate over the outputs rather than the promise in an output.

I'm not sure if this is the same but I do something a _bit_ like that [here](https://github.com/alan-turing-institute/bureau/blob/9a5e7fe00207e6a68cfddb0961877d451f16ac0e/build/bureaubuild/infrastructure.py#L164) and [here](https://github.com/alan-turing-institute/bureau/blob/9a5e7fe00207e6a68cfddb0961877d451f16ac0e/build/bureaubuild/infrastructure.py#L177).

Have you seen if creating the SRDS within an apply creates problems for pulumi destroy?

jemrobinson: We start with:

- `vm_ip_addresses: Output[List[str]]`
- `vm_names` and `vm_sizes` also `Output[List[str]]` which are extracted from an `Input[Dict[str, Dict[str, str]]]`

I'm not aware of any way to convert `Output[List[str]]` to `List[Output[str]]` since all the functions that let you perform operations on an `Output` also wrap their return values in an `Output`.

If I'm understanding correctly, the way that `bureau` works involves constructing a `Dict[str, Props]` which is an interesting idea. I think it works because you know the number of items you are creating in advance (which we potentially do here, although they're currently wrapped in `Input`s)

JimMadge: > I'm not aware of any way to convert Output[List[str]] to List[Output[str]] since all the functions that let you perform operations on an Output also wrap their return values in an Output.

Yes sorry, I didn't mean that and I don't think it is possible. But what I'm suggesting is tricky; going backwards through the code and finding where `[Output[List[str]]` is created and seeing if we could instead create a `List[Output[str]]`.

I think what I am struggling to see is, if these are just lists of properties of some number of VMs. Where is it ever necessary to have an `Output[List[T]]` rather than a `List[Output[T]]`?

I suppose in that file from `bureau` `vms` is a `Dict[str, BuildVM]`. And each `BuildVM` has `Output`s like IP addresses. But I can iterate over the dict values without using `apply`.

> I think it works because you know the number of items you are creating in advance (which we potentially do here, although they're currently wrapped in `Input`s)

That might be the difference. `vms` is made from iterating over a dict which is known at 'compile-time'. If that dict was passed as an input instead, only known at run-time, would it need to be an `Input[T]`?

jemrobinson: Captured this in #1415.

PR #1416, opened 2023-03-06
Refactor SRD creation

### LINKED ISSUE #1415: Pulumi: SRE SRDs
These are already working but can we create them outside an `apply()` function?

> Given this, I think it is worth asking. Can we apply any 'above' specifications and then use the 'real' values of vm_details to create the SRDs in a different Pulumi project?
>
> Sorry, I think this is retredding old ground but I have forgotten. Is there a reason we can't represent the properties we need to create the SRDs as one or more Collection[Output[T]]s rather than (what I'm assuming are something like) Output[List[T]]? That way you could iterate over the outputs rather than the promise in an output.
> 
> @[JimMadge](https://github.com/JimMadge) JimMadge [Mar 6, 2023](https://github.com/alan-turing-institute/data-safe-haven/pull/1412#discussion_r1126475832)

jemrobinson:

Moves SRD creation into a Python loop rather than an `apply()` function. Note that this *requires* that the `vm_details` argument is provided as `T` not an `Input[T]` as would usually be done.

Closes #1415

### :microscope: Tests

Tested in a local deployment

### DIFF
--- a/data_safe_haven/provisioning/sre_provisioning_manager.py
+++ b/data_safe_haven/provisioning/sre_provisioning_manager.py
@@ -51,7 +51,8 @@ class SREProvisioningManager(LoggingMixin):
         # Construct VM parameters
         self.research_desktops = {}
         for vm in sre_stack.output("research_desktops")["vm_outputs"]:
-            self.research_desktops[vm["name"]] = {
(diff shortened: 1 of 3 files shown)


JimMadge (approved):
That looks a lot neater to my eye, also reduces the line count :tada:.

This should also help prevent any problems when destroying stacks.

PR #1453, opened 2023-04-27
Migrate docs to readthedocs.io

JimMadge:

This PR makes adjustments so that documentation can be (automatically) build and deployed on readthedocs.io.

If I recall correctly, we had wanted to do this previously but couldn't as the repository was private. Read the Docs supports building/displaying documentation from multiple versions (assuming the tags are compatible) and producing pdf documentation. Both things which we have otherwise felt the need to engineer/maintain ourselves.

It also supports automatically building preview docs for pull requests. This PR is [here](https://data-safe-haven--1453.org.readthedocs.build/en/1453/).

I think this is worth doing, and less painful the earlier we do it.
The question is, do we host the 'old' docs somewhere else (leave them on GitHub pages?) and link to them, or try and move old tags after making changes so that older versions build?

#### Other changes

- Move to [lychee](https://lychee.cli.rs/#/?id=main) for link checking
- Use pip-compile for reproducible documentation builds

### :microscope: Tests

Deploying this branch.

### Todo

- [x] plan for old versions
- [x] update html proofing and/or link checking CI


harisood:
> The question is, do we host the 'old' docs somewhere else (leave them on GitHub pages?) and link to them, or try and move old tags after making changes so that older versions build?

Which one is more of a pain?

JimMadge:
> Which one is more of a pain?

Definitely the later. Also, moving tags feels like bad manners. People don't expect "version 4.0.0" to change after it has been released.

craddm (commented):
New docs look good.

Just to be clear, is what's going to happen that the new docs will be hosted on readthedocs, rather than supplanting the docs on Github Pages? I'm not a huge fan of leaving the vestigial docs going on another URL, at least not without some clear way of notifying people that these are outdated docs and a simple way of pointing them at the new ones.

Take the point about moving tags being bad manners, but OTOH there aren't many old versions, so if there's a good time to be rude, it might be now.

JimMadge:
> Just to be clear, is what's going to happen that the new docs will be hosted on readthedocs, rather than supplanting the docs on Github Pages? 

I think that is up for debate. But I would like to say yes. I would really rather avoid the mess of making modifications to (multiple) old branches to get the readthedocs builds working. Ditto to moving tags.

> I'm not a huge fan of leaving the vestigial docs going on another URL, at least not without some clear way of notifying people that these are outdated docs and a simple way of pointing them at the new ones.

Yes, we should avoid that.

I think some reasonable options are,

- Add a message to all old docs pages about where newer/current docs can be found (_i.e._ add something to the header template)
- New tags for old versions like `v4.0.0_rtd` which have **only** the changes required for readthedocs builds (that way _all_ docs are in one place)

Thoughts, other ideas @harisood @craddm @KirstieJane

edwardchalstrey1:
There are a few places where the docs are linked from that we'll need to update too:
- TRESA docs
- Links in the [various repos](https://github.com/alan-turing-institute/data-safe-haven-team#useful-links)

@JimMadge Please can I request that at the very least the GitHub pages site is left up until the end of the DSG (May 26)?

Perhaps having an overlap where both are live for a couple of weeks might be helpful anyway in case there are any issues in the readthedocs build we haven't spotted (although it looks good to me, and I assume there shouldn't be any rendering issues as both use sphinx)

harisood:
> People don't expect "version 4.0.0" to change after it has been released.

Agreed for the core infrastructure code, but tbh a change that allows migration of documentation to me doesn't feel too rude 😅

> Add a message to all old docs pages about where newer/current docs can be found (i.e. add something to the header template)

This feels like a good minimum solution. Not sure if GH pages has the functionality where we can redirect any rendering of the old docs to a root page that says something like this (i.e. warning, you are about to enter old documentation etc...) that people have to explicitly accept before seeing? Only concern is when you're rushing through docs do you acknowledge the old version warning.

Maybe we do this and see if we run into any trouble?

Also thinking as long as we update all public links to the new docs, and only refer to the old docs from the new docs (e.g. for older versions see these docs), we minimise the amount of people getting there accidentally

craddm:
Tend to agree with Hari that it's less of an issue when it's really just docs that change. Do think the suggestion of having slightly differently tagged versions with just the rtd changes is a nice compromise.

The header warning is _probably_ also sufficient.

harisood:
> The header warning is probably also sufficient.

What is the impact of someone using the old docs? Is it more functionality won't work (relatively small impact, they'll probs j reach out and we can redirect them) or something more serious like security issues (high impact, we should make this as difficult to do as possible)

JimMadge:
> What is the impact of someone using the old docs? Is it more functionality won't work (relatively small impact, they'll probs j reach out and we can redirect them) or something more serious like security issues (high impact, we should make this as difficult to do as possible)

Depends what version they are using and what they are doing. If someone is following docs for a version they are not using, I suppose that is user error, but we certainly shouldn't do anything to make that likely.

Modifying old commits to be compatible with readthedocs, although it will involve some work, does have the advantage that _all_ docs will be in one place. We can make the old location either 404 or redirect to the new docs so there is much less risk of a mistake.

I'm still opposed to moving existing tags. That said, we don't actually want to move tags, we just want those docs to be deployed to readthedocs. So I think new tags which indicate clearly those old version numbers is the better approach.

I think I'm leaning towards this as a solution. Avoids the mess of managing two locations.

harisood:
If it's not a massive time sink trade-off I'd agree

JimMadge:
> If it's not a massive time sink trade-off I'd agree

It would involve making (hopefully) the same set of changes at four points. I think I might be able to script it to make it easier.

If we can agree this is the way to go I should be able to do this next week.

JimMadge:
v4.0.0 docs [here](https://data-safe-haven.readthedocs.io/en/v4.0.0_rtd).

jemrobinson (approved):
Haven't reviewed this in detail but I'm happy after the discussion we had in today's meeting.

PR #1472, opened 2023-05-23
Possible Azure Device Health Service endpoints

jemrobinson:

I ran a scan over domains matching some known ADHS patterns and found some additional ones of the same type.

From [the ServiceBus docs](https://learn.microsoft.com/en-us/azure/service-bus-messaging/service-bus-quickstart-portal#create-a-namespace-in-the-azure-portal), Microsoft reserves names `ending with "-sb" or "-mgmt"` but these do not.

It's therefore not clear whether these service bus endpoints (and blob storage accounts) were registered by Microsoft or by a (possibly malicious) third party.

They might be required for AzureADConnect to work correctly or they might be irrelevant. Thoughts on what to do here are welcome. 

n/a

### :microscope: Tests

Not tested


JimMadge:
Looking at #1485, is the difference here that these endpoints don't match the wildcard `*-sb.servicebus.windows.net`?

> They might be required for AzureADConnect to work correctly or they might be irrelevant.

I feel like we should be sure. Although, I remember we agreed before that being permissive with the rules here wouldn't have a big impact on security.

jemrobinson:
> Looking at #1485, is the difference here that these endpoints don't match the wildcard `*-sb.servicebus.windows.net`?

Yes, exactly. They're not one of the names that are reserved by Microsoft so in principle anyone could have registered them.

> Although, I remember we agreed before that being permissive with the rules here wouldn't have a big impact on security.

Yes, this would only be allowing outwards connections from the domain controller, so nothing that's able to be controlled by a malicious user unless:

- they have compromised the domain controller
- they have pre-emptively registered one of those storage account names in the hope that we would allowlist it

which feels like a low-risk scenario. The other option is to wait until the domain controller tries and fails to connect to any particular endpoint (possibly during a time-critical event like a DSG) and then decide that we should add it.

JimMadge:
OK, in that case, I'm not concerned about adding the FQDNs.

The only thing that bothers me is having to maintain the list (although it is much smaller than before!) but that might be unavoidable.

> I ran a scan over domains matching some known ADHS patterns and found some additional ones of the same type.

How does that work?

Have we (or should we) open a ticket with Microsoft to ask about these endpoints?

jemrobinson:
>> I ran a scan over domains matching some known ADHS patterns and found some additional ones of the same type.
>
> How does that work?

It was just a brute-force `nslookup adhsprod${CODE}ehsyncia.servicebus.windows.net"` and `nslookup adhssyncprodpks${CODE}` for the full set of Azure region codes. As I said above - no guarantees that these are Microsoft endpoints.

> Have we (or should we) open a ticket with Microsoft to ask about these endpoints?

That's a good idea. I'm not sure where the best place to ask would be though.

jemrobinson:
@JimMadge : Was this superseded by #1745?

JimMadge:
Yes, superseded by #1745

PR #1500, opened 2023-06-28
Automate user synchronisation

### LINKED ISSUE #1498: Pulumi: SRE user synchronisation
At present users need to be explicitly synchronised from the SHM domain controller to the SRE Guacamole database. We should automate this.

jemrobinson:

- Create a Docker container to synchronise Guacamole users against Active Directory

Closes #1498 

### :microscope: Tests

Tested on a local Pulumi deployment


data_safe_haven/administration/users/active_directory_users.py
```diff
@@ -110,3 +110,14 @@ def register(self, sre_name: str, usernames: Sequence[str]) -> None:
         )
         for line in output.split("\n"):
             self.parse_as_log(line)
+
+    def remove(self, users: Sequence[ResearchUser]) -> None:
+        """Remove list of users from local Active Directory"""
+        pass
```
JimMadge:
```suggestion
```

data_safe_haven/administration/users/guacamole_users.py
```diff
@@ -33,49 +29,18 @@ def __init__(self, config: Config, sre_name: str, *args: Any, **kwargs: Any):
...
-                for user in users_to_add
-            ],
-            "group_name": "Research Users",
-        }
-        if user_data:
-            self.postgres_provisioner.execute_scripts(
-                [self.postgres_script_path / "add_users.mustache.sql"],
-                mustache_values=user_data,
-            )
+        pass
```
JimMadge:
```suggestion
```

data_safe_haven/administration/users/guacamole_users.py
```diff
@@ -33,49 +29,18 @@ def __init__(self, config: Config, sre_name: str, *args: Any, **kwargs: Any):
             / "postgresql"
         )
         self.sre_name = sre_name
+        self.group_name = f"Data Safe Haven Users SRE {sre_name}"
 
     def add(self, users: Sequence[ResearchUser]) -> None:
         """Add sequence of users to Guacamole"""
```
JimMadge:
```suggestion
```

data_safe_haven/administration/users/guacamole_users.py
```diff
@@ -90,43 +55,8 @@ def list(self) -> Sequence[ResearchUser]:
 
     def remove(self, users: Sequence[ResearchUser]) -> None:
         """Remove list of users from Guacamole"""
```
JimMadge:
```suggestion
```

data_safe_haven/administration/users/guacamole_users.py
```diff
@@ -90,43 +55,8 @@ def list(self) -> Sequence[ResearchUser]:
...
-            if self.users_:
-                self.users_ = [user for user in self.users_ if user not in users]
-        except Exception as exc:
-            raise DataSafeHavenInputException(
-                f"Could not update SRE '{self.sre_name}' database users.\n{str(exc)}"
-            ) from exc
-        finally:
-            if sql_file_name:
-                pathlib.Path(sql_file_name).unlink()
+        pass
```
JimMadge:
```suggestion
```

data_safe_haven/administration/users/guacamole_users.py
```diff
@@ -90,43 +55,8 @@ def list(self) -> Sequence[ResearchUser]:
...
 
     def set(self, users: Sequence[ResearchUser]) -> None:
         """Set Guacamole users to specified list"""
```
JimMadge:
```suggestion
```

data_safe_haven/administration/users/guacamole_users.py
```diff
@@ -90,43 +55,8 @@ def list(self) -> Sequence[ResearchUser]:
...
-        users_to_remove = [user for user in self.list() if user not in users]
-        self.remove(users_to_remove)
-        users_to_add = [user for user in users if user not in self.list()]
-        self.add(users_to_add)
+        pass
```
JimMadge:
```suggestion
```

data_safe_haven/pulumi/components/sre_gitea_server.py
```diff
@@ -353,11 +353,13 @@ def __init__(
...
             opts=ResourceOptions.merge(
                 ResourceOptions(
+                    delete_before_replace=True,
                     depends_on=[
                         file_share_gitea_caddy_caddyfile,
                         file_share_gitea_gitea_configure_sh,
                         file_share_gitea_gitea_entrypoint_sh,
-                    ]
+                    ],
+                    replace_on_changes=["containers"],
```
JimMadge, line 362: This marks that resources must be replaced when another is updated/changed?

Is "containers" a resource name or a tag?

JimMadge (changes requested):
It is a little hard to tell what is happening here.

Maybe because actions are being moved to [here](https://github.com/alan-turing-institute/guacamole-user-sync/tree/main). We should be clearer when we depend on external resources and what they do (also with the nexus allowlist).

In general, I'm getting a bit concerned we are moving to another code base where only one person knows how it works, without documentation and comments 😬.

How often does the synchronisation happen or how is it triggered? [Every 5 minutes](https://github.com/alan-turing-institute/guacamole-user-sync/blob/main/synchronise/run)?

I've never liked the way that the AD users are injected into the Guacmole DB, but maybe this isn't the time to solve that. (It would be easier if we used LDAP).

data_safe_haven/administration/users/active_directory_users.py
jemrobinson: Sorry - I have implementations for these stubs but they aren't quite working atm.

data_safe_haven/administration/users/guacamole_users.py
jemrobinson: Sorry - I have implementations for these stubs but they aren't quite working atm.

data_safe_haven/administration/users/guacamole_users.py
jemrobinson: Sorry - I have implementations for these stubs but they aren't quite working atm.

data_safe_haven/administration/users/guacamole_users.py
jemrobinson: Sorry - I have implementations for these stubs but they aren't quite working atm.

data_safe_haven/administration/users/guacamole_users.py
jemrobinson: Sorry - I have implementations for these stubs but they aren't quite working atm.

data_safe_haven/administration/users/guacamole_users.py
jemrobinson: Sorry - I have implementations for these stubs but they aren't quite working atm.

data_safe_haven/administration/users/guacamole_users.py
jemrobinson: Sorry - I have implementations for these stubs but they aren't quite working atm.

data_safe_haven/pulumi/components/sre_gitea_server.py
jemrobinson: It means that the `ContainerInstance` resource needs to be replaced whenever there is a change to the `containers` attribute. Without this, `pulumi` will try to do an update-in-place which isn't supported for this resource.

jemrobinson:
> How often does the synchronisation happen or how is it triggered? [Every 5 minutes](https://github.com/alan-turing-institute/guacamole-user-sync/blob/main/synchronise/run)?

Yes. Happy to make this configurable though.

> I've never liked the way that the AD users are injected into the Guacmole DB, but maybe this isn't the time to solve that. (It would be easier if we used LDAP).

The only places where Guacamole can store the maps of users <-> allowed connections are in a database or in an LDAP folder where you control the structure. I'm not sure whether we can alter the AD LDAP structure in the right way and I'm also not sure that we want a public-internet facing app to have write access to our AD anyway.

With those restrictions I think we're stuck with the database version for the moment.

JimMadge:
> Yes. Happy to make this configurable though.

It would make sense to have that as an environment variable for the container.

> I'm not sure whether we can alter the AD LDAP structure in the right way and 

Yes I remember that problem from before. Why I'm keen on moving to something like OpenLDAP.

> I'm also not sure that we want a public-internet facing app to have write access to our AD anyway.

I don't think Guacamole would need write access. My interpretation of [the docs](https://guacamole.apache.org/doc/gug/ldap-auth.html#how-guacamole-uses-ldap) is that each users credentials are used to query the LDAP (or, you can nominate a different account which you could restrict with whatever permissions you like).

jemrobinson:
> Yes I remember that problem from before. Why I'm keen on moving to something like OpenLDAP.

The problem with OpenLDAP was self-service password reset. We haven't found a self-hosted solution that we're happy with but we can't connect OpenLDAP to AzureAD and have password writeback.

jemrobinson:
> I don't think Guacamole would need write access. My interpretation of [the docs](https://guacamole.apache.org/doc/gug/ldap-auth.html#how-guacamole-uses-ldap) is that each users credentials are used to query the LDAP (or, you can nominate a different account which you could restrict with whatever permissions you like).

Ah OK, so we'd need to add connections to the LDAP server separately. At the moment, I find programmatically interacting with the domain controller *more* annoying than doing the same thing with a standalone database, but we can certainly take a look at this once Pulumi migration is finished.

JimMadge:
> Ah OK, so we'd need to add connections to the LDAP server separately.

Yes that's how I understand it.

I think it is designed 'for the enterprise' where LDAP is going to be the canonical source for information about users, so admins are probably used to (and well equipped to) add custom schemas and data.

Pretty easy to do with OpenLDAP. But, yes, might be difficult/impossible with AAD and there are blockers to moving away from that. Something for later.

jemrobinson:
> In general, I'm getting a bit concerned we are moving to another code base where only one person knows how it works, without documentation and comments 😬.

What's the best way to make this more developer-friendly? There are certainly places where the structure isn't ideal and getting some additional eyes on it (you and @craddm?) would definitely be helpful. We're nearly at feature-parity so it's a good time to refactor.

JimMadge:
> What's the best way to make this more developer-friendly?

That is a good question, not an easy one to answer. I think there is probably a challenge in balancing getting this done in a timely way and ending up in the ideal place.

I think sometimes this work has been merged without review because it isn't going to the latest or develop branch and one change needs to be merged before working on the next.

As a lot of this is figuring out how/if things can work with Python/Pulumi I'm more sympathetic to just writing the code and putting in a bit fixup effort later. Especially as we will need to go through some sort of security testing/assurance anyway. The downsides are,

- We will end up with *a lot* of documentation to write all at once
- There is a strong risk we loose track of part of code that were committed as 'fine for now'

Some specific things that I think would make it more developer friendly,

- Developer documentation
- Explanation of the choices make
- Architecture/map of the code (which components are where, where do we use external repos/packages and what those are)

data_safe_haven/administration/users/guacamole_users.py
```diff
@@ -33,49 +29,18 @@ def __init__(self, config: Config, sre_name: str, *args: Any, **kwargs: Any):
...
-                for user in users_to_add
-            ],
-            "group_name": "Research Users",
-        }
-        if user_data:
-            self.postgres_provisioner.execute_scripts(
-                [self.postgres_script_path / "add_users.mustache.sql"],
-                mustache_values=user_data,
-            )
+        raise NotImplementedError("Users should not be directly added to Guacamole.")
```
JimMadge: Why leave these methods in if we will not use them?

`NotImplementedError` is probably the wrong exception here

> Note: It should not be used to indicate that an operator or method is not meant to be supported at all – in that case either leave the operator / method undefined or, if a subclass, set it to [None](https://docs.python.org/3/library/constants.html#None)

[docs](https://docs.python.org/3/library/exceptions.html#NotImplementedError)

data_safe_haven/administration/users/guacamole_users.py
```diff
@@ -90,43 +55,10 @@ def list(self) -> Sequence[ResearchUser]:
...
-        except Exception as exc:
-            raise DataSafeHavenInputException(
-                f"Could not update SRE '{self.sre_name}' database users.\n{str(exc)}"
-            ) from exc
-        finally:
-            if sql_file_name:
-                pathlib.Path(sql_file_name).unlink()
+        raise NotImplementedError(
+            "Users should not be directly removed from Guacamole."
+        )
```
JimMadge: Same as above

data_safe_haven/administration/users/guacamole_users.py
```diff
@@ -90,43 +55,10 @@ def list(self) -> Sequence[ResearchUser]:
...
 
     def set(self, users: Sequence[ResearchUser]) -> None:
         """Set Guacamole users to specified list"""
-        users_to_remove = [user for user in self.list() if user not in users]
-        self.remove(users_to_remove)
-        users_to_add = [user for user in users if user not in self.list()]
-        self.add(users_to_add)
+        raise NotImplementedError("Users should not be directly altered in Guacamole.")
```
JimMadge: Same as above.

data_safe_haven/administration/users/user_handler.py
```diff
@@ -163,36 +156,60 @@ def set(self, users_csv_path: str) -> None:
...
+                for required_field in ["GivenName", "Surname", "Phone", "Email"]:
+                    if (not reader.fieldnames) or (
+                        required_field not in reader.fieldnames
+                    ):
+                        raise ValueError(
+                            f"Missing required CSV field '{required_field}'."
+                        )
+                desired_users = [
+                    ResearchUser(
+                        country="GB",
```
JimMadge, line 171: Should this be hardcoded?

Is this a limitation on phone numbers we support?

JimMadge (changes requested):
Looking good (problems with the architecture as a whole notwithstanding 🙂).

data_safe_haven/administration/users/guacamole_users.py
jemrobinson: Fixed

data_safe_haven/administration/users/guacamole_users.py
jemrobinson: Fixed

data_safe_haven/administration/users/guacamole_users.py
jemrobinson: Fixed

data_safe_haven/administration/users/user_handler.py
jemrobinson: Fixed

JimMadge (approved):

PR #1501, opened 2023-06-29
Pin container images

JimMadge:

Pin unpinned container image version. In particular nexus-allowlist which was using the 'main' tag that could point to unstable/broken builds.

This is also consistent with other container instance specifications in the project.

I think there is a discussion around what level to pin to. I have tried to pin to major versions where possible as this a avoids known, breaking changes being introduced.
Pinning to minor or patch versions gives more confidence that there will be no regressions, however it adds more burden to update versions more regularly.


jemrobinson (approved):
LGTM

PR #1502, opened 2023-07-07
Switch CLI interface to Typer

jemrobinson:

- Switch from `clikit` to `rich` for logging
- Switch from an inherited logging mixin to a component logging singleton
- Switch from `cleo` to `typer` for CLI
- Rename `'users'` command to more generic `'admin'`
- Switch to reusable parameter validation functions
- Raise an exception rather than prompting when encountering missing values
- Separate `helpers` into `functions` (for functions) and `utility` (for utility classes)

n/a

### :microscope: Tests

Tested locally


data_safe_haven/commands/base_command.py
```diff
@@ -0,0 +1,38 @@
...
+class CommandGroup(typer.Typer):
+    """Command group containing subcommands"""
+
+    def subgroup(self, cls: type["CommandGroup"], name: str, help: str) -> None:
+        """Register a command group"""
+        self.add_typer(cls(), name=name, help=help)
+
+    def subcommand(self, cls: type[BaseCommand], name: str, help: str) -> None:
+        """Register a command"""
+        self.command(name=name, help=help)(cls().entrypoint)
```
JimMadge: Is there an advantage to doing this over using decorators like in the [Typer docs](https://typer.tiangolo.com/tutorial/subcommands/add-typer/#put-them-together)?

For what it is worth, I think using the decorators is clearer and reduces the code we need to maintain.

data_safe_haven/utility/logger.py
```diff
@@ -0,0 +1,234 @@
...
+    Logging singleton that can be used for anything needing logging
+    """
+
+    date_fmt = r"%Y-%m-%d %H:%M:%S"
+    rich_format = r"[log.time]%(asctime)s[/] [%(levelname)8s] %(message)s"
+    _instance: Optional["Logger"] = None
+
+    def __new__(
+        cls, verbosity: Optional[int] = None, log_file: Optional[PathType] = None
+    ) -> "Logger":
```
JimMadge, line 104: I think it might be clearer just to create an instance of this class and import it where needed rather than instantiating it elsewhere in the code. Particularly because there is no `__init__` method, it is just a collection of class data and functions?

data_safe_haven/utility/logger.py
```diff
@@ -0,0 +1,234 @@
...
+        elif level == "ERROR":
+            return self.error(remainder)
+        elif level == "WARNING":
+            return self.warning(remainder)
+        elif level == "INFO":
+            return self.info(remainder)
+        elif level == "DEBUG":
+            return self.debug(remainder)
+        else:
+            return self.info(message.strip())
```
JimMadge, line 220: What is this applied to? I can see cases where it is run on the output of Azure CLI/API. Is there a more specific name for the method?

JimMadge (changes requested):
I took a skim over.

There are few places where I think the code can be clearer.
In general I think this looks good though. I'm more comfortable with Typer and bringing in rich seems good as that is a very mature and feature-full package.

data_safe_haven/commands/base_command.py
jemrobinson: You need to use `add_typer` for subgroups anyway. This is just to unify the behaviour between subgroups and subcommands, so that the top level command can look like:

```
application.subgroup(
    TeardownCommand,
    name="teardown",
    help="Tear down a Data Safe Haven component.",
)

# Register direct subcommands
application.subcommand(
    InitialiseCommand,
    name="init",
    help="Initialise a Data Safe Haven deployment.",
)
```

rather than

```
application.add_typer(
    TeardownCommand(),
    name="teardown",
    help="Tear down a Data Safe Haven component.",
)

@application.command()
def init(*args, **kwargs):
    """Initialise a Data Safe Haven deployment."""
    InitialiseCommand.entrypoint(*args, **kwargs)
```

which I think looks messier.

data_safe_haven/utility/logger.py
jemrobinson: So you mean actually having a `logger = Logger()` line in `logger.py` and then having other classes do `from logger import logger`?

We still need singleton behaviour of some kind because there's exactly one place (in the CLI callback function) where we have the information about verbosity and output file and we need to configure the logger there. We can't be sure whether or not it's already been instantiated at that point.

data_safe_haven/utility/logger.py
jemrobinson: I thought `logger.parse` -> parses log messages was reasonably clear, but I may have been too terse here! What would you call a function that is processing plain-text messages and emitting an appropriate log message?

JimMadge: Oh yes, I didn't quite mean that.

The method assumes things about the form of the input, that it will be `<level>:<message>` and that level is one of a set of values (otherwise we assume the level is info).

It felt like maybe this method was to parse some particular log output rather than log messages from _any_ program.

data_safe_haven/utility/logger.py
JimMadge: Yes that was I was suggesting.

Good point about the configuration though.
It looks like the logging module already provides a way to do this though,

[docs](https://docs.python.org/3/library/logging.html?highlight=logger#logger-objects)

> Loggers have the following attributes and methods. Note that Loggers should NEVER be instantiated directly, but always through the module-level function logging.getLogger(name). Multiple calls to [getLogger()](https://docs.python.org/3/library/logging.html?highlight=logger#logging.getLogger) with the same name will always return a reference to the same Logger object.

[Example using the rich handler](https://rich.readthedocs.io/en/stable/logging.html#handle-exceptions)

data_safe_haven/commands/base_command.py
JimMadge: Is this because the subcommands are custom classes rather than `Typer.Typer` instances? Is that necessary?

What I'm imagining is more like the Typer examples,

in `__main.py__` (or whatever our entrypoint is)

```python
import teardown
import admin

app = Typer.Typer()

app.add_typer(teardown.app, name='teardown')
app.add_typer(admin.app, name='admin')
```

and in `admin.py`

```python
app = Typer.Typer()

@app.command()
def add_user(...):
  ...
```

It is more idiomatic and will be easier to understand for people who have used Typer.

data_safe_haven/utility/logger.py
jemrobinson: That's true, but you still need to ensure that `logging.basicConfig` is called after processing the verbosity and output command line arguments but before any `logging.getLogger` calls. Using a singleton logger means you can put this stuff in the constructor which gets run whenever the logger is first initialised.

Also, we don't just want a `logging.Logger` object (which is what `getLogger` returns), we want a custom object that has additional methods like `parse`, `tabulate` etc.

data_safe_haven/commands/base_command.py
jemrobinson: That part is fine, but not all of the top-level commands are groups of commands, so you'll still need something like

```
import teardown
import admin
from initialise_command import entrypoint
from data_safe_haven.functions import validate_aad_guid

app = Typer.Typer()

app.add_typer(teardown.app, name='teardown')
app.add_typer(admin.app, name='admin')

@app.command(help="Initialise a Data Safe Haven deployment.")
def init(admin_group: Annotated[
            Optional[str],
            typer.Option(
                "--admin-group",
                "-a",
                help="The ID of an Azure group containing all administrators.",
                callback=validate_aad_guid,
            ),
        ] = None,
        location: Annotated[
            Optional[str],
            typer.Option(
                "--location",
                "-l",
                help="The Azure location to deploy resources into.",
            ),
        ] = None,
        name: Annotated[
            Optional[str],
            typer.Option(
                "--deployment-name",
                "-d",
                help="The name to give this Data Safe Haven deployment.",
                callback=validate_aad_guid,
            ),
        ] = None,
        subscription: Annotated[
            Optional[str],
            typer.Option(
                "--subscription",
                "-s",
                help="The name of an Azure subscription to deploy resources into.",
            ),
        ] = None,):
    entrypoint(admin_group, location, name, subscription)
```

which I did start off with, but I thought it made things messier than necessary.

data_safe_haven/utility/logger.py
jemrobinson: It's to parse log messages that we produce from systems that can't directly talk to our logging code (eg. when we run external commands on a VM and get stdout back but we want to present it as log messages).

data_safe_haven/commands/base_command.py
JimMadge: I (strongly) prefer this, the hierarchy is clear.

Could always put `def init(...)` in another file and import the `typer.Typer` instance there if it is about keeping files small and 'on topic'.

data_safe_haven/utility/logger.py
JimMadge: Do we know that those will all give log messages that play nicely with this function?

I can see the need. Maybe just adding a docstring to say what this function is for is the best thing to do for now.

jemrobinson: To be clear: this is to parse messages from **non-Python** code in the DSH codebase.

data_safe_haven/utility/logger.py
JimMadge: OK, still don't feel great about writing our own singleton implementation for the sake of a few methods.

Let's not block here though.

jemrobinson: Sure. Feel free to open an issue for this but we'll leave it alone in this PR.

data_safe_haven/cli.py
```diff
@@ -65,25 +63,25 @@ def main() -> None:
...
         help="Tear down a Data Safe Haven component.",
     )
 
     # Register direct subcommands
-    application.subcommand(
-        InitialiseCommand,
-        name="init",
-        help="Initialise a Data Safe Haven deployment.",
+    application.command(name="init", help="Initialise a Data Safe Haven deployment.")(
+        initialise_command
```
JimMadge, line 84: Might be clearer to put the `def` here with the decorator.

But I won't die on that hill. Keeping the files short/tidy and on topic is also good :+1:.

JimMadge (commented):
I think this is looking a lot better.

One thing I'm not sure about,

The Typer commands are all (or almost all) just wrappers for a classes `__call__` method. It feels like unecessary boiler plate when the contents of the `__call__` method could just be in the functions decorated with `Typer.command`. Especially as the `__init__` methods of the classes I remember only create a reference to the logger singleton.

I'm thinking,

- Is there a strong advantage to having commands as classes
- If not
  - Can we remove that pattern now or,
  - Open an issue to remove it in the future to let this merge now?

jemrobinson:
> One thing I'm not sure about,
>
>     * Is there a strong advantage to having commands as classes
>     * If not
>       * Can we remove that pattern now or,
>       * Open an issue to remove it in the future to let this merge now?

In general I dislike importing functions (rather than classes) from other files although I don't mind doing it where necessary. There *are* a couple of places where the class does a bit more (eg. `DeploySHMCommand` in `deploy_shm_command.py` has an `update_config` method which changes the config object based on input values). However, arguably this should be a method of the config object so I'm happy to change that.

JimMadge:
Hmm, might be descending into a OOP vs procedural programming argument :smile:.

I suppose my position is, classes with no real state, that we instantiate once, and a `__call__` method are just functions with extra text.

```Python
class MyCommand():
   def __init__(self):
       logger = Logger()

   def __call__(*args):
      self.logger('hello')
      ...

app = typer.Typer()

@app.command()
def mycommand(*args):
   MyCommand()(*args)
```

versus

```Python
logger = Logger()
app = typer.Typer()

@app.command()
def my_command(*args):
   logger('hello')
   ...
```

If `Logger` is a singleton, we can just create it wherever it is needed.

> However, arguably this should be a method of the config object so I'm happy to change that.

That does sound more sensible (without having a proper look).

What would you like to do? 'Fix' these things here, open an issue, don't 'fix' them?
(Conscious that this might be blocking further progress)

jemrobinson:
> > However, arguably this should be a method of the config object so I'm happy to change that.
> 
> That does sound more sensible (without having a proper look).
> 
> What would you like to do? 'Fix' these things here, open an issue, don't 'fix' them? (Conscious that this might be blocking further progress)

I'll fix this one here.

> If `Logger` is a singleton, we can just create it wherever it is needed.

I've had a look at a better way of handling logging but it's slightly complicated by the following:
- you should get a `logging.Logger` from `logging.getLogger`
- `logging.getLogger` will return a `logging.Logger` (even if this is actually a custom subclass) which messes up type-hinting/linting when you call `MyLoggingClass.my_method()`

I've got some ideas, but not sure whether they'll make it into this PR.

jemrobinson:
@JimMadge Actually, on second thoughts. Can we merge this now and I'll open a couple of new PRs for these things? I also have a PR about refactoring the config object which will be affected by these changes.

JimMadge (approved):

PR #1506, opened 2023-07-17
Pulumi: Migrate pyproject to PyPA standard and hatch

JimMadge:

Migrate to hatch/hatchling.

- Use hatchling for build backend
- Use hatch for development
- Use standards PEP508, PEP440, PEP621
- Demonstrate using hatch environments for, linting, testing, building, _etc._

**Check #1510 before merging**


JimMadge:
@jemrobinson This is not ready for review yet but would like your thoughts.

Especially on how to merge this (if we want to) without disrupting other work on the pulumi branch.

data_safe_haven/__about__.py
```diff
@@ -0,0 +1 @@
+__version__ = "v0.4.0"
```
jemrobinson: At the moment this is in `pyproject.toml`. Is that a standard location? If we also need it in `__about__.py` we can do

```
import pkg_resources
__version__ = pkg_resources.get_distribution("data-safe-haven").version
```

pyproject.toml
```diff
@@ -1,57 +1,135 @@
...
+dynamic = ["version"]
 description = "An open-source framework for creating secure environments to analyse sensitive data."
-authors = ["Data Safe Haven development team <safehavendevs@turing.ac.uk>"]
+authors = [
+  { name = "Data Safe Haven development team", email = "safehavendevs@turing.ac.uk" },
+]
+requires-python = ">=3.10"
+license = "BSD-3-Clause"
+dependencies = [
+  "azure-core~=1.26.0",
```
jemrobinson, line 16: Is `~=` equivalent to the `^=` that poetry uses?

pyproject.toml
```diff
@@ -1,57 +1,135 @@
...
+  "PLR",
+  "PLW",
+  "Q",
+  "RUF",
+  "S",
+  "T",
+  "TID",
+  "UP",
+  "W",
+  "YTT",
```
jemrobinson: Can you add comments about what these mean?

pyproject.toml
```diff
@@ -1,57 +1,135 @@
...
+]
+ignore = [
+  # Allow non-abstract empty methods in abstract base classes
+  "B027",
```
jemrobinson: Do we need this?

pyproject.toml
```diff
@@ -1,57 +1,135 @@
...
+  # Allow boolean positional values in function calls, like `dict.get(... True)`
+  "FBT003",
```
jemrobinson: Do we need this?

jemrobinson (changes requested):
Mostly happy but a few questions on details. Also, we should check for compatibility with #1509 which adds some dependencies.

JimMadge:
> Mostly happy but a few questions on details. Also, we should check for compatibility with https://github.com/alan-turing-institute/data-safe-haven/issues/1509 which adds some dependencies.

:+1: Can we agree to hold work on the `python-migration` branch after #1509 merges to let me get linting passing on this branch and merge it?

jemrobinson:
I think this and #1509 can be merged in either order. Just need to be aware that they are inter-dependent.

JimMadge:
No, because I'll need to fix all the black and ruff errors to not break CI in main. That will involve touching (potentially) every Python file in the repo.

That's why I want other work on this branch to stop before I start fixing the linting errors. Resolving merge conflicts will be too difficult otherwise.

data_safe_haven/__about__.py
jemrobinson: See eg (https://stackoverflow.com/questions/458550/standard-way-to-embed-version-into-python-package) and (https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#declaring-project-metadata) .

JimMadge: In `pyproject.toml` the version is now dynamic metadata

https://github.com/alan-turing-institute/data-safe-haven/blob/f01446eabc6237930d10e49f61d59d1687887e9d/pyproject.toml#L7

sourced from `__about__.py`

https://github.com/alan-turing-institute/data-safe-haven/blob/f01446eabc6237930d10e49f61d59d1687887e9d/pyproject.toml#L52-L53

So it is only defined in one place (and you can use hatch to update the version).
This is [#1 of the options here](https://packaging.python.org/en/latest/guides/single-sourcing-package-version/), just for hatch rather than setuptools.

I think the things you have linked won't work as they rely on setuptools, which isn't being used here.

JimMadge: We can import (or move `version` to `__init__.py`, maybe renamed `__version__`) to get the value in the code.

pyproject.toml
JimMadge: Yes. Even better, it is the [PEP440](https://peps.python.org/pep-0440/#version-specifiers) and [PEP508](https://peps.python.org/pep-0508/) way of specifying versions/requirements.

data_safe_haven/__about__.py
jemrobinson: Looks like a relatively common standard is to have `version.py` that contains:

```
__version_info__ = ('1', '2', '3')
__version__ = '.'.join(__version_info__)
```

then `__init__.py` has `from .version import __version__, __version_info__`.

Presumably this will still work with `pyproject.toml` dynamic while providing the PEP-recommended `<module>.__version__` string? As far as I can tell, `__about__.py` is not specified anywhere.

pyproject.toml
JimMadge: I don't think so, let's remove it.

JimMadge: Done in b22356de

pyproject.toml
JimMadge: There are quite a few instances of this rule being broken.

I think there is a good argument to follow this rule. `do_something(..., True)` is not clear and could mean `do_something(..., verbose=True)` or `(do_something(..., dry_run=True)`.
There is a rule to stop _us_ writing functions with position boolean arguments.

pyproject.toml
JimMadge: Done in be27d032.

These are inherited from Hatch's opinionated, sensible defaults.
We may want to remove some, but I've found there are some handy additions like more strictly enforcing conventions, forcing the use of more modern patterns/features.

data_safe_haven/__about__.py
JimMadge: Fixed in 18cba63d

pyproject.toml
JimMadge: Removed (at least for now) in f8a5e6f9

data_safe_haven/administration/users/__init__.py
```diff
@@ -1,3 +1,3 @@
-from .user_handler import UserHandler
+from data_safe_haven.administration.users.user_handler import UserHandler
```
jemrobinson: Why is this type of import better? We're in `__init__.py` here and defining what should be exported from this module. If anything *isn't* in this directory that's a problem, and I feel like this syntax makes it harder to verify that.

data_safe_haven/administration/users/active_directory_users.py
```diff
@@ -1,15 +1,17 @@
 """Interact with users in an Azure Active Directory"""
 # Standard library imports
 import pathlib
-from typing import Any, Optional, Sequence
+from collections.abc import Sequence
+from typing import Any
+
+from data_safe_haven.administration.users.research_user import ResearchUser
```
jemrobinson: This seems to be in the wrong order. If we want to separate same-package imports from same-folder imports, then same-folder ones should come afterwards.

data_safe_haven/administration/users/active_directory_users.py
```diff
@@ -70,11 +66,9 @@ def add(self, new_users: Sequence[ResearchUser]) -> None:
         for line in output.split("\n"):
             self.logger.parse(line)
 
-    def list(self, sre_name: Optional[str] = None) -> Sequence[ResearchUser]:
+    def list(self, sre_name: str | None = None) -> Sequence[ResearchUser]:
```
jemrobinson: I prefer `Optional[type]` to `type | None` in function signatures because it correctly describes the parameter as being `optional`.

data_safe_haven/administration/users/azure_ad_users.py
```diff
@@ -63,11 +57,7 @@ def list(self) -> Sequence[ResearchUser]:
                 account_enabled=user_details["accountEnabled"],
                 email_address=user_details["mail"],
                 given_name=user_details["givenName"],
-                phone_number=(
-                    user_details["businessPhones"][0]
-                    if len(user_details["businessPhones"])
-                    else None
-                ),
+                phone_number=(user_details["businessPhones"][0] if len(user_details["businessPhones"]) else None),
```
jemrobinson: Note that lots of these changes will be reverted by `black`

jemrobinson (changes requested):
`Typer` requires `Optional[type]` rather than `type | None`. Otherwise you get the message:

```
RuntimeError: Type not yet supported: str | None
```

data_safe_haven/pulumi/dynamic/ssl_certificate.py
```diff
@@ -115,14 +114,17 @@ def create(self, props: dict[str, Any]) -> CreateResult:
             )
             outs["secret_id"] = kvcert.secret_id
         except Exception as exc:
-            msg = f"Failed to create SSL certificate [green]{props['certificate_secret_name']}[/] for [green]{props['domain_name']}[/].\n{exc!s}"
+            msg = (
```
jemrobinson, line 125: This might be clearer as `" ".join([...])` rather than needing to remember to leave whitespace in the strings.

data_safe_haven/pulumi/dynamic/ssl_certificate.py
```diff
@@ -138,12 +140,14 @@ def delete(self, id_: str, props: dict[str, Any]) -> None:
                 key_vault_name=props["key_vault_name"],
             )
         except Exception as exc:
-            msg = f"Failed to delete SSL certificate [green]{props['certificate_secret_name']}[/] for [green]{props['domain_name']}[/].\n{exc!s}"
+            msg = (
+                f"Failed to delete SSL certificate [green]{props['certificate_secret_name']}[/]"
```
jemrobinson, line 154: E.g. see here where necessary whitespace is missing

data_safe_haven/pulumi/pulumi_stack.py
```diff
@@ -198,17 +198,15 @@ def login(self) -> None:
...
                     cwd=self.work_dir,
-                    stdout=subprocess.PIPE,
-                    stderr=subprocess.STDOUT,
+                    capture_output=True,
                     encoding="UTF-8",
-                ) as process:
-                    if process.stdout:
-                        self.logger.info(process.stdout.readline().strip())
+                )
+                process.check_returncode()
```
jemrobinson: Add a comment that this raises a `CalledProcessError` for a non-zero return code.

data_safe_haven/pulumi/dynamic/remote_powershell.py
```diff
@@ -50,13 +50,8 @@ def create(self, props: dict[str, Any]) -> CreateResult:
             outs=outs,
         )
 
-    def delete(self, id_: str, props: dict[str, Any]) -> None:
-        """The Python SDK does not support configuration deletion"""
-        return
```
jemrobinson: We should include at least a comment explaining why `delete` is not needed here (probably because you can't delete the fact that a script has run). Also check whether `DshResourceProvider` requires this method to exist or not.

data_safe_haven/pulumi/dynamic/dsh_resource_provider.py
```diff
@@ -44,26 +44,21 @@ def partial_diff(
     def refresh(props: dict[str, Any]) -> dict[str, Any]:
         return dict(**props)
 
-    def check(self, old_props: dict[str, Any], new_props: dict[str, Any]) -> CheckResult:
```
jemrobinson: These signatures can't be changed as they're inherited from `pulumi.dynamic.ResourceProvider`.

data_safe_haven/pulumi/components/sre_hedgedoc_server.py
```diff
@@ -258,7 +258,10 @@ def __init__(
                         ),
                         containerinstance.EnvironmentVariableArgs(
                             name="CMD_LDAP_SEARCHFILTER",
-                            value=f"(&(objectClass=user)(memberOf=CN={props.ldap_user_security_group_name},OU=Data Safe Haven Security Groups,{props.ldap_root_dn})(sAMAccountName={{{{username}}}}))",
+                            value=(
+                                f"(&(objectClass=user)(memberOf=CN={props.ldap_user_security_group_name},"
```
jemrobinson: This is a really bad string to break up. Brackets, punctuation etc. are all important and need to be easily verified.

Can you break it up into logical pieces?

I'd suggest something like:

```
"(&"
"(objectClass=user)"
f"(memberOf=CN={props.ldap_user_security_group_name},OU=Data Safe Haven Security Groups,{props.ldap_root_dn})"
f"(sAMAccountName={{{{username}}}})"
")"
```

data_safe_haven/utility/types.py
```diff
@@ -1,6 +1,6 @@
 # Standard library imports
 from pathlib import Path
-from typing import Any, Union
+from typing import Any
 
-PathType = Union[str, Path]
+PathType = str | Path
 YamlType = dict[str, dict[str, Any]]
```
jemrobinson: I think this type might be wrong. Maybe it should be `dict[str, Any | dict[str, Any]]`? Not sure how helpful that is though.

data_safe_haven/pulumi/components/sre_data.py
```diff
@@ -355,7 +355,9 @@ def __init__(
                 acl_user="rwx",
                 acl_group="rwx",
                 acl_other="rwx",
-                apply_default_permissions=False,  # due to an Azure bug this also gives ownership of the fileshare to user 65533 (preventing use inside the SRE)
+                # due to an Azure bug this also gives ownership of the
```
jemrobinson: It needs to be clear that this comment applies to `apply_default_permissions` not to a) the function in general or b) `acl_other`. I think it's easier to see this as an in-line comment.

JimMadge:
@jemrobinson Some things I found

- I've ignored [A003](https://beta.ruff.rs/docs/rules/builtin-attribute-shadowing/) where breaking the rule improves clarity. I would be happy to remove it altogether. Shadowing builtin names is bad, but when they are methods I think you are unlikely to confused them `set(abc)` vs `myclass.set(abc)`.
- I've ignored an error raised by a `print` call inline. Feels like it is a useful failure as there shouldn't be _too many_ print statements in the code (we may add them for some quick and dirty debugging).
- fea84c13 Ignored but I don't like this code. What is 6 here?
- _I_ would prefer a shorted line length. 80-90 seems good. Some of the long lines are difficult to read.

data_safe_haven/pulumi/components/sre_data.py
JimMadge: :+1: I would say it is clear because it appears in the argument list before the argument. The same way we comment what a line does before the line.

I would rather not have it as an inline comment as it is quite long _and_ on an indented line (I tend to avoid inline comments entirely).
We could change the text to make it more clear what the comment refers to?

data_safe_haven/pulumi/dynamic/ssl_certificate.py
JimMadge: I disagree. If we use `" ".join` we can _only_ split on spaces.

I think string literal concatenation is idiomatic and efficient (concatenated at compile time rather than run time).

data_safe_haven/administration/users/__init__.py
JimMadge: That is [this rule](https://beta.ruff.rs/docs/rules/relative-imports/). I like it in general. This is a case where it looks odd though.

We can [configure it to allow relative imports of siblings](https://beta.ruff.rs/docs/settings/#flake8-tidy-imports-ban-relative-imports).

jemrobinson:
> @jemrobinson Some things I found
> 
>     * I've ignored [A003](https://beta.ruff.rs/docs/rules/builtin-attribute-shadowing/) where breaking the rule improves clarity. I would be happy to remove it altogether. Shadowing builtin names is bad, but when they are methods I think you are unlikely to confused them `set(abc)` vs `myclass.set(abc)`.

Happy to change these method names, but maybe in a separate PR for clarity?
 
>     * I've ignored an error raised by a `print` call inline. Feels like it is a useful failure as there shouldn't be _too many_ print statements in the code (we may add them for some quick and dirty debugging).

I don't think there's a better way to write the version number to the console than this. I guess we could use `sys.stdout.write` but that feels silly.

>     * [fea84c1](https://github.com/alan-turing-institute/data-safe-haven/commit/fea84c13d4a93164ef8116ad7a6636e792f572ad) Ignored but I don't like this code. What is 6 here?

We are expecting there to be six tokens for the following line to work. Happy to replace this with a `try/except` if you'd prefer.

```
ResearchUser(
                        email_address=tokens[4],
                        given_name=tokens[1],
                        phone_number=tokens[3],
                        sam_account_name=tokens[0],
                        surname=tokens[2],
                        user_principal_name=tokens[5],
)
```

> I would prefer a shorted line length. 80-90 seems good. Some of the long lines are difficult to read.

I don't mind longer lines as I use a wide screen and frequently full-screen my code window, but I note that `black` uses 88. If you want to default to this (ie. don't set `line-length` in our `black` config) that's fine with me.

data_safe_haven/administration/users/__init__.py
jemrobinson: Happy if you prefer it. I'll let you decide.

data_safe_haven/administration/users/active_directory_users.py
JimMadge: I think this happened because of the comment near the other imports and the adding of the absolute path.

If relative imports are allowed the order can be like before

```Python
import pathlib
from collections.abc import Sequence
from typing import Any

from data_safe_haven.config import Config
from data_safe_haven.external import AzureApi
from data_safe_haven.functions import b64encode
from data_safe_haven.pulumi import PulumiSHMStack
from data_safe_haven.utility import FileReader, Logger

from .research_user import ResearchUser
```

data_safe_haven/administration/users/user_handler.py
```diff
@@ -81,98 +74,88 @@ def get_usernames(self) -> Dict[str, List[str]]:
...
                 user_data.append(user_memberships)
 
             # Write user information as a table
             for line in self.logger.tabulate(user_headers, user_data):
                 self.logger.info(line)
         except Exception as exc:
-            raise DataSafeHavenUserHandlingException(
-                f"Could not list users.\n{str(exc)}"
-            ) from exc
+            msg = f"Could not list users.\n{exc!s}"
```
jemrobinson: `exc!s` is definitely valid, but it is clearer/more common than `str(exc)`? No problem with either, but I want to make sure this is a deliberate choice.

data_safe_haven/backend/__init__.py
```diff
@@ -1,4 +1,4 @@
-from .backend import Backend
+from data_safe_haven.backend.backend import Backend
```
jemrobinson: Again, I find this import style overly verbose, but if you prefer it that's fine.

data_safe_haven/backend/backend.py
```diff
@@ -59,9 +57,8 @@ def create(self) -> None:
                 tags=self.tags,
             )
             if not storage_account.name:
-                raise DataSafeHavenAzureException(
-                    f"Storage account '{self.config.backend.storage_account_name}' was not created."
-                )
+                msg = f"Storage account '{self.config.backend.storage_account_name}' was not created."
+                raise DataSafeHavenAzureError(msg)
```
jemrobinson, line 59: I'm not sure I find this clearer. We create a variable that is only used on one line (the following one) rather than using brackets and a line break. Is it the indentation that's the issue here?

data_safe_haven/backend/backend.py
```diff
@@ -81,32 +78,25 @@ def create(self) -> None:
...
+                msg = f"Keyvault '{self.config.backend.key_vault_name}' was not created."
+                raise DataSafeHavenAzureError(msg)
             pulumi_encryption_key = self.azure_api.ensure_keyvault_key(
                 key_name=self.config.pulumi.encryption_key_name,
                 key_vault_name=keyvault.name,
             )
-            self.config.pulumi.encryption_key_id = pulumi_encryption_key.id.split("/")[
-                -1
-            ]
+            self.config.pulumi.encryption_key_id = pulumi_encryption_key.id.split("/")[-1]
```
jemrobinson: I prefer this, but `black` imposed the other syntax. What setting have you changed to make sure it won't do so if I run it again?

data_safe_haven/cli.py
```diff
@@ -14,19 +14,17 @@
...
-        Optional[pathlib.Path],
-        typer.Option(
-            "--output", "-o", resolve_path=True, help="Path to an output log file"
-        ),
+        pathlib.Path | None,
+        typer.Option("--output", "-o", resolve_path=True, help="Path to an output log file"),
     ] = None,
     verbosity: Annotated[
-        Optional[int],
+        int | None,
```
jemrobinson: Typer requires `Optional`

data_safe_haven/commands/users_unregister_command.py
```diff
@@ -1,12 +1,11 @@
...
 # Standard library imports
-from typing import List
 
 # Local imports
 from data_safe_haven.administration.users import UserHandler
 from data_safe_haven.config import Config
 from data_safe_haven.exceptions import (
-    DataSafeHavenException,
+    DataSafeHavenError,
 )
```
jemrobinson: This should be on one line.

data_safe_haven/config/config.py
```diff
@@ -43,29 +43,25 @@ def validate(self) -> None:
         try:
             validate_aad_guid(self.admin_group_id)
         except Exception as exc:
-            raise ValueError(
-                f"Invalid value for 'admin_group_id' ({self.admin_group_id}).\n{str(exc)}"
-            )
+            msg = f"Invalid value for 'admin_group_id' ({self.admin_group_id}).\n{exc!s}"
+            raise ValueError(msg) from exc
```
jemrobinson, line 45: Should we make these `DataSafeHavenError`s? Maybe `DataSafeHavenInputError` or `DataSafeHavenParameterError`?

data_safe_haven/exceptions/__init__.py
```diff
@@ -1,50 +1,50 @@
...
 
 
-class DataSafeHavenMicrosoftGraphException(DataSafeHavenAzureException):
+class DataSafeHavenMicrosoftGraphError(DataSafeHavenAzureError):
     pass
 
 
-class DataSafeHavenPulumiException(DataSafeHavenCloudException):
+class DataSafeHavenPulumiError(DataSafeHavenCloudError):
     pass
```
jemrobinson, line 50: Worth checking whether all these errors are currently used in the code. Some may be unnecessary at the moment.

data_safe_haven/external/api/azure_api.py
```diff
@@ -28,38 +29,43 @@
...
-from azure.mgmt.keyvault.models import AccessPolicyEntry, Permissions
-from azure.mgmt.keyvault.models import Sku as KeyVaultSku
 from azure.mgmt.keyvault.models import (
+    AccessPolicyEntry,
+    Permissions,
     Vault,
     VaultCreateOrUpdateParameters,
     VaultProperties,
 )
+from azure.mgmt.keyvault.models import Sku as KeyVaultSku
```
jemrobinson: These could/should be in one include block. What's the reason for separating them?

data_safe_haven/external/api/azure_api.py
```diff
@@ -28,38 +29,43 @@
...
     StorageAccountCreateParameters,
     StorageAccountKey,
     StorageAccountListKeysResult,
 )
+from azure.mgmt.storage.models import (
+    Kind as StorageAccountKind,
+)
+from azure.mgmt.storage.models import (
+    Sku as StorageAccountSku,
+)
```
jemrobinson: These could/should be in one include block. What's the reason for separating them?

data_safe_haven/functions/strings.py
```diff
@@ -47,20 +48,15 @@ def random_letters(length: int) -> str:
...
+    return input_string.replace(" ", separator).replace("_", separator).replace("-", separator).replace(".", separator)
 
 
 def sha256hash(input_string: str) -> str:
     """Return the SHA256 hash of a string as a string."""
     return hashlib.sha256(str.encode(input_string, encoding="utf-8")).hexdigest()
 
 
-def truncate_tokens(tokens: Sequence[str], max_length: int) -> List[str]:
+def truncate_tokens(tokens: Sequence[str], max_length: int) -> list[str]:
```
jemrobinson, line 65:
```suggestion
def truncate_tokens(tokens: Sequence[str], max_length: int) -> list[str]:
    """
    Recursively remove the final character from the longest strings in the input.
    Terminate when the total length of all strings is no greater than max_length.
    For example:
        truncate_tokens(["the", "quick", "fox"], 6) -> ["th", "qu", "fo"]
    """
```

data_safe_haven/pulumi/common/transformations.py
```diff
@@ -1,17 +1,16 @@
 """Common transformations needed when manipulating Pulumi resources"""
 # Standard library imports
```
jemrobinson: Delete this line if there are no stdlib imports

data_safe_haven/pulumi/components/shm_firewall.py
```diff
@@ -1,6 +1,5 @@
 """Pulumi component for SHM traffic routing"""
 # Standard library import
```
jemrobinson: Delete this line.

data_safe_haven/pulumi/components/shm_firewall.py
```diff
@@ -49,15 +44,16 @@ def __init__(
...
         props: SHMFirewallProps,
-        opts: Optional[ResourceOptions] = None,
+        opts: ResourceOptions | None = None,
     ):
         super().__init__("dsh:shm:FirewallComponent", name, {}, opts)
         child_opts = ResourceOptions.merge(ResourceOptions(parent=self), opts)
 
         # Important IP addresses
-        external_dns_resolver = "168.63.129.16"  # https://docs.microsoft.com/en-us/azure/virtual-network/what-is-ip-address-168-63-129-16
+        external_dns_resolver = (
```
jemrobinson: Move comment out of inline?

data_safe_haven/pulumi/components/sre_monitoring.py
```diff
@@ -1,6 +1,5 @@
 """Pulumi component for SHM monitoring"""
 # Standard library import
```
jemrobinson: Delete this line.

data_safe_haven/utility/logger.py
```diff
@@ -67,7 +65,7 @@ class LogLevelHighlighter(RegexHighlighter):
     """
 
     base_style = "logging.level."
-    highlights = [
+    highlights: ClassVar[list[str]] = [
```
jemrobinson, line 68: This should be a tuple instead of a list. That should avoid the need for `ClassVar` here.

jemrobinson (changes requested):
- Import ordering is broken: imports from local folder should come after first party imports (see eg. https://pycqa.github.io/isort/docs/configuration/custom_sections_and_ordering.html)
- In several places you've changed syntax/linebreaks that were previously imposed by black. Have you run `black` recently? Does it make any changes?
- Typer needs `Optional[type]` instead of `type | None`. I also think this form looks better in function signatures where you're indicated that an argument is `optional`.

data_safe_haven/administration/users/user_handler.py
JimMadge: This is actually new to me. So, I kind of dislike it because it looks unusual. However, it is shorter and part of the f-string grammar. I would say keep the rule.

(Also, I wonder if `f"{str(a)}"` actually calls `str()` twice. Once to evaluate the f expression then again for the conversion.)

The rule is [this one](https://beta.ruff.rs/docs/rules/explicit-f-string-type-conversion/).

I think all of these are unnecessary though as the default is to use `str()`. So I think we should remove all of these.

JimMadge: Fixed in ef21e9a6

data_safe_haven/administration/users/__init__.py
JimMadge: Let's allow it. I don't see a strong argument to always require absolute imports.

(If we find one, it is easy to change the configuration and let ruff update them all).

JimMadge: Updated in 741290b0

data_safe_haven/administration/users/active_directory_users.py
JimMadge: Fixed in 741290b0

data_safe_haven/pulumi/dynamic/ssl_certificate.py
JimMadge: Fixed in 19bad5a47040fd59008289caa49c40cd31bacd8f

data_safe_haven/backend/__init__.py
JimMadge: Fixed in 741290b0bbc51de7d7222022485b16ee35b5566e

data_safe_haven/administration/users/active_directory_users.py
JimMadge: [The rule](https://beta.ruff.rs/docs/rules/non-pep604-annotation/) here is enforcing using the new union syntax rather than `typing.Union` or `typing.Optional`. **I don't think it will distinguish when arguments are optional**.

I would go with the rule (apart from where it breaks Typer) because

- It is more concise
- It is more clear that the type is actually a union of T and None
- An `Optional[T]` is not the same as an optional argument, _e.g._

```Python
def f(x: Optional[int]) -> None:
    if x is None:
        do_something()
    else:
        do_something_else()
```
x is `Optional[int]` but it is not an optional argument.

jemrobinson: I mean that I like the use of `Optional[type]` to indicate that you *do have* an optional argument. It's not a hill I would die on though!

data_safe_haven/backend/backend.py
JimMadge: [Keeps tracebacks easy (at least easier) to read](https://beta.ruff.rs/docs/rules/f-string-in-exception/)

That is a new one to me :smile:. Sounds worth keeping.

data_safe_haven/backend/backend.py
JimMadge: Not certain what would be different between our configurations. The black rules are in `pyproject.toml` so everyone _should_ be able to arrive at the same result by running `hatch run lint:fmt`.

data_safe_haven/config/config.py
JimMadge: I think any of those will work. The correct answer is probably what is most useful for the user.

I would say this is out of scope for this PR though. Can you open an issue?

data_safe_haven/exceptions/__init__.py
JimMadge: Definitely worth doing. Do you know if there is a tool to check for unused exceptions?

Again, feels out of scope for the PR?

jemrobinson: I think a quick "grep" will do the job here.

data_safe_haven/external/api/azure_api.py
JimMadge: You tell me 2143eec87 :wink:.

A little disapointed the autoformatters don't concatenate these.

JimMadge: Looks like it is the isort style. I suppose separating aliased imports.

JimMadge: Could change this behaviour with [combine-as-imports](https://beta.ruff.rs/docs/settings/#isort-combine-as-imports). Do you have a strong opinion?

I don't mind too much but would lean towards the default for simplicity's sake.

data_safe_haven/external/api/azure_api.py
JimMadge: Same as above.

data_safe_haven/pulumi/components/shm_firewall.py
JimMadge: Fixed in f6ff62e5.

data_safe_haven/backend/backend.py
jemrobinson: This is such a minor issue that I'm not sure I would have bothered putting it into a syntax checker! I don't think it's *wrong* though, so I'm happy to keep it.

data_safe_haven/config/config.py
jemrobinson: Opened #1515

data_safe_haven/external/api/azure_api.py
jemrobinson: I'd like to see these combined if there's a set of arguments that will make this happen.

data_safe_haven/cli.py
JimMadge: Fixed in 40ae88b9

data_safe_haven/administration/users/active_directory_users.py
JimMadge: Yep. The linter doesn't distinguish between using `typing.Optional` for optional arguments and in other places. It just prefers the newer union operator syntax. So I don't think there is much hope of enforcing that.

I'll keep the rule here apart from exclusions for Typer.

data_safe_haven/pulumi/pulumi_stack.py
JimMadge: Fixed in 4058ee0b

data_safe_haven/pulumi/dynamic/ssl_certificate.py
jemrobinson: Fair enough. There are a couple of places where you lost whitespace while (manually?) doing the splitting. Can you check?

data_safe_haven/pulumi/components/sre_hedgedoc_server.py
jemrobinson: Or maybe:

```
security_groups_ou = f"Data Safe Haven Security Groups,{props.ldap_root_dn}"
user_security_group_cn = f"{props.ldap_user_security_group_name},OU={security_groups_ou}"
value=f"(&(objectClass=user)(memberOf=CN={user_security_group_cn})(sAMAccountName={{{{username}}}}))"
```

data_safe_haven/pulumi/dynamic/remote_powershell.py
JimMadge: I wasn't sure exactly the best thing to do here.

- It was clear that delete was not valid
- The parent class `DshResourceProvider` is not an `ABC` and the `delete` method is not an `abstractmethod`. So there is no error if it is not defined when creating a child of `DshResourceProvider`. ([see here](https://docs.python.org/3/library/abc.html?highlight=abc))

Having a closer look though, `DshResourceProvider`'s `update` method does call `delete`

https://github.com/alan-turing-institute/data-safe-haven/blob/4058ee0b00c168da98ede70ebaeadf36b457696e/data_safe_haven/pulumi/dynamic/dsh_resource_provider.py#L67-L76

I can't find this method every being called though.

So, maybe we should do one of

1. Redefine `update` in `RemoteScriptProvider` (seems like not a great idea).
2. Make the appropriate methods of `DshResourceProvider` abstract methods and add a comment to say what to do if `delete` is meaningless.

What do you think @jemrobinson?

In any case, I think this should be a new issue/PR.
I can revert the changes and ignore in CI.

data_safe_haven/administration/users/user_handler.py
```diff
@@ -62,7 +62,7 @@ def add(self, users_csv_path: pathlib.Path) -> None:
             # Commit changes
             self.active_directory_users.add(users)
         except Exception as exc:
-            msg = f"Could not add users from '{users_csv_path}'.\n{exc!s}"
```
jemrobinson: Won't you get `exception.__repr__` instead of `exception.__str__` by doing this?

data_safe_haven/pulumi/components/sre_hedgedoc_server.py
JimMadge: Fixed in e856efa9

data_safe_haven/pulumi/components/sre_data.py
JimMadge: Fixed in 362eafde

data_safe_haven/external/api/azure_api.py
JimMadge: Fixed in 6a73875b

data_safe_haven/administration/users/user_handler.py
JimMadge: Nope, the default conversion is `str()`.

data_safe_haven/pulumi/dynamic/remote_powershell.py
jemrobinson: Captured in #1516

data_safe_haven/pulumi/dynamic/azuread_application.py
```diff
@@ -61,31 +57,28 @@ def create(self, props: Dict[str, Any]) -> CreateResult:
...
-            ) from exc
+            msg = f"Failed to create application [green]{props['application_name']}[/] in AzureAD.\n{exc}"
+            raise DataSafeHavenMicrosoftGraphError(msg) from exc
         return CreateResult(
             f"AzureADApplication-{props['application_name']}",
             outs=outs,
         )
 
-    def delete(self, id_: str, props: Dict[str, Any]) -> None:
+    def delete(self, props: dict[str, Any]) -> None:
```
jemrobinson:
```suggestion
    def delete(self, id_: str, props: dict[str, Any]) -> None:
```

data_safe_haven/pulumi/dynamic/azuread_application.py
```diff
@@ -61,31 +57,28 @@ def create(self, props: Dict[str, Any]) -> CreateResult:
...
-            ) from exc
+            msg = f"Failed to delete application [green]{props['application_name']}[/] from AzureAD.\n{exc}"
+            raise DataSafeHavenMicrosoftGraphError(msg) from exc
 
     def diff(
         self,
-        id_: str,
-        old_props: Dict[str, Any],
-        new_props: Dict[str, Any],
+        old_props: dict[str, Any],
```
jemrobinson, line 85:
```suggestion
        id_: str,
        old_props: dict[str, Any],
```

data_safe_haven/pulumi/dynamic/blob_container_acl.py
```diff
@@ -63,15 +62,14 @@ def create(self, props: Dict[str, Any]) -> CreateResult:
...
-            ) from exc
+            msg = f"Failed to set ACLs on storage account [green]{props['storage_account_name']}[/].\n{exc}"
+            raise DataSafeHavenPulumiError(msg) from exc
         return CreateResult(
             f"BlobContainerAcl-{props['container_name']}",
             outs=outs,
         )
 
-    def delete(self, id_: str, props: Dict[str, Any]) -> None:
+    def delete(self, props: dict[str, Any]) -> None:
```
jemrobinson:
```suggestion
    def delete(self, id_: str, props: dict[str, Any]) -> None:
```

data_safe_haven/pulumi/dynamic/blob_container_acl.py
```diff
@@ -82,16 +80,13 @@ def delete(self, id_: str, props: Dict[str, Any]) -> None:
...
-        return
+            msg = f"Failed to delete custom ACLs on storage account [green]{props['storage_account_name']}[/].\n{exc}"
+            raise DataSafeHavenPulumiError(msg) from exc
 
     def diff(
         self,
-        id_: str,
-        old_props: Dict[str, Any],
-        new_props: Dict[str, Any],
+        old_props: dict[str, Any],
```
jemrobinson, line 91:
```suggestion
        id_: str,
        old_props: dict[str, Any],
```

data_safe_haven/pulumi/dynamic/compiled_dsc.py
```diff
@@ -53,15 +52,14 @@ def create(self, props: Dict[str, Any]) -> CreateResult:
             outs=dict(**props),
         )
 
-    def delete(self, id_: str, props: Dict[str, Any]) -> None:
+    def delete(self) -> None:
```
jemrobinson:
```suggestion
    def delete(self, id_: str, props: dict[str, Any]) -> None:
```

data_safe_haven/pulumi/dynamic/compiled_dsc.py
```diff
@@ -53,15 +52,14 @@ def create(self, props: Dict[str, Any]) -> CreateResult:
...
         """The Python SDK does not support configuration deletion"""
         return
 
     def diff(
         self,
-        id_: str,
-        old_props: Dict[str, Any],
-        new_props: Dict[str, Any],
+        old_props: dict[str, Any],
```
jemrobinson, line 63:
```suggestion
        id_: str,
        old_props: dict[str, Any],
```

data_safe_haven/pulumi/dynamic/file_share_file.py
```diff
@@ -96,17 +93,16 @@ def create(self, props: Dict[str, Any]) -> CreateResult:
...
-            ) from exc
+            msg = f"Failed to upload data to [green]{file_name}[/] in [green]{props['share_name']}[/].\n{exc}"
+            raise DataSafeHavenAzureError(msg) from exc
         return CreateResult(
             f"filesharefile-{props['destination_path'].replace('/', '-')}",
             outs=outs,
         )
 
-    def delete(self, id_: str, props: Dict[str, Any]) -> None:
+    def delete(self, props: dict[str, Any]) -> None:
```
jemrobinson:
```suggestion
    def delete(self, id_: str, props: dict[str, Any]) -> None:
```

data_safe_haven/pulumi/dynamic/file_share_file.py
```diff
@@ -118,15 +114,13 @@ def delete(self, id_: str, props: Dict[str, Any]) -> None:
...
-            ) from exc
+            msg = f"Failed to delete file [green]{file_name}[/] in [green]{props['share_name']}[/].\n{exc}"
+            raise DataSafeHavenAzureError(msg) from exc
 
     def diff(
         self,
-        id_: str,
-        old_props: Dict[str, Any],
-        new_props: Dict[str, Any],
+        old_props: dict[str, Any],
```
jemrobinson, line 125:
```suggestion
        id_: str,
        old_props: dict[str, Any],
```

data_safe_haven/pulumi/dynamic/remote_powershell.py
```diff
@@ -50,15 +48,10 @@ def create(self, props: Dict[str, Any]) -> CreateResult:
...
-    def delete(self, id_: str, props: Dict[str, Any]) -> None:
-        """The Python SDK does not support configuration deletion"""
-        return
-
     def diff(
         self,
-        id_: str,
-        old_props: Dict[str, Any],
-        new_props: Dict[str, Any],
+        old_props: dict[str, Any],
```
jemrobinson, line 59:
```suggestion
        id_: str,
        old_props: dict[str, Any],
```

data_safe_haven/pulumi/dynamic/ssl_certificate.py
```diff
@@ -125,15 +112,17 @@ def create(self, props: Dict[str, Any]) -> CreateResult:
...
+                f" for [green]{props['domain_name']}[/].\n{exc}"
+            )
+            raise DataSafeHavenSSLError(msg) from exc
         return CreateResult(
             f"SSLCertificate-{props['certificate_secret_name']}",
             outs=outs,
         )
 
-    def delete(self, id_: str, props: Dict[str, Any]) -> None:
+    def delete(self, props: dict[str, Any]) -> None:
```
jemrobinson:
```suggestion
    def delete(self, id_: str, props: dict[str, Any]) -> None:
```

data_safe_haven/pulumi/dynamic/ssl_certificate.py
```diff
@@ -149,15 +138,16 @@ def delete(self, id_: str, props: Dict[str, Any]) -> None:
...
+                f" for [green]{props['domain_name']}[/].\n{exc}"
+            )
+            raise DataSafeHavenSSLError(msg) from exc
 
     def diff(
         self,
-        id_: str,
-        old_props: Dict[str, Any],
-        new_props: Dict[str, Any],
+        old_props: dict[str, Any],
```
jemrobinson, line 162:
```suggestion
        id_: str,
        old_props: dict[str, Any],
```

data_safe_haven/pulumi/dynamic/dsh_resource_provider.py
JimMadge: Fixed in 65280d45

data_safe_haven/utility/types.py
JimMadge: Fixed in 6af37b02

data_safe_haven/cli.py
```diff
@@ -1,32 +1,27 @@
...
 from data_safe_haven.utility import Logger
 
 
 def callback(
     output: Annotated[
-        Optional[pathlib.Path],
-        typer.Option(
-            "--output", "-o", resolve_path=True, help="Path to an output log file"
-        ),
+        pathlib.Path | None,
```
jemrobinson:
```suggestion
        Optional[pathlib.Path],
```

data_safe_haven/cli.py
```diff
@@ -1,32 +1,27 @@
...
+        typer.Option("--output", "-o", resolve_path=True, help="Path to an output log file"),
     ] = None,
     verbosity: Annotated[
-        Optional[int],
+        int | None,
```
jemrobinson:
```suggestion
        Optional[int],
```

data_safe_haven/cli.py
```diff
@@ -36,16 +31,14 @@ def callback(
         ),
     ] = None,
     version: Annotated[
-        Optional[bool],
-        typer.Option(
-            "--version", "-V", help="Display the version of this application."
-        ),
+        bool | None,
```
jemrobinson:
```suggestion
        Optional[bool],
```

data_safe_haven/commands/typer_init.py
```diff
@@ -1,18 +1,16 @@
...
-from typing import Annotated, Optional
+from typing import Annotated
 
-# Third party imports
 import typer
 
-# Local imports
 from data_safe_haven.functions import validate_aad_guid
+
 from .initialise_command import InitialiseCommand
```
jemrobinson, line 8: Do we want this to be `from data_safe_haven.commands.initialise_command import InitialiseCommand`?

data_safe_haven/commands/typer_init.py
```diff
@@ -1,18 +1,16 @@
...
 
 
 def initialise_command(
     admin_group: Annotated[
-        Optional[str],
+        str | None,
```
jemrobinson:
```suggestion
        Optional[str], # noqa: UP007
```

data_safe_haven/commands/typer_init.py
```diff
@@ -21,23 +19,23 @@ def initialise_command(
         ),
     ] = None,
     location: Annotated[
-        Optional[str],
+        str | None,
```
jemrobinson:
```suggestion
        Optional[str], # noqa: UP007
```

data_safe_haven/commands/typer_init.py
```diff
@@ -21,23 +19,23 @@ def initialise_command(
...
         typer.Option(
             "--location",
             "-l",
             help="The Azure location to deploy resources into.",
         ),
     ] = None,
     name: Annotated[
-        Optional[str],
+        str | None,
```
jemrobinson:
```suggestion
        Optional[str], # noqa: UP007
```

data_safe_haven/commands/typer_init.py
```diff
@@ -21,23 +19,23 @@ def initialise_command(
...
         typer.Option(
             "--name",
             "-n",
             help="The name to give this Data Safe Haven deployment.",
         ),
     ] = None,
     subscription: Annotated[
-        Optional[str],
+        str | None,
```
jemrobinson:
```suggestion
        Optional[str],  # noqa: UP007
```

data_safe_haven/config/__init__.py
```diff
@@ -1,5 +1,5 @@
-from .config import Config
 from .backend_settings import BackendSettings
+from .config import Config
```
jemrobinson, line 2:
```suggestion
from data_safe_haven.config.backend_settings import BackendSettings
from data_safe_haven.config.config import Config
```

?

data_safe_haven/pulumi/components/sre_networking.py
```diff
@@ -259,7 +222,10 @@ def __init__(
                 ),
                 network.SecurityRuleArgs(
                     access=network.SecurityRuleAccess.ALLOW,
-                    description="Allow LDAP client requests over TCP (see https://devopstales.github.io/linux/pfsense-ad-join/ for details).",
+                    description=(
+                        "Allow LDAP client requests over TCP"
+                        " (see https://devopstales.github.io/linux/pfsense-ad-join/ for details)."
+                    ),
```
jemrobinson:
```suggestion
                    description=(
                        "Allow LDAP client requests over TCP. "
                        "See https://devopstales.github.io/linux/pfsense-ad-join/ for details."
                    ),
```

data_safe_haven/pulumi/components/sre_research_desktop.py
```diff
@@ -1,22 +1,20 @@
...
 from data_safe_haven.functions import b64encode, replace_separators
 from data_safe_haven.pulumi.common.transformations import (
     get_available_ips_from_subnet,
     get_name_from_rg,
     get_name_from_subnet,
     get_name_from_vnet,
 )
-from .virtual_machine import LinuxVMProps, VMComponent
+
+from .components.virtual_machine import LinuxVMProps, VMComponent
```
jemrobinson:
```suggestion
from data_safe_haven.pulumi.components.virtual_machine import LinuxVMProps, VMComponent
```
?

JimMadge: Fixed in 564e536f

data_safe_haven/commands/typer_init.py
JimMadge: I think they are the same, this file is in the `data_safe_haven.commands` module. And we are allowing relative imports of siblings now.

data_safe_haven/config/__init__.py
JimMadge: I think there are the same, this file is in the `data_safe_haven.commands` module. And we are allowing relative imports of siblings now.

data_safe_haven/exceptions/__init__.py
JimMadge: These are all used :+1:.

I have opened an issue about clarifying their use https://github.com/alan-turing-institute/data-safe-haven/issues/1518

jemrobinson (approved):
Tested by redeploying an SHM and SRE.

PR #1508, opened 2023-07-21
Add script to automate account deletion

edwardchalstrey1:

@alan-turing-institute/data-safe-haven-code-administrators I will add instructions on how/when to use this script in the Turing context to the Turing TRESA docs, but for the DSH docs I think this is sufficient

Turing internal production environments repo: https://github.com/alan-turing-institute/trusted-research/issues/224

### :microscope: Tests

- Running the script on a test SHM and checking that users not in a SG are deleted
- Running with the -dryRun flag users are listed but not deleted


deployment/administration/SRE_Delete_Unassigned_Users.ps1
```diff
@@ -0,0 +1,46 @@
...
+$null = Set-AzContext -SubscriptionId $config.subscriptionName -ErrorAction Stop
+Add-LogMessage -Level Info "Exporting user list for $($config.shm.id) from $($config.dc.vmName)..."
+# Run remote script
+$script = @"
+`$userOuPath = (Get-ADObject -Filter * | Where-Object { `$_.Name -eq "Safe Haven Research Users" }).DistinguishedName
+`$users = Get-ADUser -Filter * -SearchBase "`$userOuPath" -Properties *
+foreach (`$user in `$users) {
+    `$groupName = (`$user | Select-Object -ExpandProperty MemberOf | ForEach-Object { ((`$_ -Split ",")[0] -Split "=")[1] }) -join "|"
+    `$user | Add-Member -NotePropertyName GroupName -NotePropertyValue `$groupName -Force
+}
```
jemrobinson: Not quite sure of the logic here. You combine all groups that the user is a member of with `|` (eg. `group1|group2`) then add this as a `GroupName` property on each user?

I don't think you need to be writing properties to users for this to work.

deployment/administration/SRE_Delete_Unassigned_Users.ps1
```diff
@@ -0,0 +1,46 @@
...
+
+# Delete users not found in any group (with exception for named SG e.g. "Sandbox")
+# --------------------------------------------------------------------------------
+Add-LogMessage -Level Info "Deleting users from $($config.shm.id) not in any security group..."
+$users = $result.Value[0].Message | ConvertFrom-Csv
+foreach ($user in $users) {
+    if (!($user.GroupName)) {
+        $name = $user.SamAccountName
+        $script = "Remove-ADUser -Identity $name"
+        Invoke-RemoteScript -Shell "PowerShell" -Script $script -VMName $config.dc.vmName -ResourceGroupName $config.dc.rg
```
jemrobinson: So we run one script to get a list of users then one script per user to delete them? I think this would be much better as a single script that does this all in one pass.

jemrobinson:
>     1. The script that lives directly in the DC - this would need to be manually run by an admin by logging into the SHM after a teardown - I have tested this and it works
> 
>     2. The script that is run outside the DC and invokes the deletion - Note: currently this seems to hang at the deletion step when I've tested it
> 
> The original reason for the 2nd approach was so that it could be run as a final step of the SRE teardown script, but actually, it might be we don't want this, what to people think? If we do, it now occurs to me it may be possible to invoke the script from the 1st approach anyway.

Do the second one. Uploading scripts to the SHM is complex (we upload them to a storage account and then copy them to the SHM). It's only necessary for scripts that *need* to be run in a GUI (eg. ones that have an AAD login pop-up).

Remote scripts will hang if there is any kind of error in them (also can't be retried for 90 minutes). Make sure you test interactively but non-destructively on the DC.

deployment/administration/SRE_Delete_Unassigned_Users.ps1
edwardchalstrey1: My starting point was copying from this: https://github.com/alan-turing-institute/data-safe-haven/blob/develop/deployment/safe_haven_management_environment/setup/Copy_SHM_Users.ps1 - I can refactor

deployment/administration/SRE_Delete_Unassigned_Users.ps1
```diff
@@ -0,0 +1,46 @@
...
+`$userOuPath = (Get-ADObject -Filter * | Where-Object { `$_.Name -eq "Safe Haven Research Users" }).DistinguishedName
+`$users = Get-ADUser -Filter * -SearchBase "`$userOuPath" -Properties *
+foreach (`$user in `$users) {
+    `$groupName = (`$user | Select-Object -ExpandProperty MemberOf | ForEach-Object { ((`$_ -Split ",")[0] -Split "=")[1] }) -join "|"
+    `$user | Add-Member -NotePropertyName GroupName -NotePropertyValue `$groupName -Force
+}
+`$users | Select-Object SamAccountName,GivenName,Surname,Mobile,GroupName | `
+         ConvertTo-Csv | Where-Object { `$_ -notmatch '^#' } | `
+         ForEach-Object { `$_.replace('"','') }
+"@
```
JimMadge: Can we move the script to a new file. That way the CI will be applied to it.

deployment/administration/SRE_Delete_Unassigned_Users.ps1
JimMadge: What output do we get from the remote scripts?
It might be useful to split the operations this way if it is the only (or best) way to log the users being removed and the success/failure of each removal.

edwardchalstrey1:
> It might be useful to split the operations this way if it is the only (or best) way to log the users being removed and the success/failure of each removal.

The command in this remote script won't give an output, but this should be something that is easy for an SHM admin to check manually

@JimMadge maybe I should add a second part of the script that prints out the remaining users and their security group or something like that, for the purpose of checking?

JimMadge:
> The command in this remote script won't give an output, but this should be something that is easy for an SHM admin to check manually
> 
> @JimMadge maybe I should add a second part of the script that prints out the remaining users and their security group or something like that, for the purpose of checking?

Yes something like a before and after list of users could work. It feels like this is place where it is important to understand what has happened. An admin will need to confirm the users were removed correctly (and fix any cases where removing a user failed).

jemrobinson (changes requested):
A couple of thoughts here.

1. It would be good to add a dry-run flag (typically done using `[CmdletBinding(SupportsShouldProcess=$True)]` which gives you a -WhatIf flag) to show which users would be removed without doing it
2. The script called `SRE_Delete_Unassigned_Users.ps1` doesn't have anything to do with SREs, right? Maybe change the name.

deployment/administration/Delete_Unassigned_Users.ps1
```diff
@@ -0,0 +1,28 @@
...
+    $user | Add-Member -NotePropertyName GroupName -NotePropertyValue $groupName -Force
+}
+
+# Delete users not found in any group
+foreach ($user in $users) {
+    if (!($user.GroupName)) {
+        $name = $user.SamAccountName
+        Remove-ADUser -Identity $name
+    }
+}
```
jemrobinson: I don't love this pattern of getting an array of users and then decorating them with group name before iterating through them again. Why not just do:

```
for user in users
  if user not in group
    delete user
```

deployment/administration/Delete_Unassigned_Users.ps1
```diff
@@ -0,0 +1,28 @@
...
+try {
+    Import-Module -Name "C:\Program Files\Microsoft Azure AD Sync\Bin\ADSync" -ErrorAction Stop
+    Start-ADSyncSyncCycle -PolicyType Delta
+}
+catch [System.IO.FileNotFoundException] {
+    Write-Output "Skipping as Azure AD Sync is not installed"
+}
+catch {
+    Write-Output "Unable to run Azure Active Directory synchronisation!"
+}
```
jemrobinson: In a dry-run we'd like to be able to skip this.

deployment/administration/SRE_Delete_Unassigned_Users.ps1
```diff
@@ -0,0 +1,25 @@
...
+# -------------------------------
+$config = Get-ShmConfig -shmId $shmId
+$originalContext = Get-AzContext
+
+# Delete users not currently in a security group
+# ----------------------------------------------
+$null = Set-AzContext -SubscriptionId $config.subscriptionName -ErrorAction Stop
+Add-LogMessage -Level Info "Deleting users not assigned to any security group: $($config.shm.id) from $($config.dc.vmName)..."
+
+$script = "Delete_Unassigned_Users.ps1"
```
jemrobinson: Currently all the scripts in `administration` are ones that can be run locally. Can you put `Delete_Unassigned_Users.ps1` somewhere else (e.g. in a `remote` subdirectory of `administration`)?

jemrobinson (changes requested):

docs/source/roles/system_manager/manage_users.md
```diff
@@ -137,6 +137,14 @@ The `DC1` is the source of truth for user details. If these details need to be c
     - Click on `Users` under `Manage` and search for the user
     - Confirm the user is no longer present
 
+### {{x}} Automatically deleting all usassigned users
```
jemrobinson:
```suggestion
### {{x}} Automatically deleting all unassigned users
```

docs/source/roles/system_manager/manage_users.md
```diff
@@ -137,6 +137,14 @@ The `DC1` is the source of truth for user details. If these details need to be c
...
+
+In some situations, such as at the end of a project after an SRE has been torn down, you may want to remove all users from the SHM who are not assigned to the security group of any remaining attached SREs.
```
jemrobinson, line 142: This shouldn't be "in some situations", it should be a regular thing that we run every month.

jemrobinson (approved):
LGTM - if it works :)

docs/source/roles/system_manager/manage_users.md
edwardchalstrey1: As discussed, that will be added to TRESA docs, not here

edwardchalstrey1:
@jemrobinson please merge I'm not allowed

PR #1510, opened 2023-07-21
Refactor config files

### LINKED ISSUE #1509: Pulumi: Simplify config set up

Pulumi `python-migration` branch currently uses two config files (local and remote) and the interaction between them is complicated. The `Config` class is flexible but doesn't work well with type-hints.

- Update `Config`
- Update `DotFileSettings`

jemrobinson:

- Move `external.api` and `external.interface` into `external` (unnecessary nesting)
- Separate `PulumiStack` into `PulumiSHMStack` and `PulumiSREStack` (better than changing behaviour with strings)
- Move Azure authentication from mixin to external
- Replaced `DotFileSettings` with better named `BackendSettings`
- Store `BackendSettings` in `appdirs` config path instead of local directory
- Simplify use of `BackendSettings` to only be updated in `init` and read during `Config` construction
- Convert `Config` from a `dotmap` to a nested set of dataclasses (allows type-hinting to work correctly)

Closes #1509

**Check #1506 before merging**

### :microscope: Tests

Tested locally


data_safe_haven/config/backend_settings.py
```diff
@@ -0,0 +1,144 @@
...
+        self._location: Optional[str] = None
+        self._name: Optional[str] = None
+        self._subscription_name: Optional[str] = None
+        self.logger = Logger()
+
+        # Load previous backend settings (if any)
+        self.config_directory = pathlib.Path(
+            appdirs.user_config_dir("data_safe_haven")
+        ).resolve()
+        self.config_file_path = self.config_directory / ".dshconfig"
```
JimMadge: Probably doesn't need to be a dotfile in `~/.config/`.

JimMadge: It would probably be best to create a variable with the name of the configuration file (or full path) rather than using a string literal.

data_safe_haven/config/backend_settings.py
```diff
@@ -0,0 +1,144 @@
...
+
+class BackendSettings:
+    """Load global and local settings from dotfiles with structure like the following
+
+    azure:
+      admin_group_id: d5c5c439-1115-4cb6-ab50-b8e547b6c8dd
+      location: uksouth
+      subscription_name: Data Safe Haven Development
+    current:
+      name: Turing Development
```
JimMadge: Example secrets 😬 ?

data_safe_haven/config/backend_settings.py
```diff
@@ -0,0 +1,144 @@
...
+    def subscription_name(self) -> str:
+        if not self._subscription_name:
+            raise DataSafeHavenParameterException(
+                "Azure subscription not provided: use '[bright_cyan]--subscription[/]' / '[green]-s[/]' to do so."
+            )
+        return self._subscription_name
+
+    def read(self) -> None:
+        """Read settings from YAML file"""
+        if self.config_file_path.exists():
```
JimMadge: Do we want to raise an exception if it does not exist?

data_safe_haven/config/backend_settings.py
```diff
@@ -0,0 +1,144 @@
...
+            with open(self.config_file_path, "r", encoding="utf-8") as f_yaml:
+                settings = yaml.safe_load(f_yaml)
+            if isinstance(settings, dict):
```
JimMadge: Does `yaml.safe_load` return `None` (or similar) if it fails?
Do we want to raise an exception in that instance?

data_safe_haven/config/config.py
```diff
@@ -1,199 +1,416 @@
...
+    validate_email_address,
+    validate_ip_address,
+    validate_timezone,
+)
+from data_safe_haven.utility import SoftwarePackageCategory, YamlType
+from .backend_settings import BackendSettings
+
+
+@dataclass
+class ConfigSectionAzure:
```
JimMadge, line 35: These feel like good use cases for dataclasses :+1:.

data_safe_haven/config/config.py
```diff
@@ -1,199 +1,416 @@
...
 
-        # Try to load the full config from blob storage
+    def validate(self) -> None:
+        """Validate input parameters"""
+        try:
+            validate_aad_guid(self.admin_group_id)
+        except Exception as exc:
+            raise TypeError(
+                f"Invalid value for 'admin_group_id' ({self.admin_group_id}).\n{str(exc)}"
+            )
```
JimMadge: Sounds like these should be `ValueError`s rather than `TyperError`s.

https://github.com/alan-turing-institute/data-safe-haven/pull/1506

> Passing arguments of the wrong type (e.g. passing a [list](https://docs.python.org/3/library/stdtypes.html#list) when an [int](https://docs.python.org/3/library/functions.html#int) is expected) should result in a [TypeError](https://docs.python.org/3/library/exceptions.html?highlight=typeerror#TypeError), but passing arguments with the wrong value (e.g. a number outside expected boundaries) should result in a [ValueError](https://docs.python.org/3/library/exceptions.html?highlight=typeerror#ValueError).

data_safe_haven/config/config.py
```diff
@@ -1,199 +1,416 @@
...
+        except Exception as exc:
+            raise TypeError(
+                f"Invalid value for 'tenant_id' ({self.tenant_id}).\n{str(exc)}"
+            )
 
-    def __str__(self) -> str:
-        return str(yaml.dump(self._map.toDict(), indent=2))
+    def to_dict(self) -> Dict[str, str]:
+        self.validate()
+        return chili.encode(self)  # type: ignore
```
JimMadge: Is this for writing to file?

JimMadge: Because `chili.encode` returns `dict[Any, Any]`?

data_safe_haven/config/config.py
```diff
@@ -1,199 +1,416 @@
...
+        if name in self.pulumi.stacks.keys():
+            del self.pulumi.stacks[name]
+
+    def write_stack(self, name: str, path: pathlib.Path):
+        """Write a Pulumi stack file from config"""
+        b64string = b64decode(self.pulumi.stacks[name])
+        with open(path, "w", encoding="utf-8") as f_stack:
+            f_stack.write(b64string)
+
+    def download(self):
```
JimMadge: Would it be better to have this as a classmethod?

data_safe_haven/config/config.py
```diff
@@ -1,199 +1,416 @@
...
+                    self.backend_storage_container_name,
+                )
             )
-        except Exception as exc:
-            raise DataSafeHavenAzureException(
-                f"Configuration file could not be uploaded to '{self.backend.storage_account_name}'."
-            ) from exc
+        )
+
+    def from_yaml(self, yaml_input: YamlType) -> None:
```
JimMadge: Would it be better to have this as a classmethod?

data_safe_haven/functions/validators.py
```diff
@@ -45,10 +45,99 @@ def validate_ip_address(
...
+    if not location:
+        return None
+    if location in (
+        "asia",
+        "asiapacific",
+        "australia",
+        "australiacentral",
+        "australiacentral2",
+        "australiaeast",
+        "australiasoutheast",
```
JimMadge: Is there an Azure python module way to do this? Not sure we want to maintain this list.

JimMadge (changes requested):
Looks good.

I think this config structure makes more sense. Using dataclasses is sensible and will help with validation. (We could use something like pydantic, but I think that is overkill unless you have many dataclass-type classes).

It looks like this has enabled other places of the code to be tidied too.

data_safe_haven/config/config.py
JimMadge: I think I saw later it was for the `__str__` method.

JimMadge: In which case, it is probably better to match the return type here rather than pretending it is something else.

jemrobinson: No, it returns `StateObject = Union[None, StatefulTypes, Dict[str, StatefulTypes]]` which isn't convertable to anything useful. I'd like to write something like:

```
encoded = chili.encode(self)
if not isinstance(encoded, dict(str, str)):
    raise TypeError
return encoded
```

but don't think you can do `isinstance` on a subtype.

data_safe_haven/config/config.py
jemrobinson: Maybe. We'd have to move `_azure_api` to a class-level property too (is that a thing?). What would we gain?

data_safe_haven/config/config.py
jemrobinson: Maybe a better question is "should `from_yaml` and `download` just be moved into `__init__` since that's the only place they're used. What do you think?

data_safe_haven/functions/validators.py
jemrobinson: There is, but then you add a dependency on Azure API which is going to complicate the dependency chain for something you're mainly using to verify CLI arguments. I'll add a method to Azure API for this.

data_safe_haven/config/backend_settings.py
jemrobinson: Done

jemrobinson: I'm confused about what you're suggesting here. At some point we need to set this name from a string. Would you rather see something like this?

```
config_file_name = "config.yaml"
self.config_file_path = self.config_directory / config_file_name
```

data_safe_haven/config/backend_settings.py
jemrobinson: 1. The admin_group_id is a dummy variable
2. The others are not secret

Happy to remove/replace them though. Thoughts?

data_safe_haven/config/backend_settings.py
jemrobinson: No. The logic is:

1. Try to read variables from YAML file
2. Overwrite using any variables provided at the command line
3. Write to YAML file (which raises exceptions if any variables are missing)

data_safe_haven/config/backend_settings.py
jemrobinson: I think probably not (see above) but there is the possible case of a corrupted/unreadable file. Thoughts?

data_safe_haven/config/config.py
jemrobinson: Done. The fact that I wrote "Invalid value" in the message should have been a hint!

data_safe_haven/config/config.py
jemrobinson: Yes, it's for the `Config.__str__` method.

data_safe_haven/functions/validators.py
jemrobinson: I've added `AzureAPI.get_locations()` but in order to construct `AzureAPI` you need to know `subscription_name`. That means that we can't call anything that depends on `AzureAPI.get_locations()` in `ConfigSectionAzure` since `subscription_name` is not known here (and may not be set)

jemrobinson: I've replaced this with a minimal regex: `at least one lowercase letter followed by 0 or 1 digits then 0 or more lowercase letters`.

data_safe_haven/config/config.py
jemrobinson: Done

data_safe_haven/config/config.py
jemrobinson: Refactored

data_safe_haven/config/backend_settings.py
jemrobinson: I have wrapped this in a DataSafeHavenConfigException so if it happens it will be nicely logged.

data_safe_haven/config/config.py
jemrobinson: Done via a convenience function.

data_safe_haven/config/backend_settings.py
jemrobinson: Suggestion was to put the name at class or module level in case it's needed later. For the moment it's not, so we'll leave it. May be refactored in future.

JimMadge (approved):

PR #1526, opened 2023-07-27
Improve Pulumi Logger behaviour

jemrobinson:

- Make LoggingSingleton inherit from Singleton metaclass
- Make LoggingSingleton a logging.logger (so line-by-line tracebacks will work correctly)

### :microscope: Tests

Tested on running deployment


JimMadge:
I like this, the singleton metaclass looks really clean and simple :+1:.

I don't think I've used a metaclass before :eyes:.

jemrobinson:
Still has a problem that you can't use `logging.getLogger("data_safe_haven")` as this will return a `logging.Logger` not ` LoggingSingleton` so type-hinting would break if you call `logging.getLogger().custom_function()`.

Hopefully using `logger = LoggingSingleton()` or simply `LoggingSingleton().info("my message")` is sufficiently easy syntax to understand.

PR #1527, opened 2023-07-27
Add type checking

jemrobinson:

- Fix type errors
- Enable type checking in CI

n/a

### :microscope: Tests

Tested locally and in CI


pyproject.toml
```diff
@@ -59,10 +59,17 @@ dependencies = [
   "black>=23.1.0",
   "mypy>=1.0.0",
   "ruff>=0.0.243",
+  "types-appdirs>=1.4.3.5",
+  "types-chevron>=0.14.2.5",
+  "types-psycopg2>=2.9.21.11",
+  "types-pytz>=2023.3.0.0",
+  "types-PyYAML>=6.0.12.11",
+  "types-requests>=2.31.0.2",
```
JimMadge, line 67: I believe if you use the `mypy` arugments `--install-types --non-interactive` we shouldn't need to handle the typing modules for our dependencies.

JimMadge: Might want to create an issue to fix this later.

I need a reminder on what the `.pyi` files in the typings directory does again 😅.

jemrobinson: I think it's better to be explicit about which packages we're using rather than letting `mypy` decide.

The `.pyi` files are like C/C++ header files: they contain stubs of the available functions in the equivalent `.py` file. We use them when the upstream packages don't provide their own. We need them for:

- pulumi and pulumi-azure-native to force mypy to only process those bits of the module we're actually using (this works because `__init__.pyi` is read instead of `__init__.py`)
- typer because otherwise we get untyped decorator errors (https://stackoverflow.com/questions/65621789/mypy-untyped-decorator-makes-function-my-method-untyped)

JimMadge: > I think it's better to be explicit about which packages we're using rather than letting mypy decide.

I normally lean that way, although in this case my understanding is that mypy installs known type stub packages from the typeshed database(?). It is almost certainly ends up doing what we would do manually, so it feels sensible to remove the burden from ourselves.

PR #1533, opened 2023-08-01
Fix typing of external APIs

jemrobinson:

Fix typing issues arising from external Azure SDK libraries

n/a

### :microscope: Tests

Tested locally

### DIFF
--- a/data_safe_haven/external/api/azure_api.py
+++ b/data_safe_haven/external/api/azure_api.py
@@ -21,9 +21,15 @@ from azure.mgmt.automation import AutomationClient
 from azure.mgmt.automation.models import (
     DscCompilationJobCreateParameters,
     DscConfigurationAssociationProperty,
+    Module,
 )
 from azure.mgmt.compute import ComputeManagementClient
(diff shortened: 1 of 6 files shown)


JimMadge:
I remember talking about this. Did `typing.cast` turn out to be the most sensible thing to do (or at least more sensible than ignoring types)?

jemrobinson:
Possible options (ordered by how much I like them)

1. Get Microsoft to fix their type hints
2. Write our own type hints (turns out to be too much work in this case)
3. Use cast to manually specify the type we *know* is returned
4. Ignore type hints

If you'd like to try (1) or (2) then feel free :)

JimMadge:
Agreed that an upstream fix is probably the real answer.

I wouldn't want to maintain our own type hints for large, third-party libraries just to make `mypy` pass though. I think casts or ignores are better in that case.

Maybe a rule of thumbs is,

- `typing.cast` when you are _very_ confident about the static type
- ignore when the type is not static or is poorly defined

PR #1538, opened 2023-08-04
Remove magic number and add error handling

### LINKED ISSUE #1513: Magic number
There is a magic number in the code here:

https://github.com/alan-turing-institute/data-safe-haven/blob/f0a890632725e2b6883c8997787294b824caa1cb/data_safe_haven/administration/users/active_directory_users.py#L88-L98

It isn't clear what this number represents and there isn't handling for when it is `<6`.

JimMadge:

- Remove magic number
- Add DataSafeHavenActiveDirectoryError exception
- Add error handling for when number of fields is not what is expected

I'm a little ambivalent about the error handling. Checking that the number of columns returned is what we expect is sensible, we are interacting with an external process.

However, this

https://github.com/alan-turing-institute/data-safe-haven/blob/c8e9b7ab303364411b57b029c8aa005dfecf55b4/data_safe_haven/administration/users/active_directory_users.py#L44-L51

Is already checked here

https://github.com/alan-turing-institute/data-safe-haven/blob/c8e9b7ab303364411b57b029c8aa005dfecf55b4/data_safe_haven/administration/users/user_handler.py#L40-L51

Closes #1513

### DIFF
--- a/data_safe_haven/administration/users/active_directory_users.py
+++ b/data_safe_haven/administration/users/active_directory_users.py
@@ -4,6 +4,7 @@ from collections.abc import Sequence
 from typing import Any
 
 from data_safe_haven.config import Config
+from data_safe_haven.exceptions import DataSafeHavenActiveDirectoryError
 from data_safe_haven.external import AzureApi
 from data_safe_haven.functions import b64encode
 from data_safe_haven.pulumi import PulumiSHMStack
@@ -60,6 +61,12 @@ class ActiveDirectoryUsers:
                         ]
                     )
                 ]
+            else:
+                msg = (
+                    f"User {user.username} does not have all the required fields to "
+                    "add to the Active Directory."
+                )
+                raise DataSafeHavenActiveDirectoryError(msg)
         user_details_b64 = b64encode("\n".join(csv_contents))
         output = self.azure_api.run_remote_script(
             self.resource_group_name,
@@ -83,9 +90,10 @@ class ActiveDirectoryUsers:
             self.vm_name,
         )
         users = []
+        expected_columns = 6
         for line in output.split("\n"):
             tokens = line.split(";")
-            if len(tokens) >= 6:  # noqa: PLR2004
+            if len(tokens) >= expected_columns:
                 users.append(
                     ResearchUser(
                         email_address=tokens[4],
@@ -96,6 +104,12 @@ class ActiveDirectoryUsers:
                         user_principal_name=tokens[5],
                     )
                 )
+            else:
+                msg = (
+                    "Fewer than expected number of fields returned for Active Directory"
+                    f" user: {tokens}"
+                )
+                raise DataSafeHavenActiveDirectoryError(msg)
         return users
 
     def register(self, sre_name: str, usernames: Sequence[str]) -> None:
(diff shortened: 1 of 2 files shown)


jemrobinson:
As the flow is:

- check we're providing the right arguments
- run an external script with those arguments
- get the output
- check that the output has the right form

I think it's correct to have two checks.

data_safe_haven/administration/users/active_directory_users.py
```diff
@@ -83,9 +90,10 @@ def list(self, sre_name: str | None = None) -> Sequence[ResearchUser]:  # noqa:
             self.vm_name,
         )
         users = []
+        expected_columns = 6
```
jemrobinson, line 93: How about replacing lines 93-112 with this, which will check that there are exactly 6 tokens?

```suggestion
        for line in output.split("\n"):
            try:
                sam_account_name, given_name, surname, phone_number, email_address, user_principal_name = line.split(";")
                users.append(
                    ResearchUser(
                        email_address=email_address,
                        given_name=given_name,
                        phone_number=phone_number,
                        sam_account_name=sam_account_name,
                        surname=surname,
                        user_principal_name=user_principal_name,
                    )
                )
            except ValueError as exc:
                msg = "Unexpected number of fields returned for Active Directory user."
                raise DataSafeHavenActiveDirectoryError(msg)
```

jemrobinson (changes requested):

data_safe_haven/administration/users/active_directory_users.py
JimMadge: That's fine I think (assuming it works).

The original had `if len(tokens) >= 6:` but here the list must have exactly 6 items. Will it ever be more than 6?

I would rather use an if/else block than try/except. I think it is clearer and _usually_ more performant.

jemrobinson: Here's the script https://github.com/alan-turing-institute/data-safe-haven/blob/magic_number/data_safe_haven/resources/active_directory/list_users.ps1. We have control over what it returns, which is currently `"$($User.SamAccountName);$($User.GivenName);$($User.Surname);$($User.TelephoneNumber);$($User.Mail);$($User.UserPrincipalName)"`

The reason for `try/except` here is that `a, b, c = <sequence>` will raise a `ValueError` if there are not exactly 3 elements in `<sequence>`, so we need to catch that.

JimMadge: The script does seem to check that the required fields exist (at least some of them, maybe the others are mandatory?). Hard to say with confidence but it looks like it would always write six items separated by `;`, although some of them may be empty.

Probably still worth dealing with problems here though as the powershell script is external, could be out of sync.

> The reason for try/except here is that a, b, c = <sequence> will raise a ValueError if there are not exactly 3 elements in <sequence>, so we need to catch that.

Sure, but you could also do something like

```Python
if len(sequence) == 6:
    ...
else:
    raise ....
```

Where the thing you testing for is clearer.

jemrobinson (approved):
@JimMadge: happy with whichever option you choose to resolve the external Powershell script issue.

data_safe_haven/administration/users/active_directory_users.py
JimMadge: Fixed in c8e9b7ab

jemrobinson (approved):
LGTM

PR #1539, opened 2023-08-04
Pulumi: pin all docker image versions

jemrobinson:

Pin all docker images to MAJOR.MINOR.PATCH and set to latest version at time of writing

n/a

### :microscope: Tests

Not tested


JimMadge (commented):
What do we think about the balance between update burden and confidence when pinning to patch versions?

Related to https://github.com/alan-turing-institute/data-safe-haven/pull/1501

jemrobinson:
Probably pinning to major versions is the right idea, but I guess they don't exist for all these images. How about writing a CI job that checks version numbers?

JimMadge:
I think pinning to major whenever possible is a good approach.

I was hoping depdendabot would be able to check for updates in docker-compose-like files but I don't think it does.

jemrobinson:
I'm going to close this but maybe revisit if we can think of a sensible way to automate it.

PR #1566, opened 2023-08-22
Pulumi: SRE DNS filtering

### LINKED ISSUE #1504: Pulumi: SHM DNS filtering
Could either use Windows DNS server on the SHM or a stand-alone solution eg. [PiHole](https://discourse.pi-hole.net/).

We want to allow access to all the `targetFqdns` from here (https://github.com/alan-turing-institute/data-safe-haven/blob/develop/deployment/safe_haven_management_environment/network_rules/shm-firewall-rules.json#L71). Some of these must *only* be allowed from the SHM DC and some must be allowed from anywhere.

All other traffic should be blocked.

### 2023-08-17:
PiHole will not work as it needs to be running as PID 1 which is reserved for `/pause` by Azure Container Instances (error is `s6-overlay-suexec: fatal: can only run as pid 1`).
### 2023-08-18
AdGuard correctly blocking DNS requests
Screenshot 2023-08-18 at 19 12 06
- not currently default DNS server
- AzureContainerInstances cannot expose UDP and TCP from the same container group(!)
### 2023-08-22:
- developed a standalone (no configuration file) AdGuard docker image
- deployed into a separate, peered, virtual network

jemrobinson:

- The easiest way to change the DNS servers for all resources in a virtual network is to change the VNet DHCP settings
- This means that the DNS server
  - must already exist when the virtual network is created
  - cannot belong to the virtual network
- Currently this is done by using the SHM DC as the DNS server but since we want to phase this out this PR instead does the following:
  - create a new DNS resource group at the beginning of SRE creation
  - this contains:
    - a virtual network with a single subnet
    - an NSG for that subnet which only allows inbound traffic from this SRE
    - a self-contained container instance (i.e. one that is configured only using environment variables or secrets-mounted-as-files and doesn't need mounted storage)
    - the private DNS zone that used to belong to the SRE
- With this in place, the SRE virtual network can be peered to the DNS virtual network and use the container as the default DNS server
    
Closes #1504 

#### DNS lookup restrictions are working

Screenshot 2023-08-25 at 17 13 16

- Cannot access unauthorised TXT record
- Cannot resolve Google
- Cannot resolve Google via Azure platform DNS
- Can resolve SRE nexus server
- Can resolve Ubuntu keyserver


data_safe_haven/functions/miscellaneous.py
```diff
@@ -4,6 +4,16 @@
...
 
 
+def allowed_dns_lookups() -> list[str]:
+    dns_lookups = {
+        "clamav": ["*.clamav.net", "database.clamav.net.cdn.cloudflare.net"],
+        "oauth": ["login.microsoftonline.com"],
+        "private_dns": [f"*.{zone_name}" for zone_name in ordered_private_dns_zones()],
+        "ubuntu_setup": ["keyserver.ubuntu.com"],
+    }
+    return sorted({zone for zones in dns_lookups.values() for zone in zones})
```
JimMadge: Is there an advantage to having this and `ordered_private_dns_zones` as functions?

data_safe_haven/pulumi/components/sre_dns_server.py
```diff
@@ -0,0 +1,319 @@
...
+        adguard_adguardhome_yaml_reader = FileReader(
+            resources_path / "AdGuardHome.mustache.yaml"
+        )
+
+        # Expand AdGuardHome YAML configuration
+        adguard_adguardhome_yaml_contents = Output.all(
+            admin_username=props.admin_username,
+            admin_password_encrypted=props.admin_password.apply(
+                lambda p: bcrypt_encode(p)
+            ),
```
JimMadge: Will this stop the deploy being idempotent? (As the `bcrypt_encode` function will generate a new salt each time).

Could we instead hold that as a pulumi variable with "create, don't replace"?

JimMadge: Does adguardhome explicitly ask for a bcrypt encrypted password?

If I recall correctly it is considered a weak encryption method now.

JimMadge: I mean, store the hashed password instead of/as well as the plain text.

data_safe_haven/pulumi/components/sre_dns_server.py
```diff
@@ -0,0 +1,319 @@
...
+            upstream_dns="168.63.129.16",
```
JimMadge, line 84:
```suggestion
            # Azure virtual DNS server
            # https://learn.microsoft.com/en-us/azure/virtual-network/what-is-ip-address-168-63-129-16
            upstream_dns="168.63.129.16",
```

data_safe_haven/pulumi/components/sre_dns_server.py
```diff
@@ -0,0 +1,319 @@
...
+                    protocol=network.SecurityRuleProtocol.ASTERISK,
+                    source_address_prefix="*",
+                    source_port_range="*",
+                ),
+                # Outbound
+                network.SecurityRuleArgs(
+                    access=network.SecurityRuleAccess.ALLOW,
+                    description="Allow outbound DNS and rules list traffic over the internet.",
+                    destination_address_prefix="Internet",
+                    destination_port_ranges=["53", "80", "443"],
```
JimMadge: Does this need HTTP and HTTPS as well?

data_safe_haven/pulumi/components/sre_dns_server.py
```diff
@@ -0,0 +1,319 @@
...
+                    route_table=None,
+                ),
+            ],
+            virtual_network_name=f"{stack_name}-vnet-dns",
+            virtual_network_peerings=[],
+            opts=ResourceOptions.merge(
+                child_opts,
+                ResourceOptions(
+                    ignore_changes=["virtual_network_peerings"]
+                ),  # allow peering to SRE virtual network
```
JimMadge, line 184: We can't define the peering here?

data_safe_haven/pulumi/components/sre_dns_server.py
```diff
@@ -0,0 +1,319 @@
...
+        )
+
+        # Expand AdGuardHome YAML configuration
+        adguard_adguardhome_yaml_contents = Output.all(
+            admin_username=props.admin_username,
+            admin_password_encrypted=props.admin_password.apply(
+                lambda p: bcrypt_encode(p)
+            ),
+            upstream_dns="168.63.129.16",
+            filter_deny=["*.*"],
```
JimMadge: Could the filter just be `*`? I suppose everything has a TLD so `*.*` always matches too?

data_safe_haven/resources/dns_server/AdGuardHome.mustache.yaml
```diff
@@ -0,0 +1,25 @@
...
+  upstream_dns_file: ""
+  bootstrap_dns:
+    - "{{upstream_dns}}"
+querylog:
+  enabled: true
+filters:
+user_rules:
+  {{#filter_deny}}
+  - "{{.}}"
+  {{/filter_deny}}
```
JimMadge: Given that we want to block everything except the allowed domains, should we just hard code here?

data_safe_haven/resources/dns_server/AdGuardHome.mustache.yaml
```diff
@@ -0,0 +1,25 @@
...
+  {{#filter_allow}}
+  - "@@||{{.}}"
```
JimMadge, line 18: Do we know how `||` interacts with `*`?

For example, does `@@||*example.com` match `sub.example.com`?

data_safe_haven/resources/dns_server/AdGuardHome.mustache.yaml
```diff
@@ -0,0 +1,25 @@
...
+  blocking_mode: nxdomain
+  upstream_dns:
+    - "{{upstream_dns}}"
+  upstream_dns_file: ""
+  bootstrap_dns:
+    - "{{upstream_dns}}"
+querylog:
+  enabled: true
+filters:
+user_rules:
```
JimMadge, line 14:
```suggestion
user_rules:
  # https://github.com/AdguardTeam/AdGuardHome/wiki/Hosts-Blocklists#adblock-style-syntax
```

typings/bcrypt/__init__.pyi
```diff
@@ -0,0 +1,10 @@
+def checkpw(password: bytes, hashed_password: bytes) -> bool: ...
+def gensalt(rounds: int = 12, prefix: bytes = b"2b") -> bytes: ...
+def hashpw(password: bytes, salt: bytes) -> bytes: ...
+def kdf(
+    password: bytes,
+    salt: bytes,
+    desired_key_bytes: int,
+    rounds: int,
+    ignore_few_rounds: bool = False,
+) -> bytes: ...
```
JimMadge: Didn't see this one being used.

typings/bcrypt/__init__.pyi
```diff
@@ -0,0 +1,10 @@
+def checkpw(password: bytes, hashed_password: bytes) -> bool: ...
```
JimMadge: Didn't see this one being used.

JimMadge (changes requested):
Looks good. A few questions mostly.

data_safe_haven/functions/miscellaneous.py
jemrobinson: They could equally easily be list properties but that feels like a stylistic point. If you want them to be simple lists, that's also possible, but would involve a bit of duplication since one depends on the other.

What change would you like to see?

data_safe_haven/pulumi/components/sre_dns_server.py
jemrobinson: Haha, good catch. I've been going backwards-and-forwards on this one. We can easily make `bcrypt_encode` use the same salt each time, but we'd have to implement our own version of [this function](https://github.com/pyca/bcrypt/blob/main/src/bcrypt/__init__.py#L52) and ensure we kept it up-to-date with any upstream changes (which I don't *expect* to happen as the `bcrypt` definition is unlikely to change).

Marking this as do-not-update to Pulumi will also ignore changes to the password itself, which we'd probably rather *not* do. Thoughts?

jemrobinson: Yes, it requires a `bcrypt`-encrypted password. This is still mildly safer than what we're doing everywhere else, which is sending a plaintext password to Azure using Pulumi...

data_safe_haven/pulumi/components/sre_dns_server.py
jemrobinson: Originally it did, as AdGuard supports use of remote block lists which are retrieved over HTTP(S). I decided that maintaining our own block list that had to always be available in the same location (e.g. on GitHub) wasn't very future proof. I think we can lose 80 and 443 here in that case.

data_safe_haven/pulumi/components/sre_dns_server.py
jemrobinson: Whenever you peer two virtual networks one has to exist before the other one. Whichever one exists first will start without any peerings.

data_safe_haven/pulumi/components/sre_dns_server.py
jemrobinson: I found that `*` didn't work for some reason. Other "block-everything" lists use `*.*` too. I agree that either should be fine though.

data_safe_haven/resources/dns_server/AdGuardHome.mustache.yaml
jemrobinson: Good point, I originally had this as a standalone Docker application where you might want to specify, but you're right that there's no need to do so here.

data_safe_haven/resources/dns_server/AdGuardHome.mustache.yaml
jemrobinson: See [here](https://adguard.com/kb/general/ad-filtering/create-own-filters/) and [here](https://help.adblockplus.org/hc/en-us/articles/360062733293-How-to-write-filters)

- `@@||*example.com` is the same as `@@||example.com` which will allow `sub.example.com`

typings/bcrypt/__init__.pyi
jemrobinson: I auto-generated this from the `.py` file so it included all functions. Agree that we don't need this one.

typings/bcrypt/__init__.pyi
jemrobinson: Removed in f9f57d88

typings/bcrypt/__init__.pyi
jemrobinson: Removed in f9f57d88

data_safe_haven/resources/dns_server/AdGuardHome.mustache.yaml
jemrobinson: Done in eaae180a

data_safe_haven/pulumi/components/sre_dns_server.py
jemrobinson: Done in 1860e930

data_safe_haven/functions/miscellaneous.py
JimMadge: Just curious really. I suppose I would think it's more intuitive for `abc` to be a known data object than `abc()`.

I think you could (and we absolutely should) avoid duplicationm

```Python
a = sorted({
   ...
})

b = sorted({
   "private": a
})
```

Although, maybe that is more awkward because you need to worry about the order of the definitions.

Happy to resolve this as I don't think there is a problem.

jemrobinson: Just taken a look at the code. `ordered_private_dns_zones` takes an argument `resource_type` so you can get the private DNS zones for e.g. just `Azure Automation` if you need. So this one has to be a function. `allowed_dns_lookups` doesn't need to be, but I based it on the previous function. Would you like to see `allowed_dns_lookups` as a plain list instead?

data_safe_haven/pulumi/components/sre_dns_server.py
JimMadge: Hardcoding a salt feels like a very bad idea.

Hmm, this is tricky. My idea was Pulumi secrets for both,

- plain text admin password
- salted/encrypted password (generated once, then just fetched from Pulumi secrets)

But what we really want is for the salted/encrypted password to **only** change when the plain text password is changed.
That feels quite similar to what Pulumi does, don't update the infrastructure if you don't need to. I wonder how difficult it is to create our own Pulumi object that behaves that way.

data_safe_haven/resources/dns_server/AdGuardHome.mustache.yaml
JimMadge: That docs page is better than the wiki one I linked :smile:, we should use that instead.

Is `@@||*example.com` an explicit example? I didn't read the docs thoroughly.

As long as that has been tested and we understand I think that is fine.

The important thing is, we list our domains with and without leading wildcards (`example.com` and `*.example.com`) so we want both of those to behave as expected.
Maybe we should add a comment to those lists to explain any restrictions.

data_safe_haven/pulumi/components/sre_dns_server.py
jemrobinson: ~What do you think about this possibility? 0bb1d0ad. It will always produce the same salt for the same password.~

Actually a better option is 56e7b8e2 which uses the same salt-generating algorithm but seeds it with something other that the salt (in this case the stack name).

data_safe_haven/resources/dns_server/AdGuardHome.mustache.yaml
jemrobinson: I _think_ `@||*.example.com` will allow subdomains of `example.com` but not `example.com` itself. We only seem to use one *. rule (`*.clamav.net`) and I think we can remove it for simplicity if you prefer?

jemrobinson: Dropped wildcarded rules in 9ac839a

data_safe_haven/functions/miscellaneous.py
JimMadge: That sounds sensible, clearer than lots of dict slices over the code.

Happy to leave this as it is.

data_safe_haven/pulumi/components/sre_dns_server.py
JimMadge: This isn't really my area. I think having a known salt makes a brute force attack significantly easier (if you can see the hashed password(s)).
That might not really be one of our top concerns here though.

I did think, would another option be to hold the salt as a pulumi secret,

1. Check if salt secret exists
2. If not create a random salt
3. Use plain text password and salt to create bcrypt password (now reproducible as the salt is fixed)

I suppose that is the same as 56e7b8e except the salt is unknown.

jemrobinson: Switched to storing salt as a Pulumi secret in d6d84451.

JimMadge (approved):

PR #1614, opened 2023-09-20
Pulumi: Add smoke tests

### LINKED ISSUE #1603: Pulumi: add smoke tests

jemrobinson:

- Add some basic smoke tests (**NB.** these use the old version of `bats` available in the Ubuntu repos which has fewer features)
- Add a `FileUpload` provider to allow files to be uploaded to VMs that cannot access e.g. `raw.githubusercontent.com`
- Add an AzureApi call that runs a remote script, but catches failures caused by existing scripts and retries

Closes #1603

### :microscope: Tests

Tested on a local deployment


JimMadge (commented):
Looks good.

The Azure API doesn't have a more direct way to write a file to a disk mounted to a VM?
I suppose we use the API here as the machines will not generally be accessible by SSH?

jemrobinson:
> The Azure API doesn't have a more direct way to write a file to a disk mounted to a VM?

No, weirdly not. Maybe they didn't want to write something that would work cross-OS?

> I suppose we use the API here as the machines will not generally be accessible by SSH?

That's right - these VMs have a *very* limited allowed set of connections in and out which only cover internal DSH connections and platform level stuff (like this) which bypasses the NSG rules.

PR #1617, opened 2023-09-22
Pulumi: Improve login flow

### LINKED ISSUE #1511: Subprocess calls etc. in PulumiStack
Working through #1506 warnings were raised about possible shell injection/other attacks for subprocess calls. Having a look, I don't think I like what is happening in the `PulumiStack` class.
For example,

https://github.com/alan-turing-institute/data-safe-haven/blob/f0a890632725e2b6883c8997787294b824caa1cb/data_safe_haven/pulumi/pulumi_stack.py#L201-L226

- Nested try blocks are hard to read
- Flow control relies on an exception being raised in `self.whoami()`
- `self.whoami()` is a wrapper for `pulumi whoami`, there is a method for this in the Pulumi SDK
- We are committing to being ABI(?) compatible with Pulumi (on top of being API compatible in other points in the code)
- We hard code some decisions like the pulumi backend, removing flexibility

It feels like we have added code, and taken on maintenance burden, for something that is less flexible than saying "Run `az login` and `pulumi login` before running `dsh`" and just checking that in the code?

Does much of the rest of the code rely on these things? I'm worried we will end up with something quite hard to maintain.

JimMadge:

- Ensure Azure CLI credential is used
- Prompt to confirm the correct Azure CLI account is used
- Still use a subprocess call to get Azure CLI credential information (not sure if there is a better way to do this)
- Prompt to confirm the correct Pulumi account is used
- Attempt to log into the Pulumi account corresponding to the DSH backend
- Also uses subprocess calls
- Subprocess call _should_ be safer as they don't take arguments from the user and use absolute paths

Closes #1511

### :microscope: Tests

Tested locally by triggering the new prompts


jemrobinson:
This looks very nice. Is it ready to test?

data_safe_haven/infrastructure/stack_manager.py
```diff
@@ -1,24 +1,113 @@
...
+                stderr=subprocess.PIPE,
+                encoding="utf8",
+                env=self.env
+            )
+        except subprocess.CalledProcessError as exc:
+            msg = f"Logging into Pulumi failed.\n{exc}\n{result}"
+            raise DataSafeHavenPulumiError(msg) from exc
+
+        print(result)
+        return typer.confirm("Is this the Pulumi account you expect?\n")
```
jemrobinson: There's a wrapper for this in logger (which means it will also get logged to whichever logging handlers are configured, not just written to stdout).

data_safe_haven/infrastructure/stack_manager.py
```diff
@@ -1,24 +1,113 @@
...
+    azure_api = AzureApi(config.subscription_name)
+    backend_storage_account_keys = azure_api.get_storage_account_keys(
+        config.backend.resource_group_name,
+        config.backend.storage_account_name,
+    )
+    return {
+        "AZURE_STORAGE_ACCOUNT": config.backend.storage_account_name,
+        "AZURE_STORAGE_KEY": str(backend_storage_account_keys[0].value),
+        "AZURE_KEYVAULT_AUTH_VIA_CLI": "true",
+    }
```
jemrobinson: Is this useful anywhere outside PulumiAccount? Feels like it could be part of that class.

data_safe_haven/infrastructure/stack_manager.py
```diff
@@ -1,24 +1,113 @@
...
 from data_safe_haven.utility import LoggingSingleton
 
 
+def handle_pulumi_login(config) -> None:
+    account = PulumiAccount(config)
+
+    if not account.confirm():
+        account.login()
+        if not account.confirm():
+            raise typer.Exit()
```
jemrobinson: Is it better to raise one of our own exceptions here or does `typer.Exit()` give a cleaner shutdown?

data_safe_haven/infrastructure/stack_manager.py
```diff
@@ -1,24 +1,113 @@
...
+                [
+                    self.path,
+                    "login",
+                    f"azblob://{self.cfg.pulumi.storage_container_name}",
+                ],
+                stderr=subprocess.PIPE,
+                encoding="utf8",
+                env=self.env
+            )
+        except subprocess.CalledProcessError as exc:
```
jemrobinson: You also need to catch `FileNotFoundError` if `pulumi` does not exist.

jemrobinson (commented):
Some initial thoughts

data_safe_haven/infrastructure/stack_manager.py
JimMadge: Yes, it is also needed in `create_or_select_stack` in the `StackManager`.

So, we could create an instance of `PulumiAccount` there. Not really sure which is cleaner.

data_safe_haven/infrastructure/stack_manager.py
JimMadge: At the moment it gives a 0 return code (which feels a bit odd as it needs to be raised).
It might be more sensible to give a non-zero return code here though.

data_safe_haven/infrastructure/stack_manager.py
jemrobinson: I'm generally a fan of Smalltalk-style classes which ask each other things. So if we think that the `pulumi_env` is conceptually *part of* the `PulumiAccount` then the way to get hold of it should be to ask `PulumiAccount` "what is the value of `pulumi_env`" through a method like `env`.

data_safe_haven/infrastructure/stack_manager.py
JimMadge: That should be covered here.

https://github.com/alan-turing-institute/data-safe-haven/blob/449508a74c775b1ba4436ddc9c3174a8948a3909/data_safe_haven/infrastructure/stack_manager.py#L48-L51

data_safe_haven/infrastructure/stack_manager.py
JimMadge: fixed in 8268fbec

craddm (changes requested):
When attempting to deploy a new SHM, I hit an (extra long) error, which indicates a failure when it runs `pulumi whoami`
 
<details>

```
2023-10-20 13:51:31 [    INFO] Reading project settings from '/home/deploydsh/.config/data_safe_haven/config.yaml'.                                                                                                                                                      backend_settings.py:110
2023-10-20 13:51:33 [    INFO] AzureAD tenant ID will be cb94a6f6-ef7a-42ab-bcad-4f0b887cfd3e.                                                                                                                                                                                     config.py:176
2023-10-20 13:51:33 [    INFO] Admin email address will be mcraddock@turing.ac.uk.                                                                                                                                                                                                 config.py:183
2023-10-20 13:51:33 [    INFO] IP addresses used by administrators will be ['193.60.220.253/32'].                                                                                                                                                                                  config.py:190
2023-10-20 13:51:33 [    INFO] Fully-qualified domain name will be green.develop.turingsafehaven.ac.uk.                                                                                                                                                                            config.py:197
2023-10-20 13:51:33 [    INFO] Timezone will be Europe/London.                                                                                                                                                                                                                     config.py:204
2023-10-20 13:51:34 [    INFO] name: mcraddock@turing.ac.uk (id: 3f1a8e26-eae2-4539-952a-0a6184ec248a                                                                                                                                                                            azure_cli.py:67
tenant: 4395f4a7-e455-4f95-8a9f-1fbaef6384f9                                                                                                                                                                                                                                                    
Is this the Azure account you expect?
 [y/N]: y
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /workspaces/data-safe-haven/data_safe_haven/infrastructure/stack_manager.py:62 in confirm        │
│                                                                                                  │
│    59 │   │   # do this with a minimal dummy stack which also works with the Azure backend.      │
│    60 │   │   # A subprocess call is used here.                                                  │
│    61 │   │   try:                                                                               │
│ ❱  62 │   │   │   result = subprocess.check_output(                                              │
│    63 │   │   │   │   [self.path, "whoami", "--verbose"],                                        │
│    64 │   │   │   │   stderr=subprocess.PIPE,                                                    │
│    65 │   │   │   │   encoding="utf8",                                                           │
│                                                                                                  │
│ ╭─────────────────────────────────────────── locals ───────────────────────────────────────────╮ │
│ │ self = <data_safe_haven.infrastructure.stack_manager.PulumiAccount object at 0xffff8df8c8b0> │ │
│ ╰──────────────────────────────────────────────────────────────────────────────────────────────╯ │
│                                                                                                  │
│ /usr/local/lib/python3.10/subprocess.py:421 in check_output                                      │
│                                                                                                  │
│    418 │   │   │   empty = b''                                                                   │
│    419 │   │   kwargs['input'] = empty                                                           │
│    420 │                                                                                         │
│ ❱  421 │   return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,                      │
│    422 │   │   │      **kwargs).stdout                                                           │
│    423                                                                                           │
│    424                                                                                           │
│                                                                                                  │
│ ╭─────────────────────────────────────────── locals ───────────────────────────────────────────╮ │
│ │    kwargs = {                                                                                │ │
│ │             │   'stderr': -1,                                                                │ │
│ │             │   'encoding': 'utf8',                                                          │ │
│ │             │   'env': {                                                                     │ │
│ │             │   │   'AZURE_STORAGE_ACCOUNT': 'shmgreenbackend',                              │ │
│ │             │   │   'AZURE_STORAGE_KEY':                                                     │ │
│ │             'gkUkFf9cFNOJt4m2OkwmRZlsYYW3eNwvZW2CiCOPiJdU1gW4620a3tH7F1YPp+/LEzxEWrVnbikP+A… │ │
│ │             │   │   'AZURE_KEYVAULT_AUTH_VIA_CLI': 'true'                                    │ │
│ │             │   }                                                                            │ │
│ │             }                                                                                │ │
│ │ popenargs = (['/home/deploydsh/.pulumi/bin/pulumi', 'whoami', '--verbose'],)                 │ │
│ │   timeout = None                                                                             │ │
│ ╰──────────────────────────────────────────────────────────────────────────────────────────────╯ │
│                                                                                                  │
│ /usr/local/lib/python3.10/subprocess.py:526 in run                                               │
│                                                                                                  │
│    523 │   │   │   raise                                                                         │
│    524 │   │   retcode = process.poll()                                                          │
│    525 │   │   if check and retcode:                                                             │
│ ❱  526 │   │   │   raise CalledProcessError(retcode, process.args,                               │
│    527 │   │   │   │   │   │   │   │   │    output=stdout, stderr=stderr)                        │
│    528 │   return CompletedProcess(process.args, retcode, stdout, stderr)                        │
│    529                                                                                           │
│                                                                                                  │
│ ╭─────────────────────────────────────────── locals ───────────────────────────────────────────╮ │
│ │ capture_output = False                                                                       │ │
│ │          check = True                                                                        │ │
│ │          input = None                                                                        │ │
│ │         kwargs = {                                                                           │ │
│ │                  │   'stdout': -1,                                                           │ │
│ │                  │   'stderr': -1,                                                           │ │
│ │                  │   'encoding': 'utf8',                                                     │ │
│ │                  │   'env': {                                                                │ │
│ │                  │   │   'AZURE_STORAGE_ACCOUNT': 'shmgreenbackend',                         │ │
│ │                  │   │   'AZURE_STORAGE_KEY':                                                │ │
│ │                  'gkUkFf9cFNOJt4m2OkwmRZlsYYW3eNwvZW2CiCOPiJdU1gW4620a3tH7F1YPp+/LEzxEWrVnb… │ │
│ │                  │   │   'AZURE_KEYVAULT_AUTH_VIA_CLI': 'true'                               │ │
│ │                  │   }                                                                       │ │
│ │                  }                                                                           │ │
│ │      popenargs = (['/home/deploydsh/.pulumi/bin/pulumi', 'whoami', '--verbose'],)            │ │
│ │        process = <Popen: returncode: 255 args: ['/home/deploydsh/.pulumi/bin/pulumi',        │ │
│ │                  'whoami...>                                                                 │ │
│ │        retcode = 255                                                                         │ │
│ │         stderr = 'error: PULUMI_ACCESS_TOKEN must be set for login during non-interactive    │ │
│ │                  CLI sess'+5                                                                 │ │
│ │         stdout = ''                                                                          │ │
│ │        timeout = None                                                                        │ │
│ ╰──────────────────────────────────────────────────────────────────────────────────────────────╯ │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
CalledProcessError: Command '['/home/deploydsh/.pulumi/bin/pulumi', 'whoami', '--verbose']' returned non-zero exit status 255.

During handling of the above exception, another exception occurred:

╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /workspaces/data-safe-haven/data_safe_haven/commands/deploy.py:83 in shm                         │
│                                                                                                  │
│    80 │   ] = None,                                                                              │
│    81 ) -> None:                                                                                 │
│    82 │   """Deploy a Safe Haven Management component"""                                         │
│ ❱  83 │   deploy_shm(                                                                            │
│    84 │   │   aad_tenant_id=aad_tenant_id,                                                       │
│    85 │   │   admin_email_address=admin_email_address,                                           │
│    86 │   │   admin_ip_addresses=admin_ip_addresses,                                             │
│                                                                                                  │
│ ╭─────────────────────────── locals ───────────────────────────╮                                 │
│ │       aad_tenant_id = 'cb94a6f6-ef7a-42ab-bcad-4f0b887cfd3e' │                                 │
│ │ admin_email_address = 'mcraddock@turing.ac.uk'               │                                 │
│ │  admin_ip_addresses = ['193.60.220.253/32']                  │                                 │
│ │              domain = 'green.develop.turingsafehaven.ac.uk'  │                                 │
│ │               force = None                                   │                                 │
│ │            timezone = 'Europe/London'                        │                                 │
│ ╰──────────────────────────────────────────────────────────────╯                                 │
│                                                                                                  │
│ /workspaces/data-safe-haven/data_safe_haven/commands/deploy_shm.py:43 in deploy_shm              │
│                                                                                                  │
│   40 │   │   verification_record = graph_api.add_custom_domain(config.shm.fqdn)                  │
│   41 │   │                                                                                       │
│   42 │   │   # Initialise Pulumi stack                                                           │
│ ❱ 43 │   │   PulumiAccount(config).handle_login()                                                │
│   44 │   │   stack = SHMStackManager(config)                                                     │
│   45 │   │   # Set Azure options                                                                 │
│   46 │   │   stack.add_option("azure-native:location", config.azure.location, replace=False)     │
│                                                                                                  │
│ ╭─────────────────────────────────────────── locals ───────────────────────────────────────────╮ │
│ │       aad_tenant_id = 'cb94a6f6-ef7a-42ab-bcad-4f0b887cfd3e'                                 │ │
│ │ admin_email_address = 'mcraddock@turing.ac.uk'                                               │ │
│ │  admin_ip_addresses = ['193.60.220.253/32']                                                  │ │
│ │              config = <data_safe_haven.config.config.Config object at 0xffff8df4b490>        │ │
│ │               force = None                                                                   │ │
│ │                fqdn = 'green.develop.turingsafehaven.ac.uk'                                  │ │
│ │           graph_api = <data_safe_haven.external.api.graph_api.GraphApi object at             │ │
│ │                       0xffff8df8c820>                                                        │ │
│ │            timezone = 'Europe/London'                                                        │ │
│ │ verification_record = 'MS='                                                                  │ │
│ ╰──────────────────────────────────────────────────────────────────────────────────────────────╯ │
│                                                                                                  │
│ /workspaces/data-safe-haven/data_safe_haven/infrastructure/stack_manager.py:102 in handle_login  │
│                                                                                                  │
│    99 │                                                                                          │
│   100 │   def handle_login(self) -> None:                                                        │
│   101 │   │   """Ensure the user is using the DSH Pulumi backend"""                              │
│ ❱ 102 │   │   if not self.confirm():                                                             │
│   103 │   │   │   msg = (                                                                        │
│   104 │   │   │   │   "Attempting to login to Pulumi account using"                              │
│   105 │   │   │   │   f" container [green]{self.cfg.pulumi.storage_container_name}[/]"           │
│                                                                                                  │
│ ╭─────────────────────────────────────────── locals ───────────────────────────────────────────╮ │
│ │ self = <data_safe_haven.infrastructure.stack_manager.PulumiAccount object at 0xffff8df8c8b0> │ │
│ ╰──────────────────────────────────────────────────────────────────────────────────────────────╯ │
│                                                                                                  │
│ /workspaces/data-safe-haven/data_safe_haven/infrastructure/stack_manager.py:69 in confirm        │
│                                                                                                  │
│    66 │   │   │   │   env=self.env,                                                              │
│    67 │   │   │   )                                                                              │
│    68 │   │   except subprocess.CalledProcessError as exc:                                       │
│ ❱  69 │   │   │   msg = f"Logging into Pulumi failed.\n{exc}\n{result}"                          │
│    70 │   │   │   raise DataSafeHavenPulumiError(msg) from exc                                   │
│    71 │   │                                                                                      │
│    72 │   │   self.logger.info("Confirming Pulumi account details")                              │
│                                                                                                  │
│ ╭─────────────────────────────────────────── locals ───────────────────────────────────────────╮ │
│ │ self = <data_safe_haven.infrastructure.stack_manager.PulumiAccount object at 0xffff8df8c8b0> │ │
│ ╰──────────────────────────────────────────────────────────────────────────────────────────────╯ │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
UnboundLocalError: local variable 'result' referenced before assignment
```

</details>

I think what happens is that `handle_login()` tries to check if the user is logged in to the right pulumi account, using `confirm()`. `confirm()` then tries to run `pulumi whoami`. If you aren't already logged in to pulumi, running `pulumi whoami` makes it try to log you in. By default, it'll try to log you in to the Pulumi Cloud, which then fails - the relevant error for that is 

```
stderr = 'error: PULUMI_ACCESS_TOKEN must be set for login during non-interactive CLI sess'+5
```

`PULUMI_ACCESS_TOKEN` shouldn't be necessary when using a non-pulumi backed, like the Azure one we're using, but at this point `whoami` doesn't know what backend we want to use - we haven't run `login()` yet. 

I think the current logic'll catch people being logged in to the wrong account but doesn't correctly handle people who aren't logged in at all.

craddm:
>PULUMI_ACCESS_TOKEN shouldn't be necessary when using a non-pulumi backed, like the Azure one we're using, but at this point whoami doesn't know what backend we want to use - we haven't run login() yet.`

On this point: one way around this is to set the `PULUMI_BACKEND_URL`. Formatting is a bit off below, but you can set that like so:

```
def env(self) -> dict[str, Any]:
    """Get necessary Pulumi environment variables"""
        if not self.env_:
            azure_api = AzureApi(self.cfg.subscription_name)
            backend_storage_account_keys = azure_api.get_storage_account_keys(
                self.cfg.backend.resource_group_name,
                self.cfg.backend.storage_account_name,
            )
            self.env_ = {
                "AZURE_STORAGE_ACCOUNT": self.cfg.backend.storage_account_name,
                "AZURE_STORAGE_KEY": str(backend_storage_account_keys[0].value),
                "AZURE_KEYVAULT_AUTH_VIA_CLI": "true",
                "PULUMI_BACKEND_URL": f"azblob://{self.cfg.pulumi.storage_container_name}"
            }
        return self.env_
```

`pulumi whoami` then knows you want to use the Azure backend, not Pulumi Cloud. I haven't tested this thoroughly, so I'm not sure if it'll have any unintended consequences, but changing the above in `stack_manager.py` prevented the error I was having above. Thinking more broadly, this might hint at a way you could support other backends, as you'd be able to point pulumi in the right direction by setting this environment variable (but that'd require more thinking through!)

JimMadge:
Interesting. Setting that environment variable makes `whoami` query that backend irrespective of which (if any) you are logged in to, and doesn't force logout/login.
(_I.e._ I am logged into Pulumi cloud before _and_ after running `dsh ...` but `dsh` shows the Azure backend)
If that works with other Pulumi commands too (through the automation API) that would be a better way to go about this, rather than making a user change which account they are connected to.

(Maybe Pulumi login is really just about getting a local token so you don't have to authenticate/specify with every command and the env variables can override this)

JimMadge:
@craddm Seems to work. Setting the appropriate environment variables and passing them to the Pulumi automation API workspace means you don't need to use `pulumi login`.

You can run `dsh` command using the Azure backend without changing your Pulumi CLI account.

This is nice actually as it gives us much less code to maintain/make ABI compatible.

data_safe_haven/external/api/azure_cli.py
```diff
@@ -1,44 +1,73 @@
...
+                tenant_id=result_dict.get("tenantId"),
+            )
+
+        return self._account
+
+    def confirm(self) -> None:
+        """Prompt user to confirm the Azure CLI account is correct"""
+        account = self.account
+        self.logger.info(
+            f"name: {account.name} (id: {account.id_}\ntenant: {account.tenant_id}"
```
craddm:
```suggestion
            f"name: {account.name} (id: {account.id_}\ntenant: {account.tenant_id})"
```

craddm:
Screenshot 2023-10-23 at 16 22 36

Is there a way to track whether the Az account has been confirmed already? Getting asked to confirm multiple times in the same call. Maybe we can cache it and check if it changed, or just set a variable tracking whether it's been confirmed or not.

JimMadge:
@craddm Good catch, I thought I had fixed that. I have make the AzureCli class a singleton so it shouldn't happen now.

Oops, didn't see that change didn't push. It has been now.

craddm (approved):

craddm:
LGTM 👍

PR #1624, opened 2023-10-02
Improve Pulumi error messages

### LINKED ISSUE #1623: Pulumi: misleading error messages
When running `dsh init`, if the `--name` parameter is not provided, the error message says that the `--deployment-name` parameter is not provided. This was the old name for that parameter. 

Another misleading error message

When attempting to use commands other then `dsh init` before having run `dsh init`, the error messages states that the `--deployment-name` parameter is missing. Providing the parameter correctly errors, since `--deployment-name` parameter is for `dsh init`, not for the other commands.

another misleading error

craddm:

Fixes incorrect error messages for the deployment `name` argument passed to `dsh init`
Adds an error when the backend has not been initialised before attempting to run commands that rely on it (e.g. `dsh admin list-users`)

Note - this PR also updates the relevant .devcontainer to Pulumi version `3.80.0`

Closes #1623 

### :microscope: Tests

Tested locally:
better error

another better error

### DIFF
--- a/.devcontainer/python/Dockerfile
+++ b/.devcontainer/python/Dockerfile
@@ -3,9 +3,8 @@ FROM python:${VARIANT}-buster
 
 # Set package versions
 ARG AZURE_CLI_VERSION="2.42.0"
-ARG PWSH_VERSION="7.2.6"
-ARG PULUMI_VERSION="3.45.0"
-ARG POETRY_VERSION="1.2.2"
+ARG PWSH_VERSION="7.3.6"
+ARG PULUMI_VERSION="3.80.0"
 
 RUN apt-get update \
     && export DEBIAN_FRONTEND=noninteractive \
@@ -22,6 +21,7 @@ ARG TARGETARCH
 # Standard install method currently does not support ARM64
 # Use pip instead - https://github.com/Azure/azure-cli/issues/22875
 RUN pip3 install azure-cli==${AZURE_CLI_VERSION}
+RUN pip3 install hatch
 
 # Install Powershell
 # Pull different binaries from Github depending on system architecture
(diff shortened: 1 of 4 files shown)


jemrobinson:
It looks like this has been based off an older branch of `python-migration`. Possibly this is because you used a newer version of `develop`?

craddm:
It shouldn't be - I updated my local copy of python-migration and merged it into this branch just before making the PR

Edit: ah, which makes it a mess and hard to see what I actually changed, right

jemrobinson:
I've merged `develop` into the upstream `python-migration` ~and now this looks more sensible.~ hmm, no it doesn't. Can you `git reset --hard 283f0e93e` to drop your last merge commit?

jemrobinson (approved):
LGTM

.devcontainer/python/Dockerfile
```diff
@@ -22,6 +21,7 @@ ARG TARGETARCH
 # Standard install method currently does not support ARM64
 # Use pip instead - https://github.com/Azure/azure-cli/issues/22875
 RUN pip3 install azure-cli==${AZURE_CLI_VERSION}
+RUN pip3 install hatch
 
 # Install Powershell
```
jemrobinson: Do we need Powershell in this devcontainer?

jemrobinson (approved):
LGTM

.devcontainer/python/Dockerfile
craddm: No - fair point. Actually, do we need sphinx in there at the moment either? Or should I just leave it in for when we *do* need it? And what version of Azure CLI are you using at the moment?

jemrobinson: You're right that the requirements needed for building the docs aren't (currently) in the python-migration `pyproject.toml`. They probably ought to be added there once we've merged `python-migration` into develop, but until then I think it's fine to install the Sphinx `requirements.txt`.

PR #1629, opened 2023-10-04
Use Pulumi random provider

### LINKED ISSUE #1628: Pulumi: secret management

Replace current management of Pulumi secrets

Instead of generating these in our code and providing them as `Pulumi.config` secure values, we should use the Pulumi random provider.

jemrobinson:

Replace custom secret generation with Pulumi random provider.

Note that the following secrets were not replaced:
- [SHM `password-domain-ldap-searcher`](#1645) (as this is needed by both SHM and SRE and replacing it would involve a rewrite of how we pass secrets between stacks)
- SHM `verification-azuread-custom-domain` (could possibly be passed as an argument to the deployment code in some way)
- [SRE `password-domain-ldap-searcher`](#1645) (copied from SHM stack)
- [SRE `salt-dns-server-admin`](#1644) (this cannot trivially be generated using Pulumi's random provider)

Closes #1628 

### :microscope: Tests

~Not tested due to #1641.~ Tested on SHM green and SREs leonardo and donatello.

### DIFF
--- a/data_safe_haven/commands/deploy_shm.py
+++ b/data_safe_haven/commands/deploy_shm.py
@@ -50,7 +50,6 @@ def deploy_shm(
         )
         stack.add_option("azure-native:tenantId", config.azure.tenant_id, replace=False)
         # Add necessary secrets
-        stack.add_secret("password-domain-admin", password(20), replace=False)
         stack.add_secret(
             "password-domain-azure-ad-connect", password(20), replace=False
         )
--- a/data_safe_haven/commands/deploy_shm.py
+++ b/data_safe_haven/commands/deploy_shm.py
@@ -50,9 +50,6 @@ def deploy_shm(
         )
         stack.add_option("azure-native:tenantId", config.azure.tenant_id, replace=False)
         # Add necessary secrets
-        stack.add_secret(
-            "password-domain-computer-manager", password(20), replace=False
-        )
         stack.add_secret("password-domain-ldap-searcher", password(20), replace=False)
         stack.add_secret(
             "password-update-server-linux-admin", password(20), replace=False
--- a/data_safe_haven/commands/deploy_shm.py
+++ b/data_safe_haven/commands/deploy_shm.py
@@ -51,9 +51,6 @@ def deploy_shm(
         stack.add_option("azure-native:tenantId", config.azure.tenant_id, replace=False)
         # Add necessary secrets
         stack.add_secret(
-            "password-domain-azure-ad-connect", password(20), replace=False
-        )
-        stack.add_secret(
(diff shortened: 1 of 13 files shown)


JimMadge:
This looks like the right idea to me.

The provider has a lot of useful features, the password resources are cryptographically secure.
I think using (and contributing) to this is likely to be better than what we make ourselves.

data_safe_haven/commands/deploy_shm.py
```diff
@@ -50,17 +50,7 @@ def deploy_shm(
...
         stack.add_option("azure-native:tenantId", config.azure.tenant_id, replace=False)
         # Add necessary secrets
-        stack.add_secret("password-domain-admin", password(20), replace=False)
-        stack.add_secret(
-            "password-domain-azure-ad-connect", password(20), replace=False
-        )
-        stack.add_secret(
-            "password-domain-computer-manager", password(20), replace=False
-        )
         stack.add_secret("password-domain-ldap-searcher", password(20), replace=False)
```
JimMadge, line 53: Is this a credential Guacamole users to find users?
Are we not able to replace it here?

JimMadge (commented):
Looks good. Definitely a move in the right direction.

It would be good to remove the need for our own password/secret generation and handling all together.

I see that you've already written about this in the PR. Perhaps we can open issues for each/all of these and give an impression of how much effort they are?
No need to block this.

JimMadge:
> SRE salt-dns-server-admin (this cannot trivially be generated using Pulumi's random provider)

Why is that?

data_safe_haven/commands/deploy_shm.py
jemrobinson: It's the password for a user account on the domain controller, so both the SHM (to create it) and the SRE (to use it) need to know what it is.

jemrobinson:
> > SRE salt-dns-server-admin (this cannot trivially be generated using Pulumi's random provider)
> 
> Why is that?

It needs to be a byte string of the form: `$prefix$rounds$output` where `prefix` is either `2a` or `2b`, `rounds` is a zero-padded integer from 4-31 inclusive and output is a 16 byte integer encoded into a base64 byte string using a non-standard version of base64. We currently use the `bcrypt` library to do this, but the `gensalt` function gets its own random number from `urandom` - it doesn't take a random number (which we could provide) as input.

So, we *could* use the Pulumi random provider but we'd need to write our own version of this function and keep it up to date with any future changes to the bcrypt protocol.

JimMadge:
I think I remember looking at the salt before, making sure it was idempotent.

Taking a quick look, I think we are handling that in the most sensible way. Create the data, store it in configuration.

PR #1649, opened 2023-10-19
Add all contributors table and instructions for how to update

edwardchalstrey1:

Adds README table and badge, plus instructions that link to https://allcontributors.org/docs/en/cli/overview


.all-contributorsrc
```diff
@@ -0,0 +1,4 @@
+{
+  "projectName": "data-safe-haven",
+  "projectOwner": "alan-turing-institute"
+}
```
JimMadge: There are quite a lot of configuration options we should look at.

In particular, can we make sure that the all contributors table is also added somewhere in the docs?

[Here](https://github.com/sa-tre/satre-specification/blob/main/.all-contributorsrc) is an example.

CONTRIBUTING.md
```diff
@@ -267,7 +266,17 @@ The following people have made one or more commits to the project:
 | Kirstie Whitaker     | [@KirstieJane](https://github.com/KirstieJane)             |
 | Rachel Winstanley    | [@rwinstanley1](https://github.com/rwinstanley1)           |
 | Warwick Wood         | [@warwick26](https://github.com/warwick26)                 |
-| Kevin Xu             | [@kevinxufs](https://github.com/kevinxufs)                 |
+| Kevin Xu             | [@kevinxufs](https://github.com/kevinxufs)                 | -->
```
JimMadge: I think we can remove this now as the all contributors table will supersede it.

README.md
```diff
@@ -32,6 +33,17 @@ There are lots of ways to contribute, not just writing code!
...
+
+<!-- ALL-CONTRIBUTORS-LIST:START - Do not remove or modify this section -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-disable -->
+
+<!-- markdownlint-restore -->
+<!-- prettier-ignore-end -->
+
+<!-- ALL-CONTRIBUTORS-LIST:END -->
+
```
JimMadge: Let's bootstrap this with all of our previous contributors.

You can use the CLI to do this.

JimMadge (changes requested):
Good idea (why haven't we done this already!?!).

If you install the CLI you should be able to test this locally.

edwardchalstrey1:
Closing this for now as per Martin's comment on slack:

> I approved and then disapproved the all-contributors bot as it needs write access to code, which we’ve decided we won’t allow for the DSH codebase as part of our controls over who can make changes to the codebase.

@JimMadge I guess if you were keen to have this then this can reopened later after a discussion with the team

JimMadge:
@edwardchalstrey1 Good point, the distinction between pinned actions and an unpinned bot is a little subtle but important.

However, I think we can still use all contributors and should reopen this PR. We can commit to using the all contributors CLI to manager the contributors table. I think that is still better than what we currently do. Any workflows we could add to ensure consistency or try and spot missing contributors would be great.

The contributing docs could be updated to say don't use the bot, but ask a maintainer.

.all-contributorsrc
edwardchalstrey1: @JimMadge I've added a page on readthedocs too. 

Can you take a look whether you think there are any other options you'd like to configure?

Also, in terms of contributions, I have just put "code" and "doc" and guessed when people were likely to have been doc editors only - might be worth taking a look at the table and seeing if you think modifications should be made

JimMadge (changes requested):
This is looking great.

A couple of things I think we should do,

- On the preview the contributors page appears under a "More" drop down in the menu bar. Can we make it more visible? Maybe there is a theme option for number of items before the drop down?
- I think we should take the approach of low threshold to contribution. There are [more contribution types](https://allcontributors.org/docs/en/emoji-key) that we could acknowledge. Would you like me to have a go at updating the contributions?
- It would be worth asking in our Slack channels for contributors we might have missed. _Especially_ contributions which don't appear in the commit history like funding, training, conversations, bug reports.

edwardchalstrey1:
> On the preview the contributors page appears under a "More" drop down in the menu bar. Can we make it more visible? Maybe there is a theme option for number of items before the drop down?

Done

README.md
```diff
@@ -53,12 +53,12 @@ See our [Code of Conduct](CODE_OF_CONDUCT.md) and our [Contributor Guide](CONTRI
...
-      <td align="center" valign="top" width="14.28%"><a href="http://helen.st/cv"><img src="https://avatars.githubusercontent.com/u/217966?v=4?s=100" width="100px;" alt="Helen Sherwood-Taylor"/><br /><sub><b>Helen Sherwood-Taylor</b></sub></a><br /><a href="#ideas-helenst" title="Ideas, Planning, & Feedback">🤔</a> <a href="#content-helenst" title="Content">🖋</a></td>
+      <td align="center" valign="top" width="14.28%"><a href="https://helen.st/cv"><img src="https://avatars.githubusercontent.com/u/217966?v=4?s=100" width="100px;" alt="Helen Sherwood-Taylor"/><br /><sub><b>Helen Sherwood-Taylor</b></sub></a><br /><a href="#ideas-helenst" title="Ideas, Planning, & Feedback">🤔</a> <a href="#content-helenst" title="Content">🖋</a></td>
```
JimMadge: This probably won't work (if there is no valid SSL cert).

Manual changes like this will always be overwritten when you run `all-contributors` too.
It would be better to exclude these urls from the check.

JimMadge (changes requested):

.github/workflows/documentation.yaml
```diff
@@ -52,6 +52,6 @@ jobs:
           make html SPHINXOPTS="-W"
       - name: Link Checker
         uses: lycheeverse/lychee-action@v1.7.0
-        with:
-          args: --config='./.lychee.toml' --no-progress './docs/build/html/**/*.html'
-          fail: true  # fail on broken links
+        run: |
+          find ./docs/build/html -type f -name '*.html' ! -path './docs/build/html/contributing/index.html' | xargs lychee --config='./.lychee.toml' --no-progress
+        fail: true  # fail on broken links
```
edwardchalstrey1: @JimMadge could you check this looks right to you? It's supposed to be running for all html pages in the build except the contributor page

JimMadge (approved):

Issue #1640, opened 2023-10-09
Pulumi: Move command line arguments to configuration


JimMadge:
## :white_check_mark: Checklist

- [x] I have searched open and closed issues for duplicates.
- [x] This is a request for a new feature in the Data Safe Haven or an upgrade to an existing feature.
- [x] The feature is still missing in the [latest version](https://github.com/alan-turing-institute/data-safe-haven/releases).
- [x] I have read through the [documentation](https://alan-turing-institute.github.io/data-safe-haven/).
- [x] This isn't an open-ended question (open a [discussion](https://github.com/alan-turing-institute/data-safe-haven/discussions) if it is).

## :strawberry: Suggested change

A number of the `dsh` commands take command line arguments which change what is deployed.
For example, `dsh deploy sre` can specify allowed IPs, workspaces, allowed packages.
There is a risk here that changes could be made to a deployed environment by accident, and to avoid making changes admins will need to find a way to share the correct arguments.

This contrasts with the shared configuration in the backend storage account.
I think we should simplify/unify this and perhaps add some functionality.
The aim should be for an admin to only have to declare the "name" of the shm/sre.

## :steam_locomotive: How could this be done?

- Move command line arguments to configuration
- Commands to set configuration
  - interactive?
  - upload/download file?
  - get/set individual options (like `git config`)
- Ask confirmation for changes or use `--force` to override

jemrobinson:
I suggest that, like `git config`, this is stored somewhere where a user *can* edit it, but is expected to use the `dsh config` commands to interact with it most of the time. We can repurpose our current `backend_settings` config file and location for this.

PR #1655, opened 2023-10-26
Add context command

### LINKED ISSUE #1651: Pulumi: Restructure/break down init

Related to #1640 

The `init` subcommand does a number of things,

- Creates Azure infrastructure
  - Storage account
    - Pulumi backend
    - Configuration store
  - Management group
  - Anything else?
- Sets some _local_ configuration
  - Which subscription, management group, shm

This means you use init for a number of logically distinct tasks

- Creating new backend infrastructure
- Switching between SHMs
- Updating configuration

I think this could be clearer by breaking these tasks down and using different sub commands.

There is also a lack of symetry with `dsh init` creating the backend infrastructure and `dsh teardown backend` destroying it, in contract to, for example, `dsh create sre` `dsh teardown sre`.

Command structure ideas,

Add `backend`/`context` and `config` commands

- dsh
  - backend/context
    - create/init
    - destroy
    - switch (switch between backends/projects like you currently would with `init`)
    - config (alternatively, this could be a top level subcommand)
  - create
  - teardown

Restructure around backend, SHM, SRE (or, bring backend into `create`/`teardown`)
(We don't need to do all of this at once)

- dsh
  - backend
    - create
    - teardown
    - switch
    - config
  - shm
  - sre

JimMadge:

- Replace Backend with Context (context to avoid confusion with Pulumi backend)
- Replace `init` and `teardown backend` commands with the `context` command group
- Add additional commands to help manage contexts (add, remove, update, show, switch, available)
- Move `init` arguments to ContextSettings configuration (editable in the file, or with `context add`, `context update`, `context remove`)
- Configuration supports multiple contexts, and switching between them
- Introduce testing framework, add tests for Context/ContextSettings classes and context commands

Context is overloaded in the codebase. There is the ContextSettings and Context (_local_ data giving the minimum information needed to connect to the shared infrastructure for managing DSH) and Context and ConfigSectionContext (a class for managing the context _infrastructure_, and corresponding configuration data used in operations later).
It would be good to clarify this to avoid confusion.

Adding tests has revealed a lot of warning are emitted when running `dsh`, mostly deprecation warnings for pulumi-azure modules.
I think they are otherwise suppressed, would be worth keeping an eye on them.

#### Todo

- [x] Fix tests in CI (they pass locally, may be due to package version differences)
- [x] Check bootstrapping contexts configuration works
- [x] Update usage guide

Closes #1651
Contributes to #1640 

### :microscope: Tests

- Creating/removing contexts
- Unit tests


data_safe_haven/README.md
```diff
@@ -9,16 +9,17 @@ Install the following requirements before starting
...
 - Run the following to initialise the deployment [approx 5 minutes]:
 
-```bash
-> dsh init
+```console
+> dsh context add ...
+> dsh context create
 ```
 
 You will be prompted for various project settings.
```
craddm: This is no longer true, everything has to be entered as flags

data_safe_haven/commands/context.py
craddm, line 1: I like this in general but it raises quite a few questions. Some of this is me perhaps overthinking it worrying about ways people might mess things up, and wondering how much we should stop them from doing that.
The docs/purpose of the functions could use a bit of clarification and checking for consistency.

We start by adding a context. When adding a context, you give a `key` - the command is `dsh context add [OPTIONS] KEY` - which is described as the "name of the context to add". But there's also a `--name` argument, which is the "human friendly" name of the deployment. For some functions you then supply the key, like `dsh context remove`. But for `dsh context switch` you supply the `name`, which is also described in the help as the `name of the context to switch to`. So sometimes KEY is the name of the context, and sometimes NAME is the name of the context. For `dsh context switch` you actually need to supply the `key` rather than the name.

1. I assume you're separating `key` and `name` deliberately - is it helpful to separate the context name and the deployment name, or is it better to just give a single name? e.g. You could set up two otherwise identical contexts with different keys, but why? 
2. I guess these will be addressed when we do the complete website docs, but the help info for `dsh context create` and `dsh context update` could provide a bit more info about what they do.
3. Is it possible to get Typer to complain if no flags are passed? e.g. `dsh context update` does nothing if no flags are set.
4. What is the workflow after updating the context? Suppose I update the `location` from `uksouth` to `ukwest` - what should happen when I run another command? If I already have something deployed, should it then deploy new things in the new location? At the moment, running, say, `dsh context create` after updating the location actually just seems to check the existing location - 
image

5. Could you add short form flags for each argument to the various `context` commands, as existed for `init`?

data_safe_haven/commands/context.py
```diff
@@ -0,0 +1,174 @@
...
+        location=location,
+        name=name,
+        subscription_name=subscription,
+    )
+    settings.write()
+
+
+@context_command_group.command()
+def remove(
+    key: Annotated[str, typer.Argument(help="Name of the context to add.")],
```
craddm:
```suggestion
    key: Annotated[str, typer.Argument(help="Name of the context to remove.")],
```

craddm: `Remove` removes the local copy, rather than the remote one, right?

craddm (changes requested):
Note that `dsh deploy shm` doesn't work after `dsh context create`, presumably because the context does not include the `aad_tenant_id`
image

JimMadge:
> Note that `dsh deploy shm` doesn't work after `dsh context create`, presumably because the context does not include the `aad_tenant_id` image

This isn't a regression, it was true before. The `--aad-tenant-id` argument is provided to `dsh deploy shm`.
That _is_ something I want to change in #1640 though.

So, _I_ would say leave that for another PR.

data_safe_haven/commands/context.py
JimMadge: Agreed about name/key and update confusion.
I've updated the help descriptions and argument names in 84dd4e2bbe0f699135aadeafdf03ed0eb9338279
Do you think that is enough? I think we need to balance keeping the CLI help messages brief, and can put a more detailed explanation in the docs.

> You could set up two otherwise identical contexts with different keys, but why?

That is true, it isn't the intention but we don't prevent a user from making conflicting/overlapping contexts.
We _could_ do that, but I think that would want to be a new issue as it is a new feature.

A context has a parameter called "name" and we refer to contexts by their keys, which must be unique. You _could_ use a parameter of each context, but it would be awkward (and more expensive) to select them `[c if c.name == name for c in contexts][0]` vs `contexts[key]`. And you would add the additional restriction that names also need to be unique, even if they are deployed to different subscriptions/locations.

> Is it possible to get Typer to complain if no flags are passed? e.g. dsh context update does nothing if no flags are set.

Probably, or at least you could raise an exception if all are `None`. I'm not sure if I see a strong advantage, I think saying "update nothing" and nothing changing makes sense.

> What is the workflow after updating the context? Suppose I update the location from uksouth to ukwest - what should happen when I run another command? If I already have something deployed, should it then deploy new things in the new location? At the moment, running, say, dsh context create after updating the location actually just seems to check the existing location

I think that must also have been true before as I haven't touched that code.
It would be nice if `context create` were idempotent, but as we are using the Azure Python SDK if would be a fair bit of effort.
I've updated the description to say that `update` updates the settings specifically.
If we want to change/explain this I think it can be a new issue.

> Could you add short form flags for each argument to the various context commands, as existed for init?

Could do. I intentionally didn't though.
I feel that the choices have fairly large consequences, so using clear names rather than letters (or even worse, positional arguments!) helps make sure we understand what each argument is.

craddm: > Agreed about name/key and update confusion. I've updated the help descriptions and argument names in [84dd4e2](https://github.com/alan-turing-institute/data-safe-haven/commit/84dd4e2bbe0f699135aadeafdf03ed0eb9338279) Do you think that is enough? I think we need to balance keeping the CLI help messages brief, and can put a more detailed explanation in the docs.

Helps!

> > You could set up two otherwise identical contexts with different keys, but why?
> 
> That is true, it isn't the intention but we don't prevent a user from making conflicting/overlapping contexts. We _could_ do that, but I think that would want to be a new issue as it is a new feature.
> 
> A context has a parameter called "name" and we refer to contexts by their keys, which must be unique. You _could_ use a parameter of each context, but it would be awkward (and more expensive) to select them `[c if c.name == name for c in contexts][0]` vs `contexts[key]`. And you would add the additional restriction that names also need to be unique, even if they are deployed to different subscriptions/locations.

More awkward/expensive may be true, but cost doesn't seem like an issue here really - it's just searching through a very small list, once in a while. In the context of a deployment, it's the blink of an eye. 
 
> > Is it possible to get Typer to complain if no flags are passed? e.g. dsh context update does nothing if no flags are set.
> 
> Probably, or at least you could raise an exception if all are `None`. I'm not sure if I see a strong advantage, I think saying "update nothing" and nothing changing makes sense.

Perhaps it's me - I feel like it should be explicit that nothing is changing. Why write a new config file if nothing has changed? Not something I feel strongly about I guess.

> > What is the workflow after updating the context? Suppose I update the location from uksouth to ukwest - what should happen when I run another command? If I already have something deployed, should it then deploy new things in the new location? At the moment, running, say, dsh context create after updating the location actually just seems to check the existing location
> 
> I think that must also have been true before as I haven't touched that code. It would be nice if `context create` were idempotent, but as we are using the Azure Python SDK if would be a fair bit of effort. I've updated the description to say that `update` updates the settings specifically. If we want to change/explain this I think it can be a new issue.

I still just think it's quite confusing. If I run `dsh context add <foo>`, then run `dsh context create`, then run `dsh context update <foo>`, then run `dsh context create` again, nothing changes: 

image

It reads project settings from the context file, but does nothing with them. It just uses the config that already exists without changing it, then re-uploads it at the end, unchanged. So it seems `dsh context update` is only useful before `dsh context create`. Which isn't necessarily a problem, just, again, feel like it should be more explicit.

> > Could you add short form flags for each argument to the various context commands, as existed for init?
> 
> Could do. I intentionally didn't though. I feel that the choices have fairly large consequences, so using clear names rather than letters (or even worse, positional arguments!) helps make sure we understand what each argument is.

On reflection, I think I'm ok with this given we're mostly expecting these arguments to get recorded in config files now anyway.

JimMadge: Ok, can you give a steer on what (if anything) you think should be changed at how?

Some of these things are (I feel) outside the scope of this issue, being in code that wasn't touched here. We can open issues if we feel like those are worth doing though.

I'm fairly strongly objected to removing keys for contexts. Removing a unique identifier feels like asking for trouble.

craddm: So I think what I'm trying to get my head round is what is changing the local config file meant to achieve? So if the purpose here is just to get a command in that changes the local config file, the code as is undoubtedly does that. 

Whether what I'm wondering about it involves changes to *this* code or *other* code depends a little on what changing the config file is supposed to achieve. Currently the only use for changing the local config file seems to be changing it when nothing is deployed or before tearing down and redeploying existing resources. 

I feel like either this code should say something like "You already have resources deployed, changes made here will not have an effect until the resources are deleted and redeployed", or the other code should say something like "The local config differs from existing resources, maybe you should redeploy". Or perhaps these are things that should be in accompanying documentation.

I think I'm just trying to work out when it is *meant* to be useful to change the local config file in this way, and thus whether it's useful for the new code here to explicitly inform people that what they are doing needs additional steps to be actually useful.

JimMadge: > So if the purpose here is just to get a command in that changes the local config file, the code as is undoubtedly does that.

Yes, just that and nothing more :smile:.

The way I'm seeing this is like git, git remotes in particular.
The context configuration file is a specification of what contexts exist and enough information to connect to them.
Similar to how git remotes are a record of repositories.
I think it is also important to understand that these settings are **only** for the context infrastructure, not for the TRE itself.

I don't think we should want editing the record of contexts to make changes to any infrastructure, or create warnings?
Also, I'm not certain it is possible.
How do you know if resources were already deployed or if local differs from remote?
All the program knows is the data in the contexts file.

I could see the argument to just remove the update command as it might not be much use.
Maybe it is only really useful in some rare situations like "I made one mistake when running `dsh context create`" or "I decided to change a parameter before deploying the context infrastructure".

craddm (approved):
Agreed that this does what it needs to do for now!

Issue #1658, opened 2023-11-01
Pulumi: Use Pydantic


JimMadge:
## :white_check_mark: Checklist

- [x] I have searched open and closed issues for duplicates.
- [x] This is a request for a new feature in the Data Safe Haven or an upgrade to an existing feature.
- [x] The feature is still missing in the [latest version](https://github.com/alan-turing-institute/data-safe-haven/releases).
- [x] I have read through the [documentation](https://alan-turing-institute.github.io/data-safe-haven/).
- [x] This isn't an open-ended question (open a [discussion](https://github.com/alan-turing-institute/data-safe-haven/discussions) if it is).

## :strawberry: Suggested change

Pydantic provides type validation, value validation and serialisation for classes deriving from it's `BaseModel` class or `dataclass` wrapper.

This provides functionality currently coming from a combination of `chili` and self written code (and in a PR I'm currently working on `schema`).
Using Pydantic would unify these things across the codebase.
It would be useful for validating data in a number of places including `config.py` and `{backend,context}_settings.py`.
In particular, class methods to read from a dict are automatically created, and could be used to make simple class methods to read from a file.

I've spent a fair bit of time trying to use `chili` for deserialising/validating the contexts configuration file in #1655.
It isn't very nice to work with,

- Using a `Mapper` breaks checks for missing keys
- Missing key checks don't seem to happen for nested models
- Need to write you own classmethods for reading/writing
- Encoder/Decoder have to be defined outside of the class

I think I will stop trying and plan to move both to Pydantic later.

JimMadge:
Closed in #1661

PR #1674, opened 2023-11-28
Add basic config commands

JimMadge:

Adds a minimal set of config commands to allow the program to work with the new Config class in #1658.


craddm:
Ran out of time to do a full review right now, but got as far as creating the template. Tried to upload an incorrectly completed one and failed, sensibly.

craddm:
Ok, so as of now these functions do what they are supposed to do - they allow creation/modification/upload of a config file. Couple of notes:

1. There's no validation of the `fqdn`.
2. I can't deploy an SHM after running creating the context and the config. This may not matter for this PR, though.

JimMadge:
I'll look at the fqdn validation (I think there was a function for that).

I wouldn't be surprised if it breaks deployment, certainly with the current deployment commands. I want to change that as well though. I think it would make most sense to address that in a different PR, although it does mean we will break the head of the branch. What do you think @craddm?

craddm:
I agree, probably best in another PR. No problem with that at this stage. Here's the specific error:

```
2024-01-15 10:54:13 [   ERROR]  stderr: error: error listing stacks: could not list bucket: blob (code=NotFound): GET                              cli.py:98
https://shmfishescontext.blob.core.windows.net/pulumi                                                                                                       
2024-01-15 10:54:13 [   ERROR] --------------------------------------------------------------------------------                                    cli.py:98
2024-01-15 10:54:13 [   ERROR] RESPONSE 404: 404 The specified container does not exist. 
```

JimMadge:
@craddm I've added validation for FQDNs.

craddm (approved):
LGTM

PR #1699, opened 2024-01-09
Improve handling of SRE names

### LINKED ISSUE #1687: Pulumi: Check SRE names
There are a number of commands/functions where an SRE name is required.

These tend to be processed like `sre_name = alphanumeric(sre_name).lower()`.
There is no check at that point like `assert sre_name in config.sres.keys()`.
It would be good to add a method to,

1. sanitise the provided SRE name
2. ensure that the SRE exists (or doesn't exist in the case of adding one) (related to #1651)

JimMadge:

Improve how SRE names are handled,

- Use a single function for sanitising the name from user input across the code
- Remove unused methods/lines for adding/removing SREs (this is done in the config)

Closes #1687


JimMadge:
@craddm @jemrobinson could we get reviews on this? It impacts #1716

data_safe_haven/config/config.py
```diff
@@ -284,18 +285,17 @@ def is_complete(self, *, require_sres: bool) -> bool:
             return False
         return True
 
+    @staticmethod
+    def sanitise_sre_name(name: str) -> str:
+        return alphanumeric(name).lower()
+
     def sre(self, name: str) -> ConfigSectionSRE:
         """Return the config entry for this SRE creating it if it does not exist"""
```
jemrobinson:
```suggestion
        """Return the config entry for this SRE, raising an exception if it does not exist"""
```

data_safe_haven/config/config.py
```diff
@@ -284,18 +285,17 @@ def is_complete(self, *, require_sres: bool) -> bool:
...
-            self.sres[name] = ConfigSectionSRE(index=highest_index + 1)
+            msg = f"SRE {name} does not exist"
+            raise DataSafeHavenConfigError(msg)
         return self.sres[name]
 
-    def remove_sre(self, name: str) -> None:
-        """Remove SRE config section by name"""
-        if name in self.sres.keys():
-            del self.sres[name]
-
```
jemrobinson, line 299: This is called by the `teardown sre` command. If we remove it from there, then SRE config entries will no longer be deleted from the config file on teardown.

jemrobinson (commented):
Mostly looks fine, but we should think about how to remove deleted SREs from the config file if there's no remove_sre function.

data_safe_haven/config/config.py
JimMadge: Hmm, I think that is what I was intending (apart from breaking the code where it is called). Consistent with `deploy sre` not adding anything to the configuration.

(Another option would be a more, "make configuration" -> `dsh sync` workflow, but I'm not entirely convinced that is right).

I could see some utility in bringing an SRE down without deleting it's configuration. Similar to how `pulumi destroy` works. At least in testing I think I want to redeploy the same thing from scratch fairly often.

jemrobinson: OK, if it's deliberate then I think you just need to remove the call from `teardown sre`.

JimMadge: Done in 6db60c4

jemrobinson (approved):
LGTM

Issue #1438, opened 2023-04-05
Data Administrators security group doesn't provide database access (or any access to the SRE)


edwardchalstrey1:
## :white_check_mark: Checklist

- [x] I have searched open and closed issues for duplicates.
- [x] This is a problem observed when **managing** a Data Safe Haven.
- [x] I can reproduce this with the [latest version](https://github.com/alan-turing-institute/data-safe-haven/releases).
- [x] I have read through the [documentation](https://alan-turing-institute.github.io/data-safe-haven/).
- [x] This isn't an open-ended question (open a [discussion](https://github.com/alan-turing-institute/data-safe-haven/discussions) if it is).

## :computer: System information

- **Data Safe Haven version:** 4.0.3
- **Operating system details:** mac

## :no_entry_sign: Describe the problem

A user of an SRE should have the ability to create schemas in postgres via being in the `SG <SRE ID> Data Administrators` group, however there seems to be a problem with this group. At the moment, being a member of each of these groups results in the following problems:

1. `SG <SRE ID> Research Users`: can log into the TRE and connect to postgres with `psql -h PSTGRS-<SRE ID>.<SHM ID>.turingsafehaven.ac.uk -p 5432 -d postgres`, but I try to create a schema with `create schema test;` we get `ERROR: permission denied for database postgres` which is what we expect. User was in both this group and the data admin group below when he got this.
2. `SG <SRE ID> Data Administrators` or `SG <SRE ID> System Administrators` If a user is one of these, then they can log into the SRE first login, but not the Ubuntu login, we just get the `No recent connections` screen. I think ideally both of these access levels should allow login.

When I log in as the research user and connect to the psql db and run `\du` to check the user privileges, users in the `SG <SRE ID> Data Administrators` don't show up.

I have tested this on multiple SREs so this is not specific to one particular instance.

Linked to #1392 

## :steam_locomotive: Workarounds or solutions

Connected as a superuser, and edited the permissions manually
    - On the PSTGRS VM  (logging in via Serial Console in Azure) do `sudo -u postgres psql`
    - Then `GRANT CREATE ON DATABASE postgres TO "user.name";`

edwardchalstrey1:
~~I think I know the problem here~~: `"dataAdminIpAddresses"` needs to be set in the config when `"databases"` is set, but the DSH docs don't make this clear - I will test

[...]

Issue #1392, opened 2023-02-21
Update info on how to use databases in the SRE


edwardchalstrey1:
## :white_check_mark: Checklist

- [x] I have searched open and closed issues for duplicates
- [x] This is a request for a change to documentation
- [x] This isn't a feature request or bug report
- [x] This isn't an open-ended question (open a [discussion](https://github.com/alan-turing-institute/data-safe-haven/discussions) if it is).

## :ledger: Where is the problem?

Currently there is documentation on [how to access databases in an SRE that have been set up already](https://alan-turing-institute.github.io/data-safe-haven/develop/roles/researcher/user_guide_guacamole.html#access-databases). But not on anything on creating database tables and working with psql/msql directly as oppose to via Python, R, Dbeaver.

## :memo: Required documentation changes

It would be good to explain better how:

- PostreSQL (and Microsoft SQL) can be used for creating databases, when researchers have access to the `SG <SRE ID> Data Administrators` privileges
- All users can create tables in the public schema

A situation arose on a recent project where a researcher wanted to use PostGIS for spatial analyses, but wasn't sure how to do so and it wasn't clear to me either

craddm:
Check if documentation makes it clear what users should and should not be able to do, and whether any such documentation is clearly visible

craddm:
Some notes:
1) As Ed noted above, everyone can create tables in the public schema
2) Intended behaviour is that data admins can create tables in the data schema, and research users can only view them. This works after #1708 
3) the postgis schema (and the topology schema) does not currently have permissions set for anyone to do anything with it. not clear if anything needs to change there, though, as users should still be able to use PostGIS features in their tables (e.g. can still create a _point column in a new table in the Public schema)
4) System Administrators are supposed to be given superuser roles but are not.

PR #1708, opened 2024-01-23
Fix PostgreSQL permissions and data schema, and relevant docs

craddm:

Moves the script that ensures users have appropriate roles for modifying certain database tables to the correct directory, and corrects the mustache file that adds this script to the server. Previously, the script was not present on the server. Thus users were not given the right permissions and no data schema was created.

Changes the database trigger to avoid using `pg_has_role`. For superusers, `pg_has_role` reports that they have every possible user role on the server, irrespective of what other roles they are actually assigned.

Adds a cron job that runs every 10 minutes, 1 minute after the LDAP users are updated, to trigger an update of user roles.

Updates docs for system managers regarding the intended use of the security groups on the DC and their impact on database access rights.

Relates to #1438 and #1392

### :microscope: Tests

Deployed a new PostgreSQL database, created and modified a table in the `data` schema using a user in the `Data Administrators` security group and confirmed that a `Research Users` user could only read, not modify the table.

Checked that user permissions are correctly updated on the PostgreSQL database. The permissions now update appropriately every 10 minutes, including removing a user from the Sys Admins group.


JimMadge (approved):
Should it be closes https://github.com/alan-turing-institute/data-safe-haven/issues/1438 and https://github.com/alan-turing-institute/data-safe-haven/issues/1392?

craddm:
That's something I wanted to discuss.

#1438 is partly covered by this, but this doesn't address point 2 of #1438 - without being in Research Users group, sys admins and data administrators can't get to the database to modify anything. If we are fine with that (and the docs now spell out that they need to be research users to be able to login to the SRD), then this closes #1438

#1392 is slightly more about telling users how to use the databases. I've just added links to tutorials for MSSQL and PostgreSQL that hopefully suffice.

Issue #1689, opened 2023-12-14
Move encoded stacks from config to another location


JimMadge:
Stacks, which hold configuration/settings for specific Pulumi stacks, are currently stored in the configuration as base64 encoded strings.

This means the configuration has to be written to (and changed) whenever Pulumi makes changes, even if as far as the user is concerned the configuration remains the same.
It also adds the possibility of corrupting the stack by changing the configuration after #1674. I can't imagine anyone wants to edit their infrastructure as a b64 string :smile:.

I also don't think this information would be considered configuration.

It would be better to move these to a different place. Certainly a different file/object, possibly a different container in the context.

---

At the same time, this would be the time to move the hard-coded strings for Pulumi backend storage account name. Discussed in #1711.

https://github.com/alan-turing-institute/data-safe-haven/blob/9e695cbbdadac7d07ebdf8142867b414c9423890/data_safe_haven/config/config.py#L58-L61

JimMadge:
This (probably :thinking:) also blocks adding protection against modifying the config. At least whenever a stack changes you would need to force push to the configuration.

JimMadge:
Also, on a closer look. These are _not_ the stacks.

The stacks are stored in the Pulumi backend at `.pulumi/stacks/<project>/*.{json,json.bak}`.
These are the pulumi configuration, equivalent of `Pulumi.<stack>.yaml`

PR #1711, opened 2024-01-24
Create pulumi container

### LINKED ISSUE #1710: Pulumi: storage container not created
`dsh context create` should create a `pulumi` storage container, but it actually creates the `config` container twice instead.

jemrobinson:

Ensure that `dsh context create` also creates the `pulumi` storage container

Closes #1710

### :microscope: Tests

Tested on fresh deployment

### DIFF
--- a/data_safe_haven/config/__init__.py
+++ b/data_safe_haven/config/__init__.py
@@ -1,8 +1,10 @@
-from .config import Config
+from .config import Config, ConfigSectionPulumi, ConfigSectionTags
 from .context_settings import Context, ContextSettings
 
 __all__ = [
     "Config",
+    "ConfigSectionPulumi",
+    "ConfigSectionTags",
     "Context",
     "ContextSettings",
 ]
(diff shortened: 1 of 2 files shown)


JimMadge (approved):
LGTM :+1:.

I had to look around the code a bit because I couldn't recall exactly what I had done.

I remember thinking it is a bit odd to import things from the config classes in "context" space because you should create the context before a config.

But these are hardcoded values, so if they belong most naturally in the config I think that is fine. At least, not a bug or priority to move.

jemrobinson:
Agreed that this should probably live in the `Context` class but I noticed you already had it as part of `ConfigSectionPulumi`. I'll leave it up to you whether to fix this now or merge this PR as-is.

JimMadge:
I'm not really sure. From a "structure of the program" perspective, it would neatly fit in the Context. I think I might have convinced myself it didn't conceptually fit there as the context doesn't really know about Pulumi. However, in the case of the name of the storage account that is part of the context, it totally does.

Added to an existing Issue #1689.

PR #1735, opened 2024-02-08
Upgrade to PostgreSQL flexible server

### LINKED ISSUE #1733: Pulumi: Upgrade to Azure Flexible PostgreSQL server

As stated [here](https://azure.microsoft.com/en-gb/updates/azure-database-for-postgresql-single-server-will-be-retired-migrate-to-flexible-server-by-28-march-2025/) Azure Database for PostgreSQL Single Server is being retired and must be replaced with Azure Database for PostgreSQL Flexible Server.

This only affects Pulumi as the Powershell code is not using database-as-a-service.

Replace usage of Azure Database for PostgreSQL Single Server with Azure Database for PostgreSQL Flexible Server.

jemrobinson:

- Move from Azure PostgreSQL server to Azure flexible PostgreSQL server
- Update to `guacamole-user-sync v0.2.0` which fixes a bug with special characters in passwords

Closes #1733 

### :microscope: Tests

Tested on a new SRE

### DIFF
--- a/data_safe_haven/README.md
+++ b/data_safe_haven/README.md
@@ -47,6 +47,12 @@ Run `dsh deploy shm -h` to see the necessary command line flags and provide them
 > dsh config upload config.yaml
 ```
 
+- As private endpoints for flexible PostgreSQL are still in preview, the following command is currently needed:
+
+```console
+> az feature register --name "enablePrivateEndpoint" --namespace "Microsoft.DBforPostgreSQL"
+```
+
 - Next deploy the infrastructure  [approx 30 minutes]:
 
 ```console
--- a/data_safe_haven/external/interface/azure_postgresql_database.py
+++ b/data_safe_haven/external/interface/azure_postgresql_database.py
@@ -2,16 +2,15 @@ import datetime
 import pathlib
 import time
 from collections.abc import Sequence
-from typing import Any
+from typing import Any, cast
 
 import psycopg
 import requests
 from azure.core.polling import LROPoller
-from azure.mgmt.rdbms.postgresql import PostgreSQLManagementClient
-from azure.mgmt.rdbms.postgresql.models import (
+from azure.mgmt.rdbms.postgresql_flexibleservers import PostgreSQLManagementClient
+from azure.mgmt.rdbms.postgresql_flexibleservers.models import (
     FirewallRule,
     Server,
-    ServerUpdateParameters,
 )
 
 from data_safe_haven.exceptions import (
@@ -72,7 +71,7 @@ class AzurePostgreSQLDatabase:
                 f"host={self.db_server.fully_qualified_domain_name}",
                 f"password={self.db_server_admin_password}",
                 f"port={self.port}",
-                f"user={self.db_server.administrator_login}@{self.server_name}",
+                f"user={self.db_server.administrator_login}",
                 "sslmode=require",
             ]
         )
@@ -100,15 +99,17 @@ class AzurePostgreSQLDatabase:
         """Get the database connection."""
         while True:
             try:
-                connection = psycopg.connect(self.connection_string)
-                break
-            except psycopg.OperationalError as exc:
-                if n_retries > 0:
+                try:
+                    connection = psycopg.connect(self.connection_string)
+                    break
+                except psycopg.OperationalError as exc:
+                    if n_retries <= 0:
+                        raise exc
                     n_retries -= 1
                     time.sleep(10)
-                else:
-                    msg = f"Could not connect to database.\n{exc}"
-                    raise DataSafeHavenAzureError(msg) from exc
+            except Exception as exc:
+                msg = f"Could not connect to database.\n{exc}"
+                raise DataSafeHavenAzureError(msg) from exc
         return connection
 
     def load_sql(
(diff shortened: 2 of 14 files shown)


data_safe_haven/README.md
```diff
@@ -47,6 +47,12 @@ Run `dsh deploy shm -h` to see the necessary command line flags and provide them
 > dsh config upload config.yaml
 ```
 
+- As private endpoints for flexible PostgreSQL are still in preview, the following command is currently needed:
+
+```console
+> az feature register --name "enablePrivateEndpoint" --namespace "Microsoft.DBforPostgreSQL"
+```
```
JimMadge, line 54: Unfortunate, but I don't think it is worth doing this another way if it isn't in Pulumi Azure Native yet.

data_safe_haven/external/interface/azure_postgresql_database.py
```diff
@@ -198,28 +192,36 @@ def set_database_access(self, action: str) -> None:
...
-                    ServerUpdateParameters(public_network_access="Disabled"),
-                )
+            rules = cast(
+                list[FirewallRule],
+                self.db_client.firewall_rules.list_by_server(
+                    self.resource_group_name, self.server_name
+                ),
             )
+            for rule in rules:
+                if rule.name:
```
JimMadge: Is it possible for a rule to have no name attribute?
If it doesn't do we really want to just skip it?

data_safe_haven/infrastructure/components/composite/postgresql_database.py
```diff
@@ -45,31 +45,36 @@ def __init__(
...
             server_name=props.database_server_name,
             sku=dbforpostgresql.SkuArgs(
-                capacity=2,
-                family="Gen5",
-                name="GP_Gen5_2",
-                tier=dbforpostgresql.SkuTier.GENERAL_PURPOSE,  # required to use private link
+                name="Standard_B2s",
+                tier=dbforpostgresql.SkuTier.BURSTABLE,
+                # name="Standard_D2s_v3",
+                # tier=dbforpostgresql.SkuTier.GENERAL_PURPOSE,  # required to use private link
```
JimMadge: Are we expecting to change to this size later?

data_safe_haven/infrastructure/components/composite/postgresql_database.py
```diff
@@ -45,31 +45,36 @@ def __init__(
...
+            location=props.location,
             resource_group_name=props.database_resource_group_name,
             server_name=props.database_server_name,
             sku=dbforpostgresql.SkuArgs(
-                capacity=2,
-                family="Gen5",
-                name="GP_Gen5_2",
-                tier=dbforpostgresql.SkuTier.GENERAL_PURPOSE,  # required to use private link
+                name="Standard_B2s",
+                tier=dbforpostgresql.SkuTier.BURSTABLE,
```
JimMadge, line 70: It would be good to make this configurable, along with the storage size.

Let's not block this though, could you open an issue?

data_safe_haven/README.md
jemrobinson: Yes, I think this will be temporary until either:

1. they roll this out more widely
2. direct integration into a VNet is fixed (currently not possible to do this except through the portal)

data_safe_haven/external/interface/azure_postgresql_database.py
jemrobinson: I think it *is* possible (but unlikely) for a rule to have no name. If this happens we can't delete it as `begin_delete` takes a rule name as input.

An equivalent workaround would be to wrap the `begin_delete` in a `try/except` - would you prefer this?

data_safe_haven/infrastructure/components/composite/postgresql_database.py
jemrobinson: With non-flexible servers we *do* have to use the GP SKU but this seems not to be true for flexible. This is erroneous, dead code that I will remove.

data_safe_haven/infrastructure/components/composite/postgresql_database.py
jemrobinson: Good idea.

data_safe_haven/infrastructure/components/composite/postgresql_database.py
jemrobinson: Done

data_safe_haven/infrastructure/components/composite/postgresql_database.py
jemrobinson: Captured in #1736

data_safe_haven/external/interface/azure_postgresql_database.py
JimMadge: I think the main thing is to not silently pass over failing to remove a rule.
Might be better to give a warning rather than raise an exception?

I have made some changes around what I was thinking in 0704ba2. What do you think?

jemrobinson: Looks good. I've added an explicit type-cast to `str` as my type-checker insists that `rule.name` is always `None` since it's initialised as `None` and no functions explicitly change it (`az` libraries seem to do a lot of direct editing of the `__dict__` of objects).

JimMadge (approved):

PR #1739, opened 2024-02-12
Add script to renew NFS share Stored Access Policies

### LINKED ISSUE #895: Rotation of SAS tokens
#### :scroll: Description
We currently use SAS tokens in a few places, most notably for accessing the ingress and egress storage accounts from inside the SRE. These have a default lifetime of 1 year. We need to work out a process for retiring/rotating these tokens that is as seamless as possible.

#### :strawberry: Desired behaviour
A procedure for rotating SAS tokens has been implemented.

JimMadge:

Add script to update SAS Access Policies for storage mounted to SRDs.
Creating new SAS tokens is not needed as they are bound to the permissions, start, and end dates of the Access Policy.

Closes #895 

### :microscope: Tests

Tested on SRE,
  - Policies confirmed to be updated in the portal
  - Access to mounted containers still works
  - Restarted VM to ensure that the files in containers were not cached

Thanks to @craddm for helping me test and spotting that the SAS tokens are bound to a Stored Access Policy :pray:.


deployment/secure_research_environment/remote/secure_research_desktop/scripts/write_sas_tokens.sh
```diff
@@ -0,0 +1,5 @@
+#! /usr/bin/env bash
+
+sed -i 's/^sasToken/sasToken "${BACKUP}"/' /opt/configuration/credentials-backup.secret
```
jemrobinson: Base64 is allowed to use "/" so I would suggest using "|" as your sed field marker.

deployment/secure_research_environment/setup/Update_SRE_SAS_Tokens.ps1
```diff
@@ -0,0 +1,50 @@
...
+    # Create token
+    $sasToken = New-StorageReceptacleSasToken -ContainerName $receptacleName -PolicyName $sasPolicy.Policy -StorageAccount $persistentStorageAccount
+    # Write to KeyVault
+    $sasTokens[$receptacleName] = Resolve-KeyVaultSecret -VaultName $config.sre.keyVault.name -SecretName $config.sre.storage.persistentdata.containers[$receptacleName].connectionSecretName -DefaultValue $sasToken -AsPlaintext -ForceOverwrite
+}
+
+# Get list of SRDs
+# ----------------
+Add-LogMessage -Level Info "Retrieving list of SRD VMs..."
+$VMs = Get-AzVM -ResourceGroupName $config.sre.srd.rg | `
```
jemrobinson: I feel like we do this selection in a few places. Is there no common function for this?

deployment/secure_research_environment/setup/Update_SRE_SAS_Tokens.ps1
```diff
@@ -0,0 +1,50 @@
...
+    Where-Object { $_.Name -like "*SRD*" }
+
+# Update blobfuse credentials on each SRD
+# ---------------------------------------
+$scriptPath = Join-Path $PSScriptRoot ".." "remote" "secure_research_desktop" "scripts" "write_sas_tokens.sh"
+for $VM in $Vms {
+    $null = Invoke-RemoteScript -VMName $VM.Name -ResourceGroupName $VM.ResourceGroupName -Shell "UnixShell" -ScriptPath $scriptPath -Parameter $sasTokens
+}
```
jemrobinson: I think you might need to reboot the VMs here to force the mount to disconnect and reconnect with the new SAS token. Alternatively, you could force a remount in the `write_sas_tokens.sh` script.

docs/source/roles/system_manager/manage_deployments.md
```diff
@@ -26,6 +26,16 @@ PS> ./Update_SRE_SSL_Certificate.ps1 -shmId <SHM ID> -sreId <SRE ID>
...
 - where `<SRE ID>` is the {ref}`secure research environment ID <roles_deployer_sre_id>` for this SRE
 
+(renew_sas)=
+
+## {{locked_with_key}} Renew SRE SAS tokens
+
+The [SRE storage containers](role_researcher_user_guide_shared_storage) for input data, backup and output are all provided by blob storage.
+The SRDs use [SAS tokens](https://learn.microsoft.com/en-us/azure/storage/common/storage-sas-overview) to authenticate and access the data.
+
+When the containers are deployed the SAS tokens are valid for one year.
```
jemrobinson: Is 1 year the right time? We had this as a generic "seems like long enough for most projects" placeholder. It could be that we want to cycle SAS tokens more regularly than that. Maybe not something to change in this PR though.

deployment/secure_research_environment/remote/secure_research_desktop/scripts/write_sas_tokens.sh
JimMadge: Changed in https://github.com/alan-turing-institute/data-safe-haven/pull/1739/commits/b2fbc44dd1aaa426e070d11e77a6b38230bab487

deployment/secure_research_environment/setup/Update_SRE_SAS_Tokens.ps1
JimMadge: A few places but I think it is often with different filters so it might not be worth trying to put into a function.

deployment/secure_research_environment/setup/Update_SRE_SAS_Tokens.ps1
JimMadge: Prompting the systemd mount units to restart might do this.
As this will be fairly rare and will interrupt access to the data anyway, a restart seems sensible.

JimMadge: Went with the systemd route in 005dfe1.

jemrobinson:
Looks fine to me if someone can test it.

JimMadge:
IIRC we found a problem with the scripts "arguments" containing disallowed characters and the exception suggested base64 encoding, so we will need to look at that.

JimMadge:
Base64 issue fixed :tada:.

The script runs and SAS tokens appear to be updated correctly. The script seems to stall then, presumably the `systemd` commands.

JimMadge:
Still stalling on the script. Everything seems fine, perhaps the script isn't returning for some reason? Do we need to add `exit 0` or something?

deployment/secure_research_environment/setup/Update_SRE_SAS_Tokens.ps1
```diff
@@ -0,0 +1,65 @@
...
+# Update blobfuse credentials on each SRD
+# ---------------------------------------
+$sasTokensBase64 = @{}
+$sasTokens.Keys | ForEach-Object {
+    $sasTokensBase64[($_ + '_B64')] = ($sasTokens[$_] | ConvertTo-Base64)
+}
+$scriptPath = Join-Path $PSScriptRoot ".." "remote" "secure_research_desktop" "scripts" "write_sas_tokens.sh"
+foreach ($VM in $Vms) {
+    Add-LogMessage -Level Info "Updating SAS tokens on '$($VM)'"
+    $null = Invoke-RemoteScript -VMName $VM.Name -ResourceGroupName $VM.ResourceGroupName -Shell "UnixShell" -ScriptPath $scriptPath -Parameter $sasTokensBase64
```
jemrobinson: I think this should be something like: `-Parameter @{"BACKUP_B64" = "$BACKUP_B64"; "EGRESS_B64" = "$EGRESS_B64"; "INGRESS_B64" = "$INGRESS_B64" }`

JimMadge: Is that not equivalent to what this is already doing?

If the arguments are not all there, could it cause the script to hang?

jemrobinson: Maybe try with a dummy remote script that just prints out ENV so you can see what inputs you have?

craddm: A dummy script that just prints out ENV also hangs. Wondering if it's just that the SRD can't reply back.

jemrobinson: Once an `Invoke-RemoteScript` call fails, the VM is locked up for 90+ minutes for future calls.

JimMadge: That may be the problem that :smile:. It was definitely wrong before.
Is there a way to force the script or "unlock" the VM?

jemrobinson: I'm not aware of one. Easiest way is probably to redeploy the VM.

deployment/common/AzureStorage.psm1
```diff
@@ -55,7 +55,9 @@ function Deploy-SasAccessPolicy {
         [Parameter(Mandatory = $false, ParameterSetName = "ByShareName", HelpMessage = "Container name")]
         [string]$ShareName,
         [Parameter(Mandatory = $false, HelpMessage = "Validity in years")]
-        [int]$ValidityYears = 20
+        [int]$ValidityYears = 20,
```
jemrobinson, line 58: A nice use of a "not-my-problem" default value :)

deployment/secure_research_environment/setup/Update_Stored_Access_Policies.ps1
```diff
@@ -0,0 +1,38 @@
...
+Import-Module $PSScriptRoot/../../common/Logging -Force -ErrorAction Stop
+
+
+# Get config and original context before changing subscription
+# ------------------------------------------------------------
+$config = Get-SreConfig -shmId $shmId -sreId $sreId
+$originalContext = Get-AzContext
+$null = Set-AzContext -SubscriptionId $config.sre.subscriptionName -ErrorAction Stop
+
+# Update each SAS Access Policy to be valid for one year from now
```
jemrobinson, line 23: Should the default here (1 more year) be different from the original default (20 years)?

docs/source/roles/system_manager/manage_deployments.md
```diff
@@ -26,6 +26,22 @@ PS> ./Update_SRE_SSL_Certificate.ps1 -shmId <SHM ID> -sreId <SRE ID>
...
 - where `<SRE ID>` is the {ref}`secure research environment ID <roles_deployer_sre_id>` for this SRE
 
+(renew_sas)=
+
+## {{locked_with_key}} Renew SRE Container Access Policies
+
+The [SRE storage containers](role_researcher_user_guide_shared_storage) for input data, backup and output are all provided by blob storage.
+The SRDs use [SAS tokens](https://learn.microsoft.com/en-us/azure/storage/common/storage-sas-overview) bound to a [Stored Access Policy](https://learn.microsoft.com/en-us/azure/storage/common/storage-stored-access-policy-define-dotnet) to authenticate and access the data.
+
+When the containers are deployed the Stored Access Policy is valid for one year.
```
jemrobinson, line 36: Not true as currently written.

jemrobinson (commented):
How does this deal with the possibility of SAS tokens being leaked (and therefore needing to be replaced)? Or are you arguing that we shouldn't be worried about that? If so, what's the reason for refreshing the policies every year?

JimMadge:
> How does this deal with the possibility of SAS tokens being leaked (and therefore needing to be replaced)? Or are you arguing that we shouldn't be worried about that? If so, what's the reason for refreshing the policies every year?

In that case (which I believe would required privilege escalation on an SRD) you would still only be able to use that token from either inside the SRE (through the private end point) or from one of the approved IP addresses (which we only use for the system managers, who have privileged access already).

So, I'm quite confident that is a low risk. Although, if you did know that it had happened, I think it would be sensible to revoke the old policy and create a new one.

I think here I'm trying to balance fixing the immediate issue in a timely way with making the ideal change. Keeping in mind this is soon to be deprecated.

docs/source/roles/system_manager/manage_deployments.md
JimMadge: The "valid for one year" part?

I believe it is because the default value is overwritten with a hard-coded value of 1

https://github.com/alan-turing-institute/data-safe-haven/blob/e8d67c7fae8cd23e265132f10792d5cd826a26f9/deployment/secure_research_environment/setup/Setup_SRE_Storage_Accounts.ps1#L64-L68

jemrobinson:
I guess my point is that if we're not worried about leakage of SAS tokens, then why not just set the default validity to the maximum value? Then we never have to worry about refreshing the tokens or policies.

JimMadge:
> I guess my point is that if we're not worried about leakage of SAS tokens, then why not just set the default validity to the maximum value?

Yes, I'm not trying to dodge that and it is a good question.

Here the maximum seems to be "valid forever". I think that is generally considered bad practice. Probably because of the risk of that forgetting about that kind of token and not revoking it when it is no longer needed.

In our case, that doesn't feel like a big risk as it can only be used in a very restricted number of places and (maybe) we expect the storage to be torn down when it is no longer needed. Might not be true for other users.

My thinking was, that is another issue.
Although, it might be possible to change those Stored Access Policies to be valid forever and do away with the renewal altogether.

craddm (approved):
LGTM. Tested on the current SRE and seems to be working correctly.

JimMadge:
@jemrobinson do you want to block/close this PR, or open an issue for a better solution in the future?

My feeling is, a discussion about whether open-ended access policies are really such a bad idea here is worth having. However, I don't think that will be done in time for the next release, so this PR is a sufficient for now fix.

jemrobinson (approved):
Approving as I don't want to block this PR. I'm not 100% sure what problem it's solving though as I think the solution here is equivalent to just setting a really long default validity. If you could open a new issue under the 4.2.1 milestone, I think that would be fine.

Issue #1738, opened 2024-02-12
Outdated software versions on current version of DSH


craddm:
## :white_check_mark: Checklist

- [x] I have searched open and closed issues for duplicates.
- [x] This is a request for a new feature in the Data Safe Haven or an upgrade to an existing feature.
- [x] The feature is still missing in the [latest version](https://github.com/alan-turing-institute/data-safe-haven/releases).
- [x] I have read through the [documentation](https://alan-turing-institute.github.io/data-safe-haven/).
- [x] This isn't an open-ended question (open a [discussion](https://github.com/alan-turing-institute/data-safe-haven/discussions) if it is).

## :strawberry: Suggested change

Some of the software applications we use are several versions out of date. 

For example, the current version of Guacamole in use by our SREs is `1.3.0`, which is now several years and versions out of date. This should be updated to the latest version, `1.5.4`.

Other software may also need to be updated. e.g. the version of `nginx` used on the Guacamole server is also several versions out of date.

A list of relevant features/servers (to be updated when found):

- [x] Guacamole remote desktop
- [x] Codi MD
- [x] Proxy/update servers

## :steam_locomotive: How could this be done?

Review the deployment scripts to look for places where software such as Guacamole and nginx is pinned to a specific version, and check if an update is warranted.

PR #1741, opened 2024-02-13
Update software on guacamole server

craddm:

Updates the versions of Guacamole, nginx, and Postgres used on the Guacamole remote desktop server. 

In addition, sometimes the Postgres database did not finish initializing before Docker launched Guacamole, and thus Guacamole sometimes did not function. This PR adds a health check to the `docker-compose.yaml` to make Docker wait until the Postgres database is fully responsive. This necessitated migrating from Docker Compose v1 - `docker-compose` - to Docker Compose v2 - `docker compose`. Compose v1 is no longer supported, and the necessary syntax to perform the health check was not supported.

Relates to #1738 but does not close it, as this PR only focuses on the Guacamole server. 

### :microscope: Tests

Deployed a Guacamole server within an SRE and checked that it functioned correctly.

### DIFF
--- a/deployment/secure_research_environment/cloud_init/cloud-init-guacamole.mustache.yaml
+++ b/deployment/secure_research_environment/cloud_init/cloud-init-guacamole.mustache.yaml
@@ -110,7 +110,7 @@ packages:
   - clamav-freshclam
   - clamav-unofficial-sigs
   - docker.io
-  - docker-compose
+  - docker-compose-v2
   - gcc
   - libpq-dev
   - make
@@ -170,7 +170,7 @@ runcmd:
   - sleep 1m
   - systemctl status docker
   - docker --version
-  - docker-compose --version
+  - docker compose --version
 
   # Set up the guacamoledaemon user
   - echo ">=== Configuring guacamoledaemon user... ===<"
@@ -197,15 +197,15 @@ runcmd:
 
   # Deploy Guacamole using Docker
   - echo ">=== Deploying Guacamole with Docker...  ===<"
-  - su guacamoledaemon -c "docker-compose -f /opt/guacamole/docker-compose.yaml up -d"
+  - su guacamoledaemon -c "docker compose -f /opt/guacamole/docker-compose.yaml up -d"
 
   # Generate the necessary SQL config for the local PostgreSQL database and run it
   - echo ">=== Configuring local PostgreSQL database... ===<"
-  - su guacamoledaemon -c "docker-compose -f /opt/guacamole/docker-compose.yaml exec -T guacamole /opt/guacamole/bin/initdb.sh --postgres" > /opt/postgresql/scripts/db_init.sql
+  - su guacamoledaemon -c "docker compose -f /opt/guacamole/docker-compose.yaml exec -T guacamole /opt/guacamole/bin/initdb.sh --postgresql" > /opt/postgresql/scripts/db_init.sql
   - echo "CREATE ROLE ldap_users;" >> /opt/postgresql/scripts/db_init.sql
   - echo "CREATE ROLE ldap_groups;" >> /opt/postgresql/scripts/db_init.sql
   - ls -alh /opt/postgresql/scripts/db_init.sql
-  - su guacamoledaemon -c "docker-compose -f /opt/guacamole/docker-compose.yaml exec -T postgres psql -U guacamole -f /scripts/db_init.sql"
+  - su guacamoledaemon -c "docker compose -f /opt/guacamole/docker-compose.yaml exec -T postgres psql -U guacamole -f /scripts/db_init.sql"
 
   # Add initial set of LDAP users with pg-ldap-sync
   - echo ">=== Adding LDAP users with pg-ldap-sync... ===<"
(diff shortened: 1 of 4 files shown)


deployment/secure_research_environment/cloud_init/cloud-init-guacamole.mustache.yaml
```diff
@@ -170,7 +170,7 @@ runcmd:
   - sleep 1m
   - systemctl status docker
   - docker --version
-  - docker-compose --version
+  - docker compose --version
```
jemrobinson, line 173: Isn't `docker compose` part of `docker.io` now? Do we need `docker-compose-v2` as well?

deployment/secure_research_environment/cloud_init/resources/guacamole_docker_compose.mustache.yaml
```diff
@@ -26,11 +26,16 @@ services:
             POSTGRES_DB: guacamole
             POSTGRES_USER: guacamole
             POSTGRES_PASSWORD: {{{guacamole.internalDbPassword}}}
+        healthcheck:
```
jemrobinson, line 29: Does `healthcheck` actually do anything? I know it's supposed to, but does it work?

deployment/secure_research_environment/cloud_init/cloud-init-guacamole.mustache.yaml
craddm: No, `docker compose` didn't work until `docker-compose-v2` was installed. I would also note that `docker.io` is a bit behind the official Docker release, but it'd require adding the official docker repo as a source to use that. I can see that at some point we switched from the latter to `docker.io`, so avoided doing that for consistency.

deployment/secure_research_environment/cloud_init/resources/guacamole_docker_compose.mustache.yaml
craddm: Deploying guacamole failed twice after I'd updated to the newer version, seemingly because the Postgres database had not finished initializing when docker started Guacamole. After adding the healthcheck it's worked fine. Not sure if there's a way to find logs of the healthcheck.

jemrobinson: That sounds like convincing evidence to me :)

JimMadge: `HEALTHCHECK` only adds a "health" status for containers. It doesn't restart them if they fail, you need another service to do that.

([I wrote one](https://github.com/JimMadge/healthy) for my homeserver :smile:)

craddm: Which is fine - all I want it to do is make guacamole wait until postgres is finished initialising and healthy, not restart anything.

deployment/secure_research_environment/cloud_init/cloud-init-guacamole.mustache.yaml
JimMadge: I think I'm still missing `docker compose` on my Ubuntu systems.

deployment/secure_research_environment/cloud_init/resources/guacamole_docker_compose.mustache.yaml
JimMadge: Oh, I think it also won't do that.
Starting containers in the right order can be a bit tricky.

However, most containers that depend on databases wait until they can see the database before doing anything. I would expect Guacamole should do that.

craddm: that's why I did this, it wasn't waiting.

JimMadge: I would be surprised if this fixes that problem. I think it is a race condition.

If you look at the cloud-init `docker-compose up -d` is run before the database is initialised.
It might be better to add a restart after doing that, or not bringing the Guacamole/Guacd container up until after initialising the database.

craddm: I think I have indeed slightly misdiagnosed the problem, but think the healthcheck may actually fix it.

Here's part of the cloud-init log from a failed deployment, without the healthcheck implemented.

image

The line of code that fails is 

`su guacamoledaemon -c "docker compose -f /opt/guacamole/docker-compose.yaml exec -T postgres psql -U guacamole -f /scripts/db_init.sql"`

The previous errors I've seen have been slightly different (always at the same point), but the general issue is the same - the database isn't ready to run the `db_init.sql` script.

So it's not that guacamole isn't up - it's a race, as you say. 

However, I think the healthcheck prevents this from happening, because it forces guacamole to wait until the db is healthy and accepting connections before starting. So by extension, by the time we get to the line that runs `db_init.sql`, the db is ready to run it, and we don't hit this problem.

Here's part of the cloud-init log from a successful deployment, with the healthcheck step.

image

Note also the init script is created by guacamole, so we can't init the db until guacamole is up:

`- su guacamoledaemon -c "docker compose -f /opt/guacamole/docker-compose.yaml exec -T guacamole /opt/guacamole/bin/initdb.sh --postgresql" > /opt/postgresql/scripts/db_init.sql`

This process is different from the initialization done when starting the database up in the first place.

JimMadge (approved):

PR #1744, opened 2024-02-16
Update Nexus proxy server for T2/T3 package access

craddm:

Updates the Nexus proxy server used for serving R and Python packages for T2 and T3 SREs to the latest version. Also updates the nginx proxy it sits behind.

Relates to #1738 

### :microscope: Tests

Updated server and tested functionality via T2 SRE. Not tested with a T3 SRE.

### DIFF
--- a/deployment/safe_haven_management_environment/cloud_init/resources/nexus_docker_compose.mustache.yaml
+++ b/deployment/safe_haven_management_environment/cloud_init/resources/nexus_docker_compose.mustache.yaml
@@ -13,7 +13,7 @@ services:
     restart: always
   nginx:
     container_name: nginx_nexus_compose
-    image: nginx:1.21.6
+    image: nginx:1.25.3
     ports:
       - 80:80
     volumes:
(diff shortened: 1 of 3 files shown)


deployment/safe_haven_management_environment/cloud_init/resources/nexus_docker_compose.mustache.yaml
```diff
@@ -13,7 +13,7 @@ services:
     restart: always
   nginx:
     container_name: nginx_nexus_compose
-    image: nginx:1.21.6
+    image: nginx:1.25.3
```
JimMadge:
```suggestion
    image: nginx:1.25
```

Might be better to avoid patch level here so that we get bug fixes.

JimMadge (changes requested):

deployment/safe_haven_management_environment/cloud_init/resources/nexus_docker_compose.mustache.yaml
craddm: Ok - same could apply to the other places we use nginx, so will go back and change that too

JimMadge: Yes, I think a good rule of thumb is pin on minor versions. That way you should avoid breaking changes, but get bug fixes.

Not all contains create a tag `x.y` which points to the latest minor release of `x` though.

JimMadge (approved):

PR #1775, opened 2024-04-08
Fix Docker versions script

jemrobinson:

- Update Docker versions script with correct path and make it run on develop

n/a

### :microscope: Tests

Tested in CI

### DIFF
--- a/.github/scripts/update_docker_versions.py
+++ b/.github/scripts/update_docker_versions.py
@@ -67,7 +67,7 @@ def annotate(
     return annotated
 
 
-for filename in (pathlib.Path("data_safe_haven") / "pulumi").glob("**/*.py"):
+for filename in (pathlib.Path("data_safe_haven") / "infrastructure").glob("**/*.py"):
     needs_replacement = False
     lines = []
     with open(filename) as f_pulumi:
@@ -87,6 +87,8 @@ for filename in (pathlib.Path("data_safe_haven") / "pulumi").glob("**/*.py"):
                     print(f"Updating {image} from {v_current} to {v_latest} in {filename}")  # noqa: T201
                     needs_replacement = True
                     output = line.replace(v_current, v_latest)
+                else:
+                    print(f"Leaving {image} at {v_current} (latest version) in {filename}")  # noqa: T201
             lines += output
 
     if needs_replacement:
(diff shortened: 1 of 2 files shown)


.github/workflows/update_docker_versions.yaml
```diff
@@ -4,7 +4,7 @@ name: Update Pulumi Docker image versions
 # Run workflow on pushes to matching branches
 on:  # yamllint disable-line rule:truthy
   push:
-    branches: [python-migration]
+    branches: [develop]
   schedule:
     - cron: "0 0 */7 * *"  # run once per week
```
JimMadge:
```suggestion
    - cron: "0 3 * * 1"  # run at 3:00 every Monday
```

I think what is currently there will run when `day%7==0` instead of every seven days.

.github/workflows/update_docker_versions.yaml
```diff
@@ -4,7 +4,7 @@ name: Update Pulumi Docker image versions
 # Run workflow on pushes to matching branches
 on:  # yamllint disable-line rule:truthy
   push:
-    branches: [python-migration]
+    branches: [develop]
```
JimMadge: Do we need this to run on pushes if it is running every week?

.github/scripts/update_docker_versions.py
```diff
@@ -67,7 +67,7 @@ def annotate(
     return annotated
 
 
-for filename in (pathlib.Path("data_safe_haven") / "pulumi").glob("**/*.py"):
+for filename in (pathlib.Path("data_safe_haven") / "infrastructure").glob("**/*.py"):
```
JimMadge, line 70: In the future it would be good to try and collect all the container images and versions together somewhere more visible. Lets not solve that here though.

JimMadge (changes requested):

.github/workflows/update_docker_versions.yaml
jemrobinson: Probably not? Can you add a suggestion here that removes this?

.github/workflows/update_docker_versions.yaml
jemrobinson: Even worse - I think it's "At 00:00 on every 7th day-of-month."!

JimMadge (approved):

Issue #1570, opened 2023-08-24
Pulumi: replace domain controller with AzureAD proxy


jemrobinson:
## :white_check_mark: Checklist

- [x] I have searched open and closed issues for duplicates.
- [x] This is a request for a new feature in the Data Safe Haven or an upgrade to an existing feature.
- [x] The feature is still missing in the [latest version](https://github.com/alan-turing-institute/data-safe-haven/releases).
- [x] I have read through the [documentation](https://alan-turing-institute.github.io/data-safe-haven/).
- [x] This isn't an open-ended question (open a [discussion](https://github.com/alan-turing-institute/data-safe-haven/discussions) if it is).

## :strawberry: Suggested change

We should write a proxy server that can accept LDAP requests and query AzureAD for the answers. Ideally this should be available as a Docker image.

## :steam_locomotive: How could this be done?

~The Python [ldaptor](https://ldaptor.readthedocs.io) package might be a good place to start.~
I've got a proof-of-concept working [here](https://github.com/alan-turing-institute/apricot) which I've called Apricot.

- [x] Add required AzureAD/GraphAPI functionality to the code (see #1778)
- [x] Update SRE components to use Apricot for authentication/identity (see #1772)
- [x] Remove the SHM domain controller (see #1805)

PR #1772, opened 2024-04-08
Use Apricot for authentication/identity

jemrobinson:
## :white_check_mark: Checklist

- [x] You have given your pull request a meaningful title (_e.g._ `Enable foobar integration` rather than `515 foobar`).
- [x] You are targeting the appropriate branch. If you're not certain which one this is, it should be **`develop`**.
- [x] Your branch is up-to-date with the **target branch** (it probably was when you started, but it may have changed since then).
- [x] You have marked this pull request as a **draft** and added `'[WIP]'` to the title if needed (if you're not yet ready to merge).
- [x] You have formatted your code using appropriate automated tools (for example `./tests/AutoFormat_Powershell.ps1 -TargetPath <path to file or directory>` for Powershell).

### :arrow_heading_up: Summary

- [x] Add an Apricot server to the SRE
- [x] Update SRE components to use Apricot for authentication/identity

### :closed_umbrella: Related issues

Depends on #1778
Second part of #1570
Closes #1645

### :microscope: Tests

Tested on a fresh SRE deployment

PR #1778, opened 2024-04-10
Add AzureAD/EntraID functionality

jemrobinson:

- Add necessary AzureAD/EntraID functionality for managing users and groups directly in the cloud
- Add `remove`, `register` and `unregister` methods to AzureADUsers
- Fix default GraphAPI scopes for SRE deployment
- Add function to create a service principal for AzureAD applications
- Update `create_group` function to no longer require `gidNumber`
- Drop support for external `linux_schema` in AzureAD
- Add functions to grant approval for application/delegated permissions
- Add `remove_user` function to GraphAPI
- Only `remove_user_from_group` if the user belongs to the group
- Grant requested permissions to applications during refresh, in case these weren't done at creation time

First step towards #1570

### :microscope: Tests

Tested on a new deployment

### DIFF
--- a/data_safe_haven/administration/users/azure_ad_users.py
+++ b/data_safe_haven/administration/users/azure_ad_users.py
@@ -85,11 +85,10 @@ class AzureADUsers:
 
     def remove(self, users: Sequence[ResearchUser]) -> None:
         """Remove list of users from AzureAD"""
-        for user_to_remove in users:
-            matched_users = [user for user in self.list() if user == user_to_remove]
-            if not matched_users:
-                continue
-            user = matched_users[0]
+        for user in filter(
+            lambda existing_user: any(existing_user == user for user in users),
+            self.list(),
+        ):
             try:
                 self.graph_api.remove_user(user.username)
                 self.logger.info(f"Removed '{user.preferred_username}'.")
--- a/data_safe_haven/administration/users/azure_ad_users.py
+++ b/data_safe_haven/administration/users/azure_ad_users.py
@@ -46,16 +46,8 @@ class AzureADUsers:
                     request_json, user.email_address, user.phone_number
                 )
             self.logger.info(
-                f"Ensured user '{user.preferred_username}' exists in AzureAD"
+                f"Ensured user '[green]{user.preferred_username}[/]' exists in AzureAD"
             )
-        # Decorate all users with the Linux schema
-        self.set_user_attributes()
-        # # Ensure that all users belong to an associated group the same name and UID
-        # for user in self.list():
-        #     self.graph_api.create_group(user.username, user.uid_number)
-        #     self.graph_api.add_user_to_group(user.username, user.username)
-        #     # Also add the user to the research users group
-        #     self.graph_api.add_user_to_group(user.username, self.researchers_group_name)
 
     def list(self) -> Sequence[ResearchUser]:
         user_list = self.graph_api.read_users()
@@ -113,55 +105,3 @@ class AzureADUsers:
         users_to_add = [user for user in users if user not in self.list()]
         self.add(users_to_add)
 
-    def set_user_attributes(self) -> None:
-        """Ensure that all users have Linux attributes"""
-        # next_uid = max(
-        #     [int(user.uid_number) + 1 if user.uid_number else 0 for user in self.users]
-        #     + [10000]
-        # )
-        # for user in self.users:
-        #     try:
-        #         # Get username from userPrincipalName
-        #         username = user.user_principal_name.split("@")[0]
-        #         if not user.homedir:
-        #             user.homedir = f"/home/{username}"
-        #             self.logger.debug(
-        #                 f"Added homedir {user.homedir} to user {user.preferred_username}"
-        #             )
-        #         if not user.shell:
-        #             user.shell = "/bin/bash"
-        #             self.logger.debug(
-        #                 f"Added shell {user.shell} to user {user.preferred_username}"
-        #             )
-        #         if not user.uid_number:
-        #             # Set UID to the next unused one
-        #             user.uid_number = next_uid
-        #             next_uid += 1
-        #             self.logger.debug(
-        #                 f"Added uid {user.uid_number} to user {user.preferred_username}"
-        #             )
-        #         if not user.username:
-        #             user.username = username
-        #             self.logger.debug(
-        #                 f"Added username {user.username} to user {user.preferred_username}"
-        #             )
-        #         # Ensure that the remote user matches the local model
-        #         patch_json = {
-        #             GraphApi.linux_schema: {
-        #                 "gidnumber": user.uid_number,
-        #                 "homedir": user.homedir,
-        #                 "shell": user.shell,
-        #                 "uid": user.uid_number,
-        #                 "user": user.username,
-        #             }
-        #         }
-        #         self.graph_api.http_patch(
-        #             f"{self.graph_api.base_endpoint}/users/{user.azure_oid}",
-        #             json=patch_json,
-        #         )
-        #         self.logger.debug(f"Set Linux attributes for user {user.preferred_username}.")
-        #     except Exception as exc:
-        #         self.logger.error(
-        #             f"Failed to set Linux attributes for user {user.preferred_username}.\n{str(exc)}"
-        #         )
-        pass
--- a/data_safe_haven/administration/users/azure_ad_users.py
+++ b/data_safe_haven/administration/users/azure_ad_users.py
@@ -3,6 +3,7 @@
 from collections.abc import Sequence
 from typing import Any
 
+from data_safe_haven.exceptions import DataSafeHavenMicrosoftGraphError
 from data_safe_haven.external import GraphApi
 from data_safe_haven.functions import password
 from data_safe_haven.utility import LoggingSingleton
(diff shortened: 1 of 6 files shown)


data_safe_haven/administration/users/azure_ad_users.py
```diff
@@ -84,84 +77,34 @@ def list(self) -> Sequence[ResearchUser]:
...
-        #         self.logger.error(
-        #             f"Unable to remove '{user.preferred_username}' from group '{self.researchers_group_name}'"
-        #         )
-        pass
+        """Remove list of users from AzureAD"""
+        for user_to_remove in users:
+            matched_users = [user for user in self.list() if user == user_to_remove]
+            if not matched_users:
+                continue
+            user = matched_users[0]
```
JimMadge: Feels like there must be a better way to do this like,

```python
matched_users = [user for user in self.list() if user in users]
for user in matched_users:
    ...
```

martintoreilly:
👀  Does this remove the "local" SHM AD servers and the SCE NPS servers?

jemrobinson:
> 👀 Does this remove the "local" SHM AD servers and the SCE NPS servers?

- This PR just adds some necessary functionality to our GraphAPI/AzureAPI wrappers.
- #1772 adds a new SRE identity server that gets its users directly from AzureAD/EntraID
- TBD next PR will remove the SHM AD servers

We've already dropped the NPS servers in 4.2.0 as part of dropping support for Microsoft Remote Desktop

data_safe_haven/external/api/graph_api.py
```diff
@@ -225,43 +230,32 @@ def create_application(
...
+                    self.grant_application_role_permissions(application_name, scope)
+                for scope in delegated_scopes:
+                    self.grant_delegated_role_permissions(application_name, scope)
+                while True:
+                    if application_sp := self.get_service_principal_by_name(
+                        application_name
+                    ):
+                        if self.read_application_permissions(application_sp["id"]):
+                            break
+                    time.sleep(10)
```
JimMadge:
```suggestion
                attempts = 0
                max_attempts = 5
                while attempts < max_attempts:
                    if application_sp := self.get_service_principal_by_name(
                        application_name
                    ):
                        if self.read_application_permissions(application_sp["id"]):
                            break
                    time.sleep(10)
                    attempts += 1
                
                if attempts == max_attempts:
                    raise DataSafeHavenException("Maximum attempts to validate service principle permissions exceeded")
```

Infinite loops feel like a bad idea.
Could we do this an raise an appropriate exception?

data_safe_haven/external/api/graph_api.py
```diff
@@ -573,12 +593,145 @@ def get_id_from_username(self, username: str) -> str | None:
...
             return None
 
+    def grant_role_permissions(
+        self,
+        application_name: str,
+        *,
+        application_role_assignments: Sequence[str],
+        delegated_role_assignments: Sequence[str],
+    ) -> None:
+        """Grant admin permission for requested application roles"""
```
JimMadge: I can't tell from the description and code what this function is doing.

Is it granting roles to an applications service principal?
Is it giving roles permissions?

data_safe_haven/external/api/graph_api.py
```diff
@@ -573,12 +593,145 @@ def get_id_from_username(self, username: str) -> str | None:
...
+
+        # Grant any requested delegated role permissions
+        for role_name in delegated_role_assignments:
+            self.grant_delegated_role_permissions(application_name, role_name)
+
+    def grant_application_role_permissions(
+        self, application_name: str, application_role_name: str
+    ) -> None:
+        """
+        Grant permissions for a particular role to an application.
```
JimMadge: Is it granting "permissions for a role" or granting "a role"? Is there a difference?

data_safe_haven/external/api/graph_api.py
```diff
@@ -573,12 +593,145 @@ def get_id_from_username(self, username: str) -> str | None:
...
+
+        Raises:
+            DataSafeHavenMicrosoftGraphError if the secret could not be created or already exists
+        """
+        try:
+            # Get service principals for Microsoft Graph and this application
+            microsoft_graph_sp = self.get_service_principal_by_name("Microsoft Graph")
+            if not microsoft_graph_sp:
+                msg = "Could not find Microsoft Graph service principal."
+                raise DataSafeHavenMicrosoftGraphError(msg)
```
JimMadge, line 658: If this fails, is it a user error or does it probably mean something like no network connection or a name change in the Microsoft API?

If it isn't a problem the user introduced, maybe the error message could be clearer.

data_safe_haven/external/api/graph_api.py
```diff
@@ -573,12 +593,145 @@ def get_id_from_username(self, username: str) -> str | None:
...
+            application_sp = self.get_service_principal_by_name(application_name)
+            if not application_sp:
+                msg = "Could not find application service principal."
```
JimMadge:
```suggestion
                msg = f"Could not find application service principal for application {application_name}."
```

data_safe_haven/external/api/graph_api.py
```diff
@@ -573,12 +593,145 @@ def get_id_from_username(self, username: str) -> str | None:
...
+            )
+            # If there are existing permissions then we need to patch
+            application = next(
+                (
+                    app
+                    for app in response.json().get("value", [])
+                    if app["clientId"] == application_sp["id"]
+                ),
+                None,
+            )
```
JimMadge, line 732: Will this only ever get the first item in the generator? Is that correct?

I don't think I understand what is happening here.

data_safe_haven/external/api/graph_api.py
```diff
@@ -852,10 +1021,24 @@ def remove_user_from_group(
...
-                f"{self.base_endpoint}/groups/{group_id}/members/{user_id}/$ref",
-            )
+            # Check whether user is in group
+            json_response = self.http_get(
+                f"{self.base_endpoint}/groups/{group_id}/members",
+            ).json()
+            # Remove user from group if it is a member
+            if user_id in [
+                group_member["id"] for group_member in json_response["value"]
+            ]:
```
JimMadge:
```suggestion
            if user_id in (
                group_member["id"] for group_member in json_response["value"]
            ):
```

Got to save those clock cycles :smile:.

JimMadge (changes requested):
A few suggestions and questions.

I really like using dynamic components to keep things "in Pulumi" where we can.

I think I struggled a bit with the Graph API stuff because it feels quite abstract.
Regarding the user management commands, do we want CLI commands for convenience (avoid using a browser) or because it automates some quirks or extra steps DSH requires?

jemrobinson:
> Regarding the user management commands, do we want CLI commands for convenience (avoid using a browser) or because it automates some quirks or extra steps DSH requires?

I guess initially because I want to maintain our current setup of adding/removing users and groups using the CLI. We can definitely remove this in future, but I think it should be a conscious decision, not a by-product of another large change.

data_safe_haven/administration/users/azure_ad_users.py
jemrobinson: Yes, you're right. I can't see any reason why that wouldn't work.

_edit_: I think we need to use `==` not `in` since they're not the same object.

data_safe_haven/external/api/graph_api.py
jemrobinson: Granting roles to a service principal is done in two steps:

1. Assign the role to the application
2. Approve the request (this is granting permission)

Different RBAC permissions are needed for these two things. The second one will prompt for approval (which doesn't work on the command line) unless it has been approved by a Global Admin.

data_safe_haven/external/api/graph_api.py
jemrobinson: This function is doing both `assign role` and `grant permissions`. Can you think of a better name?

data_safe_haven/external/api/graph_api.py
jemrobinson: It probably means that Microsoft have done a major behind-the-scenes refactor, since the Microsoft Graph application (and service principal) are installed by default in all EntraID directories and are needed to make some REST calls (like this one).

Do you have a suggested better wording?

data_safe_haven/external/api/graph_api.py
jemrobinson: It's the equivalent of doing:

```
filtered = [x for x in <sequence> if <condition>]
application = filtered[0] if filtered else None
```

JimMadge: And we only ever need the first item here?

jemrobinson: Yes. We're trying to get the application with `clientId == application_sp["id"]` and there should only be one (or zero).

data_safe_haven/external/api/graph_api.py
JimMadge: What I'm thinking of is something to make sure the user knows it isn't a configuration problem. Like how `pip` will say something like "this is a problem with the package you are trying to install, non pip".

Maybe "This is likely a problem with dsh, not your configuration or infrastructure"?

data_safe_haven/external/api/graph_api.py
JimMadge: Is it always the same permissions? The only arguments are an application and a role.

JimMadge: Reading the other conversation, I think I might understand now :smile:.

data_safe_haven/external/api/graph_api.py
JimMadge: OK, I think I see. Is "granting permissions" MS language? I think that is what confused me because "assigning a role" sounds very RBAC where as "granting permissions" sounds quite ad-hoc (give individual users, individual permissions).

I suppose I would be tempted to just call it `assign_role` in the function name and docstring, then use comments in the function to explain that in practice that is two steps (assignment and approval).

Does that sound right?

data_safe_haven/external/api/graph_api.py
JimMadge: Same comment as in the other conversation. Something like `grant_role` for _our_ API and explain the details of how MS handles this in comments for future developers.

data_safe_haven/external/api/graph_api.py
JimMadge: Great, in that case maybe add a comment to explain that.

I think it would be easy to assume it is a mistake (at least I thought it might be :smile:).

data_safe_haven/external/api/graph_api.py
JimMadge: Is this resolved?

JimMadge (changes requested):
A few suggestions on tidying things up :+1:.

data_safe_haven/external/api/graph_api.py
jemrobinson: Done

data_safe_haven/external/api/graph_api.py
jemrobinson: Done

data_safe_haven/external/api/graph_api.py
jemrobinson: Done

data_safe_haven/external/api/graph_api.py
jemrobinson: Done

jemrobinson:
Thanks - I especially liked the clarity on explaining *our* API (not Microsoft's)!

JimMadge (approved):

PR #1787, opened 2024-04-12
Add dependabot configuration

JimMadge:


jemrobinson (approved):
I'm sure this is fine, but can you explain what it does? Maybe comments in the file since this is YAML?

PR #1803, opened 2024-04-15
Update dependabot config

JimMadge:

- Group dsh dependency version updates into a single PR
- Update `pyproject.toml` and `requirements.txt`

Supersedes #1801 #1797 #1795 #1792 #1789 #1788


pyproject.toml
```diff
@@ -32,7 +32,6 @@ dependencies = [
   "azure-storage-file-datalake~=12.10",
   "azure-storage-file-share~=12.10",
   "chevron~=0.14",
-  "chili~=2.1",
```
jemrobinson, line 35: Did we replace chili with pydantic?

requirements.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with Python 3.11
+# This file is autogenerated by pip-compile with Python 3.12
```
jemrobinson, line 2: Do we want to be using 3.12? I'd prefer to use latest-stable if possible, and I think that's still 3.11. What do you think?

jemrobinson (approved):
A couple of comments, but feel free to merge when they're addressed

pyproject.toml
JimMadge: Yes, I think this must have snuck back in through a merge somewhere.

requirements.txt
JimMadge: 3.12 is stable now.

I think sticking to the latest stable release is a good idea. An older, supported version (_e.g._ 3.11, 3.10, 3.9, 3.8 right now) would also be fine.
That might be sensible if a significant dependency takes time to update to the latest, or if a Python feature we use gets removed/changed.

JimMadge: We can update the `pyproject.toml` to say `>=3.12` or `==3.12`. Which do you think makes more sense?

jemrobinson: Let's have `==3.12` for the moment. Then updating can/should be a deliberate choice.

PR #1811, opened 2024-04-15
Correct Python version

JimMadge:

Changing `requires-python` to `==3.12` is a problem as this expands to `3.12.0`. This PR changes the version specification to `==3.12.*` which is equivalent to `>=3.12.0, < 3.13.0`

### DIFF
--- a/data_safe_haven/cli.py
+++ b/data_safe_haven/cli.py
@@ -19,13 +19,13 @@ from data_safe_haven.utility import LoggingSingleton
 
 def callback(
     output: Annotated[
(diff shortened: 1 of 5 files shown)


jemrobinson (approved):
LGTM

jemrobinson:
Has one of these changes updated the `ruff` version? It seems to be catching some things that haven't been recently changed.

JimMadge:
The Ruff version hasn't changed. I think Ruff can adjust its rules to match particular Python versions, so I would guess changing `pyproject.toml` changes its behaviour.

PR #1815, opened 2024-04-16
Add UniqueList annotated type

### LINKED ISSUE #1721: Pulumi: Use an Annotated Container Type for validation
Because this is quite similar to the database unique list validation, it might be possible to do something like create an Annotated Container Type `UniqueList[T]` and a validator for it.

Do you think that is possible? Let's open an issue if so and not block this.

_Originally posted by @JimMadge in https://github.com/alan-turing-institute/data-safe-haven/pull/1716#discussion_r1467558000_

JimMadge:

Closes #1721

### DIFF
--- a/data_safe_haven/utility/annotated_types.py
+++ b/data_safe_haven/utility/annotated_types.py
@@ -1,4 +1,5 @@
-from typing import Annotated
+from collections.abc import Hashable
+from typing import Annotated, TypeAlias, TypeVar
 
 from pydantic import Field
 from pydantic.functional_validators import AfterValidator
@@ -11,6 +12,7 @@ from data_safe_haven.functions.validators import (
     validate_fqdn,
     validate_ip_address,
     validate_timezone,
+    validate_unique_list,
 )
 
 AzureShortName = Annotated[str, Field(min_length=1, max_length=24)]
@@ -22,3 +24,9 @@ Fqdn = Annotated[str, AfterValidator(validate_fqdn)]
 Guid = Annotated[str, AfterValidator(validate_aad_guid)]
 IpAddress = Annotated[str, AfterValidator(validate_ip_address)]
 TimeZone = Annotated[str, AfterValidator(validate_timezone)]
+TH = TypeVar("TH", bound=Hashable)
+# type UniqueList[TH] = Annotated[list[TH], AfterValidator(validate_unique_list)]
+# mypy doesn't support PEP695 type statements
+UniqueList: TypeAlias = Annotated[  # noqa:UP040
+    list[TH], AfterValidator(validate_unique_list)
+]
--- a/data_safe_haven/config/config.py
+++ b/data_safe_haven/config/config.py
@@ -26,6 +26,7 @@ from data_safe_haven.functions import (
     b64decode,
     b64encode,
 )
+from data_safe_haven.functions.validators import validate_unique_list
 from data_safe_haven.utility import (
     DatabaseSystem,
     LoggingSingleton,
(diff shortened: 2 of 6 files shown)


data_safe_haven/utility/annotated_types.py
```diff
@@ -22,3 +24,9 @@
 Guid = Annotated[str, AfterValidator(validate_aad_guid)]
 IpAddress = Annotated[str, AfterValidator(validate_ip_address)]
 TimeZone = Annotated[str, AfterValidator(validate_timezone)]
+TH = TypeVar("TH", bound=Hashable)
+# type UniqueList[TH] = Annotated[list[TH], AfterValidator(validate_unique_list)]
+# mypy doesn't support PEP695 type statements
+UniqueList: TypeAlias = Annotated[  # noqa:UP040
+    list[TH], AfterValidator(validate_unique_list)
+]
```
jemrobinson, line 32: Why do we need this instead of `UniqueList: TypeAlias = Annotated[list[Hashable], AfterValidator(validate_unique_list)]`?

jemrobinson (approved):
LGTM. One question about how the type definition works.

data_safe_haven/utility/annotated_types.py
JimMadge: I think that gave me a typing error like "expected `list[Hashable]` but got `list[int]`". So I think that type annotation would be "A list with items of the type class: Hashable" rather than "A list of objects where `isinstance(x, Hashable) == True`".

Maybe I did something wrong.

JimMadge:
```
data_safe_haven/config/config.py:155: error: Bad number of arguments for type alias, expected: 0, given: 1  [type-arg]
data_safe_haven/config/config.py:198: error: Incompatible types in assignment (expression has type "list[DatabaseSystem]", variable has type "list[Hashable]")  [assignment]
data_safe_haven/config/config.py:198: note: "List" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
data_safe_haven/config/config.py:198: note: Consider using "Sequence" instead, which is covariant
data_safe_haven/config/config.py:202: error: "Hashable" has no attribute "value"  [attr-defined]
data_safe_haven/config/config.py:254: error: Argument 1 to "validate_unique_list" has incompatible type "list[int]"; expected "list[Hashable]"  [arg-type]
data_safe_haven/config/config.py:254: note: "List" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
data_safe_haven/config/config.py:254: note: Consider using "Sequence" instead, which is covariant
data_safe_haven/infrastructure/stacks/declarative_sre.py:271: error: Argument "databases" to "SREUserServicesProps" has incompatible type "list[Hashable]"; expected "list[DatabaseSystem]"  [arg-type]
Found 5 errors in 2 files (checked 107 source files)
```

jemrobinson: OK, I guess the only other solution would be to put `x: list[Hashable] = [1, 2, 3]` everywhere which is *much* worse. I'm not totally sure why the current solution works, but I'm glad it does!

JimMadge: @jemrobinson Happy to accept what is here or can you see a way around that?

jemrobinson:
Given that a `UniqueList` is actually a `set()` could/should we be using `set` everywhere instead?

data_safe_haven/utility/annotated_types.py
jemrobinson: Definitely happy to accept this. See below for my remaining question about whether we should just be using `set`.

JimMadge:
> Given that a `UniqueList` is actually a `set()` could/should we be using `set` everywhere instead?

That is a good question. I think we don't want to do that here because we are parsing user input. If they provide an invalid input, we should raise an error rather than trying to fix it. `set` would remove duplicates rather than raising an exception.

PR #1820, opened 2024-04-17
Move pulumi context settings and persistent data

### LINKED ISSUE #1689: Move encoded stacks from config to another location

JimMadge:

- Add `ConfigClass` base class for configuration synced to Azure Storage in the Context
  - Add `from_remote_or_create` method to fetch config if it exists in Azure Storage, otherwise create a new instance
- Add `DSHPulumiProject` and `DSHPulumiConfig` classes for storing persistent Pulumi project information
- Use `ConfigClass` for both `Config` and `DSHPulumiConfig`
- Remove the need for `Context` to construct `Config`. Removes a lot of potential circular imports.
- Use information from `DSHPulumiProject` instances to create Pulumi workspaces and stacks, avoiding the need to write `Pulumi.<stack>.yaml` and `Pulumi.yaml` files in a non-temp work directory
- Propagated above changes through infrastructure and commands modules
- Many tests
  - Adding initial tests for `StackManager` classes
  - Add mocking frameworks for testing some Azure/Pulumi functions

Closes #1689 #1828 

### :microscope: Tests

New deployment, repeating commands successfully fetches Pulumi persistent data (stack settings, stack state)


JimMadge:
Working on this, the names are becoming a mess.

Terms [here](https://www.pulumi.com/docs/using-pulumi/automation-api/concepts-terminology/).

We have two Pulumi Programs, one for SHM, one for SRE.
Each 'thing' we deploy is a new Pulumi Project, with a Local Workspace containing project configuration `Pulumi.yaml` and _a single_ stack with configuration `Pulumi.<stack>.yaml`. (We might not need the Local Workspace at all #1828)

We _could_ use stacks a bit more like they are intended, with dev and prod deployments. I don't see us doing that now though. So, the state is, each 'thing' is the only stack of a Pulumi Project.

I think that has lead to conflating the terms program/stack and not using the term project.

I propose

- `DSHPulumiConfig` - Class to hold persistent Pulumi information that is written to the config share (multiple instances of `DSHPulumiProject`. _I.e._ _not_ stacks which are in the Pulumi backend, _not_ the Pulumi backend address/encryption key which are in the context
- `DSHPulumiProject` - Class to hold persistent information about Pulumi projects, which right now is just a name and a single stack configuration (like `Pulumi.<stack>.yaml`)
- We should be more strict and consistent with naming and be careful not to call projects stacks

@craddm @jemrobinson thoughts?

JimMadge:
Also, I must be tired because I implemented a class where you can look up items using strings, which must be unique, and didn't use a `dict` 🤦.

jemrobinson:
Can you give a quick sketch of what would be in each of these config files? Could be e.g. a YAML representation of the pieces of data in each of them.

JimMadge:
> Can you give a quick sketch of what would be in each of these config files?

The Pulumi files, or the classes we are creating?

For the later, I would write the `DSHPulumiConfig` class to the storage account in the context in a similar way to `Config`, taking advantage of the Pydantic serializer.

jemrobinson:
> The Pulumi files, or the classes we are creating?

The data structure of the classes we're creating (I'm not worried about the serialisation part). I'm thinking something like:

```yaml
DSHPulumiConfig
- azure
  - subscription_id
  - tenant_id
- something_else
```

```yaml
DSHPulumiProject
- something
  - A
  - B
- something_else
  - C
  - D  
```

JimMadge:
Currently,

```
DSHPulumiConfig:
  projects:
    - DSHPulumiProject: # e.g. an SHM or an SRE
      - name: a
      - config: ...
    - DSHPulumiProject: # e.g. an SHM or an SRE
      - name: b
      - config: ...
```

jemrobinson:
What do we gain by allowing multiple `DSHPulumiProject` to live under one `DSHPulumiConfig`? As written, there are no keys which belong to the top-level `DSHPulumiConfig` object except instances of `DSHPulumiProject`. I'm sure there's more to what you're suggesting than that.

data_safe_haven/config/pulumi.py
```diff
@@ -0,0 +1,147 @@
...
+        with open(config_path, "w") as f_config:
+            f_config.write(self.config)
+
+
+class PulumiConfig(BaseModel, validate_assignment=True):
+    stacks: UniqueList[PulumiStack]
+
+    def __getitem__(self, key: str) -> PulumiStack:
+        if not isinstance(key, str):
+            msg = "'key' must be a string."
```
jemrobinson, line 32:
```suggestion
            msg = "'{key}' must be a string."
```

JimMadge:
> What do we gain by allowing multiple `DSHPulumiProject` to live under one `DSHPulumiConfig`? As written, there are no keys which belong to the top-level `DSHPulumiConfig` object except instances of `DSHPulumiProject`. I'm sure there's more to what you're suggesting than that.

Being able to look up projects by a key, which I believe we need when Pulumi outputs are shared between projects, `dsh_pulumi_config['my_shm']`, `dsh_pulumi_config['my_sre']`. Serialising all Pulumi projects and writing them to a file in the context.

It is essentially a container at the moment, but we want it to have move functionality than just a plain `dict` or `list`. If there is more persistent Pulumi data, default args for example, this would be a natural place to store them.

data_safe_haven/administration/users/guacamole_users.py
```diff
@@ -3,18 +3,21 @@
 from typing import Any
 
 from data_safe_haven.config import Config
+from data_safe_haven.config.context_settings import Context
```
jemrobinson: As a rule-of-thumb (can't find a reference to it at the moment) it's better not to use deeply-nested imports in Python code. This makes your dependencies more fragile and means that it's harder to refactor. In this case, I think we'd be better with the following:

```suggestion
from data_safe_haven.config import Config, Context
```

with an equivalent change to `config > __init__.py`.

data_safe_haven/infrastructure/stack_manager.py
```diff
@@ -57,20 +59,22 @@ class StackManager:
...
     def __init__(
         self,
+        context: Context,
         config: Config,
         program: DeclarativeSHM | DeclarativeSRE,
     ) -> None:
-        self.account = PulumiAccount(config)
+        self.account = PulumiAccount(context, config)
+        self.context: Context = context
         self.cfg: Config = config
```
jemrobinson: Not sure you need type markers here since the type is defined in the function signature.

data_safe_haven/config/pulumi.py
JimMadge: This is intentional, 'key' the argument rather than `{key}` the value.

data_safe_haven/administration/users/guacamole_users.py
JimMadge: I did this because there were circular dependencies.

I think that might not be a problem any more so I'll look to tidy that.

jemrobinson: For local dependencies I prefer `from .context_settings import Context`. If they're really circular, we should think about how to break the coupling.

data_safe_haven/config/config_class.py
```diff
@@ -0,0 +1,69 @@
...
+from data_safe_haven.exceptions import (
+    DataSafeHavenConfigError,
+    DataSafeHavenParameterError,
+)
+from data_safe_haven.external import AzureApi
+
+T = TypeVar("T", bound="ConfigClass")
+
+
+class ConfigClass(BaseModel, validate_assignment=True):
```
jemrobinson: I prefer more descriptive class names if possible. How about?

```suggestion
class SerialisableConfig(BaseModel, validate_assignment=True):
```

data_safe_haven/functions/__init__.py
```diff
@@ -35,6 +36,7 @@
     "random_letters",
     "replace_separators",
     "seeded_uuid",
+    "sanitise_sre_name",
```
jemrobinson:
```suggestion
    "sanitise_sre_name",
    "seeded_uuid",
```

jemrobinson (changes requested):
I haven't gone through this in detail but I've got a couple of general thoughts.

1. I prefer to keep one class per file, with the filename as a snake-case name of the class. This makes it a bit easier (for me) to find where things are defined. Do you think that's sensible?
2. I saw you've replaced the local workdir setting - does this work?
3. I can't get this to work! Have you tested it? If so, can you give me some instructions about how to do it?

data_safe_haven/administration/users/guacamole_users.py
JimMadge: Done in 2012ff6

data_safe_haven/commands/deploy.py
```diff
@@ -28,6 +31,8 @@ def shm(
     """Deploy a Safe Haven Management component"""
     context = ContextSettings.from_file().assert_context()
     config = Config.from_remote(context)
```
jemrobinson, line 32: What if there's no remote file (e.g. on first run)?

data_safe_haven/commands/deploy.py
```diff
@@ -28,6 +31,8 @@ def shm(
...
+    pulumi_config = DSHPulumiConfig.from_remote(context)
```
jemrobinson: What if there's no remote file (e.g. on first run)?

JimMadge: Good point. I'll add something like "get or create DSHPulumiConfig" :+1:.

jemrobinson: What command creates the remote Pulumi config?

JimMadge: To write it to Azure Storage? `DSHPulumiConfig.upload` but you would need to check if one already exists and create an empty instance of the class if not.

jemrobinson: I'm thinking about the workflow that someone would have when trying to deploy this. Something like:

```
dsh context add # makes the local context
dsh context create # deploys Azure resources
dsh config template > config.yaml # creates a local YAML file
dsh config upload config.yaml # upload the local config to Azure
dsh deploy shm
```

this last line breaks because it looks for a remote Pulumi config but none of the previous commands have created it.

Also - this workflow seems a bit overcomplicated, but we can fix that elsewhere.

JimMadge: I see, I would have it happen at `$ dsh deploy shm`.

JimMadge: I prefer to keep the workflow explicit like that. I wouldn't like the command which adds a description of a context to also deploy infrastructure.

JimMadge:
CI is failing because Pulumi non-interactive login is failing. Not sure if there is a way to do a non-interactive local login.

data_safe_haven/commands/deploy.py
jemrobinson: Fair enough. Perhaps we could replace `dsh config upload config.yaml` with something like `dsh deploy shm --config config.yaml` which would prompt for "are you sure?" if there's a remote config that differs from `config.yaml`.

jemrobinson:
> CI is failing because Pulumi non-interactive login is failing. Not sure if there is a way to do a non-interactive local login.

Can we mock it?

JimMadge:
> > CI is failing because Pulumi non-interactive login is failing. Not sure if there is a way to do a non-interactive local login.
> 
> Can we mock it?

No, or, not unless we mock the entire Pulumi CLI :smile:.

jemrobinson:
> > > CI is failing because Pulumi non-interactive login is failing. Not sure if there is a way to do a non-interactive local login.
> > 
> > 
> > Can we mock it?
> 
> No, or, not unless we mock the entire Pulumi CLI 😄.

This might be time to take a deeper look at the Pulumi testing framework (especially the unit-testing part: https://www.pulumi.com/docs/using-pulumi/testing/unit/).

... in another issue though, preferably.

JimMadge:
I think I've fixed that, although there is another problem.
It works locally, so I think _that_ part is OK but it might be failing due to lack of Pulumi/Azure CLI credentials.

data_safe_haven/commands/deploy.py
JimMadge: Fixed in 65ad597.

I've only put it in for deploy SHM for now (as the other commands couldn't work without this file already existing, I think).

data_safe_haven/commands/deploy.py
JimMadge: I think that sounds sensible. It would be easy (and probably useful) to keep the `$ dsh config upload ...` command. Both functions should just need to call a couple of `Config` methods.

Could you open an issue, if there isn't one already.

jemrobinson: If this is in a usable state by this afternoon, we can discuss it during a live run-through.

jemrobinson: I'm happy to have more powerful advanced commands (or options to commands) but I'd like to keep the minimal required commands simple and easy to understand. If possible, they would just be: `dsh initialise; dsh deploy` but I realise we're not there at the moment!

JimMadge: I think this is a philosophical argument.

I would say that isn't simple or easy to understand. Having so few commands means they must be complex and handle a lot of processes without input from the user.

I would say something more like a tree (`dsh shm {deploy, destroy}`, `dsh sre {deploy, destroy}` is simple and easy to understand because the commands are clear, limited and descriptive.

Could you open an issue so I can resolve the conversation here?

data_safe_haven/config/config_class.py
JimMadge: Done in 7a8cc97

data_safe_haven/commands/deploy.py
jemrobinson: See #1836 for discussion.

JimMadge:
> I prefer to keep one class per file, with the filename as a snake-case name of the class. This makes it a bit easier (for me) to find where things are defined. Do you think that's sensible?

I wouldn't be so strict as one class per file. That might end up with lots of very short files. I find it more difficult to find things if I have to start guessing at filenames or using grep. It also means you might need a lot of buffers open to make a small change.

However, I like files to have a clear purpose and not be very long.

JimMadge:
This is working now and I think I've addressed all the comments.

Issue #1802, opened 2024-04-15
Add load balancing in front of Azure Container Instances


jemrobinson:
## :white_check_mark: Checklist

- [x] I have searched open and closed issues for duplicates.
- [x] This is a request for a new feature in the Data Safe Haven or an upgrade to an existing feature.
- [x] The feature is still missing in the [latest version](https://github.com/alan-turing-institute/data-safe-haven/releases).
- [x] I have read through the [documentation](https://alan-turing-institute.github.io/data-safe-haven/).
- [x] This isn't an open-ended question (open a [discussion](https://github.com/alan-turing-institute/data-safe-haven/discussions) if it is).

## :strawberry: Suggested change

Azure Container Instances need to be deployed to a subnet and are given an IP address from that subnet. This is usually (but not always) the first available IP address. When they are rebooted, they often change IP address and other services that refer to them are not updated.

See [here](https://learn.microsoft.com/en-us/azure/container-instances/container-instances-update#limitations) for more details.

## :steam_locomotive: How could this be done?

If we add appropriate load balancing in front of all ACI services, we can ensure that they're reachable regardless of their exact IP address.

jemrobinson:
Placing an Azure Load Balancer in front of container instances in a networked container group is not currently supported. See [here](https://github.com/MicrosoftDocs/azure-docs/issues/69658) for more information.

jemrobinson:
Consider an `az cli` sidecar like [this](https://www.garyjackson.dev/posts/azure-container-instance-dns-private-ip/) or [this](https://zure.com/blog/detect-changing-ip-addresses-of-azure-container-instances-in-vnet/).

[Here](https://github.com/tmeckel/aci-dns-manager) is one that's already been written, using an Azure service principal to make the necessary changes.

JimMadge:
This is concerning and surprising.

I think we should look at alternatives as it feels like perhaps container instances are not suppose to be used in this way.

Switching to kubernetes might save us a lot of headaches.

jemrobinson:
I looked at Kubernetes when I first started the rewrite. You really need to buy into it from the beginning and structure the whole project around it. Sticking with Azure, it's possible that e.g. Azure Container Apps might be a good fit here. Equally, Azure might start supporting Load Balancing soon (doesn't feel like it should be a large amount of effort for them).

JimMadge:
The load balancing issues seem to have been open for a while and it looks like the feature was removed from the roadmap.

[...]

PR #1821, opened 2024-04-17
Add local DNS for SRE identity server

jemrobinson:

Add a local DNS record for the SRE identity server which points to the Azure Container Instance private IP address. This does not fix the issue identified in #1802 but it means that changes only need to be made in the private DNS zone, which should be easier to do.

Contributes to #1802.

### :microscope: Tests

Tested on a pre-existing SRE


craddm (approved):
LGTM

PR #1827, opened 2024-04-19
Fix dependabot

jemrobinson:

Dependabot is currently failing to correctly update `requirements.txt` from `pyproject.toml` (see e.g. #1809). It is also double-processing the requirements under `docs`.

This PR replaces the Python part of Dependabot's config with the following

- A new `docs` environment in `pyproject.toml`
- A workflow which uses `pip-compile` to update `requirements.txt` and `docs/requirements.txt`
- Changes all three `update-X` workflows to run both on pushes to `develop` and every week.

Closes #1809

### :microscope: Tests

Untested


.github/dependabot.yml
```diff
@@ -4,30 +4,6 @@ version: 2
 updates:
   # GitHub Action update PRs
   - package-ecosystem: "github-actions"
-    directory: "/"  # Checks /.github/workflows/
+    directory: "/"  # This is recursive so it also checks .github/workflows/
```
JimMadge: I think that isn't why this work.

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#directory

> For GitHub Actions, you do not need to set the directory to /.github/workflows. Configuring the key to / automatically instructs Dependabot to search the /.github/workflows directory, as well as the action.yml / action.yaml file from the root directory.

.github/workflows/update_python_dependencies.yaml
```diff
@@ -0,0 +1,67 @@
...
+          python-version: 3.12
+
+      - name: Install dependencies
+        run: pip install hatch pip-tools
+
+      - name: Update 'default' dependencies
+        run: pip-compile -U pyproject.toml -o requirements.txt
+
+      - name: Update 'docs' dependencies
+        run: hatch env show docs --json | jq -r '.docs.dependencies | .[]' | pip-compile -U - -o docs/requirements.txt
```
JimMadge: I think this was one of the reasons I went with a `requirements.in`, because there isn't a convenient command like `pip-compile pyproject.toml --env docs`.

If we could make that a script rather than being hardcoded in the workflow that could be good.

pyproject.toml
```diff
@@ -58,6 +58,15 @@ path = "data_safe_haven/version.py"
 [tool.hatch.envs.default]
 pre-install-commands = ["pip install -r requirements.txt"]
 
+[tool.hatch.envs.docs]
```
JimMadge, line 61:
```suggestion
[tool.hatch.envs.docs]
pre-install-commands = ["pip install -r docs/requirements.txt"]
```

To make sure you have the pinned dependencies like for the default env.

.github/workflows/update_docker_versions.yaml
```diff
@@ -1,7 +1,9 @@
 ---
-name: Update Pulumi Docker image versions
+name: Update Docker image versions
 
 on:  # yamllint disable-line rule:truthy
+  push:
+    branches: [develop]
```
JimMadge: If we have weekly runs I wouldn't feel the need to also trigger these workflows on pushes.

(Didn't we recently remove that from some workflows :smile:?)

JimMadge (changes requested):
Looks good.

A few suggestions.
I would prefer not to run the update workflows periodically _and_ on pushes.

.github/dependabot.yml
jemrobinson: I understand what that's saying, but I think the reason *why* is works is because Dependabot searches the full tree under the path you provide.

For example, these are the current Dependabot settings for `pip`.

This one with

```
package-ecosystem: "pip"
directory: "/"
```

Screenshot 2024-04-19 at 15 28 03

checks both the base directory and the `docs` directory

... while this one with

```
package-ecosystem: "pip"
directory: "/docs/"
```

Screenshot 2024-04-19 at 15 28 09

checks only the `docs` directory.

.github/workflows/update_python_dependencies.yaml
jemrobinson: I think the main reason for going with `requirements.in` here was that we didn't have a top-level `pyproject.toml` as this workflow existed in the Powershell version.

There is a [hatch-pip-compile](https://github.com/juftin/hatch-pip-compile) plugin, but we need the generation of the `requirements` file to happen outside `hatch`, as one of the hatch steps is to install from this file!

pyproject.toml
jemrobinson: Sure, but `docs` doesn't currently have any commands in it. Let me add some!

.github/workflows/update_docker_versions.yaml
jemrobinson: Happy to do both/either. Dependabot was set up weekly, so I can copy that.

pyproject.toml
jemrobinson: Done in 5a03056

.github/workflows/update_docker_versions.yaml
jemrobinson: Done in ec370e6da

.github/dependabot.yml
JimMadge: Ah, interesting!

I guess that would mean those two definitions would both try to update the docs requirements which isn't what we want.

JimMadge: Maybe what the docs says is a subset of "directories will be searched recursively".

pyproject.toml
JimMadge: Even better :+1:.

That's actually a big advantage of using an env. `hatch run docs:build` instead of making/sourcing a venv then running make.

.github/workflows/update_python_dependencies.yaml
JimMadge: > I think the main reason for going with requirements.in here was that we didn't have a top-level pyproject.toml as this workflow existed in the Powershell version.

That makes sense :smile:.

That plugin sounds a bit awkward. If that chain of commands were just in a file like `update_docs_requirements.sh` I just think it would be more visible, easier to reproduce out of CI.

jemrobinson: ~I don't think the overhead of moving this into its own script is worthwhile. It would make sense if we were doing so for more than one environment though.~

Actually, I've changed my mind and put it into a short bash script in 1d8303122

.github/scripts/update_python_dependencies.sh
```diff
@@ -0,0 +1,23 @@
...
+if ! command -v pip-compile > /dev/null; then
+    echo "pip-compile could not be found"
+    exit 1
+fi
+
+# Run pip-compile
+if [ "$ENV_NAME" = "default" ]; then
+    pip-compile -U pyproject.toml -o "$TARGET"
+else
+    hatch env show --json | jq -r ".${ENV_NAME}.dependencies | .[]" | pip-compile -U - -o "$TARGET"
```
JimMadge:
```suggestion
    hatch env show --json | jq -r ".${ENV_NAME}.dependencies | .[]" | pip-compile -U -o "$TARGET"
```

Assuming that is a mistake.

JimMadge (changes requested):

.github/scripts/update_python_dependencies.sh
```diff
@@ -0,0 +1,23 @@
+#! /bin/sh
```
JimMadge:
```suggestion
#! /usr/bin/env sh
```

.github/scripts/update_python_dependencies.sh
jemrobinson: No, this is necessary. It means `pip-compile -U <the result of the pipeline>` while without this you get `pip-compile -U` which defaults to the `default` environment of `pyproject.toml`.

jemrobinson: Would you prefer `pip-compile - -U -o "$TARGET"`?

jemrobinson:
```suggestion
    hatch env show --json | jq -r ".${ENV_NAME}.dependencies | .[]" | pip-compile - -U -o "$TARGET"
```

JimMadge (changes requested):

JimMadge (approved):

PR #1839, opened 2024-04-26
Pulumi storage fixes, including adding encryption key

craddm:

Adds additional fixes from co-working session, and preliminary attempts to include encryption key in `pulumi.yaml`


data_safe_haven/commands/context.py
```diff
@@ -18,7 +19,11 @@
 @context_command_group.command()
 def show() -> None:
     """Show information about the selected context."""
-    settings = ContextSettings.from_file()
+    try:
+        settings = ContextSettings.from_file()
+    except DataSafeHavenConfigError:
+        print("No context selected.")
+        return
```
JimMadge: We should add a test for the case where there is no context.

data_safe_haven/commands/deploy.py
```diff
@@ -77,7 +74,9 @@ def shm(
     except DataSafeHavenError as exc:
         msg = f"Could not deploy Data Safe Haven Management environment.\n{exc}"
         raise DataSafeHavenError(msg) from exc
-
+    finally:
+        # Upload Pulumi config to blob storage
+        pulumi_config.upload(context)
```
JimMadge, line 80: I can't remember if there are tests for the deploy command but we should add one to check that upload is always called whether or not an exception is raised.

data_safe_haven/config/pulumi.py
```diff
@@ -10,6 +10,7 @@
 class DSHPulumiProject(BaseModel, validate_assignment=True):
     """Container for DSH Pulumi Project persistent information"""
 
+    encrypted_key: str
```
JimMadge, line 13: Again, need to update tests for this change.

JimMadge (commented):
All good changes, do you want to get these merged into `pulumi_storage` or shall we work on this branch until we have fixed all issues? @craddm @jemrobinson

jemrobinson:
@JimMadge let's merge this in (once your comments are addressed). If we need to make more changes we can do that on a new branch.

data_safe_haven/commands/context.py
jemrobinson: Can you add this or open an issue? I think you're most familiar with this set of changes and the test suite.

data_safe_haven/commands/deploy.py
jemrobinson: Can you add this or open an issue?

craddm:
Are we just merging this in before getting the encrypted key thing working, and then getting it working on the main `pulumi_storage` branch?

JimMadge:
With regards to opening issues, I'm not really comfortable deferring tests or merging changes which break PR. I don't think there is any great rush here so we should do that before merging.

jemrobinson:
Also agree with Matt that we should get the encrypted key piece working. Are you able to test this @JimMadge ?

JimMadge:
Good point, we should check that change is doing what we hope :+1:. If not we can cherry-pick or merge the other changes anyway.

I'll try to find time to test this. This set of PRs are my only priority when it comes to development.

data_safe_haven/commands/context.py
JimMadge: fixed in a3d9eba2686240f73964dbafa13a14ddbaaa24f0 and fb280b9536e12fcd82bf9c3624f8f97f9ed619b4

data_safe_haven/config/pulumi.py
JimMadge: done in 1f0b05a840abac019e46bf44d8d404981f8c7a2d and 42a11c0164f66e40d89ad43de464ed4975a15d76

data_safe_haven/commands/deploy.py
JimMadge: Done in [db1fc3d](https://github.com/alan-turing-institute/data-safe-haven/pull/1839/commits/db1fc3d40bea75dfd850c4e7dd0cf1ed4a4277ae)

JimMadge (approved):

PR #1840, opened 2024-04-26
Separate config classes

jemrobinson:

Separate out serialisable classes (YAMLSerialisable and AzureSerialisable) and ensure that all config files (config, pulumi and contextsettings) are inheriting from one of these.

n/a

### :microscope: Tests

Tested locally


JimMadge (changes requested):
Can we make sure we have tests for all of these changes.

Maybe we should adjust the triggers so that tests/linting run on PRs to any branch?

jemrobinson:
Is this already done? The YAMLSerialisable and AzureSerialisable classes aren't intended to be used standalone, and the classes that inherit from them are already tested.

JimMadge:
> Is this already done? The YAMLSerialisable and AzureSerialisable classes aren't intended to be used standalone, and the classes that inherit from them are already tested.

No, there are tests for `SerialisableConfig` which are broken now.

I think we should still write tests with simple examples of how we expect these classes to be used. You could look at the `SerialisableConfig` tests for examples.

jemrobinson:
@JimMadge: I've written some tests, but note that I actually *don't* expect these classes to be used on their own :)

data_safe_haven/utility/__init__.py
```diff
@@ -1,17 +1,23 @@
...
 from .types import PathType
+from .yaml_serialisable_model import YAMLSerialisableModel
 
 __all__ = [
+    "AzureLocation",
+    "AzureLongName",
     "config_dir",
     "DatabaseSystem",
     "FileReader",
+    "Guid",
```
JimMadge, line 16: I would prefer to import things like this from the modules rather than putting them in `utility`.

It isn't really clear what a `utility.AzureLocation` is but it is clear what a `utility.annotated_types.AzureLocation` is.
I want to avoid `utility` becoming a collections of miscellaneous things.

tests/context/test_context_settings.py
```diff
@@ -152,7 +152,7 @@ def test_missing_selected(self, context_yaml):
         context_yaml = "\n".join(context_yaml.splitlines()[1:])
         msg = "\n".join(
             [
-                "Could not load context settings.",
+                "Could not load YAMLSerialisableModel configuration.",
```
JimMadge: Shouldn't this be "Could not load ContextSettings configuration." using `cls.config_type`?

tests/utility/test_yaml_serialisable_model.py
JimMadge, line 1: Missing tests for `from_pathpath` and `to_filepath` here.

JimMadge (changes requested):
Looks good, a few little things to tidy up.

> I've written some tests, but note that I actually don't expect these classes to be used on their own

Yes but I don't think I follow your point. If we were to write an ABC we would still need to test it. I think the model here of making a toy "real class" in the tests is a sensible way to do that and neatly separates testing these classes from our real "real classes".

data_safe_haven/utility/__init__.py
jemrobinson: I'd like to keep things organised in a way that *everything* that needs to be imported from outside the same folder can be imported as `from data_safe_haven.<folder name> import <something>`. We might need some thoughts about better ways to group things (I absolutely *do* agree that `utility` has the potential to become a messy misc folder if we're not careful).

jemrobinson: For example, I think it would make sense to put `annotated_types.py` and `types.py` into a directory called `types` so that we could use `from data_safe_haven.types import AzureLocation`. Probably better as its own PR though.

JimMadge: And you want at most one level, like `data_safe_have.types` rather than `data_safe_haven.utility.types`?

I don't think I'm too bothered unless you have a very large number of levels. As long as it makes sense. I could see annotated types being part of utilities, or its own module.

Is there a risk from confusing types and annotated types? 😱

jemrobinson: Here's the essay/blog I was looking for earlier: https://benhoyt.com/writings/python-api-design/. Have a look and see whether you agree or disagree :).

tests/context/test_context_settings.py
jemrobinson: Added a missing config_type.

tests/utility/test_yaml_serialisable_model.py
jemrobinson: Done in 51e647141.

data_safe_haven/utility/__init__.py
JimMadge: I scanned through the blog post. I think the advice is all good.

Particularly, how making a simple, flat API means you can easily avoid breaking the API even when restructuring.
In our case, it isn't _as big_ a deal as we don't expect anyone to use this package as an API.

I think we can take that approach then, but we have (or might have) some other problems we need to fix,

- modules/packages that are not descriptive (enough). What is a `utility.Guid`, what is a `utility.FileReader`. It would be better then to have a `data_safe_haven.types`, `data_safe_haven.annotated_types` modules/packages. (In the first instance, they can probably be modules to avoid a boilerplate `__init__.py`.
- Redundant packages where you have a directory with only `__init.py__` and `a.py`, and the `__init__.py` just imports and exposes everything from `a.py`

jemrobinson: I think we have two APIs actually.

1. The structure of commands under the `dsh` CLI app (which I think you very sensibly wanted to organise as something like `dsh <segment> <verb>`. This one could be structured like the DSH data model ("these things belong to the SRE...")

2. The internal API of connections between modules inside the project. This one could be structured like the Python data model ("these things are all subclasses of X").

JimMadge: Issue [here](https://github.com/alan-turing-institute/data-safe-haven/issues/1846)

JimMadge (approved):

Issue #1836, opened 2024-04-26
Structure of commands


jemrobinson:
## :white_check_mark: Checklist

- [x] I have searched open and closed issues for duplicates.
- [x] This is a request for a new feature in the Data Safe Haven or an upgrade to an existing feature.
- [x] The feature is still missing in the [latest version](https://github.com/alan-turing-institute/data-safe-haven/releases).
- [x] I have read through the [documentation](https://alan-turing-institute.github.io/data-safe-haven/).
- [x] This isn't an open-ended question (open a [discussion](https://github.com/alan-turing-institute/data-safe-haven/discussions) if it is).

## :strawberry: Suggested change

We should agree on the structure of our user-facing command "API".

Current workflow looks something like:

```
dsh context add # makes the local context
dsh context create # deploys Azure resources
dsh config template > config.yaml # creates a local YAML file
dsh config upload config.yaml # upload the local config to Azure
dsh deploy shm # deploy an SHM
... some time passes ...
dsh config show > config.yaml # download the config file locally and edit it
dsh config upload config.yaml # upload the new config with e.g. an SRE called "sandbox" added
dsh deploy sre sandbox # deploy an SRE
```

I think this could be simplified and also made less (potentially) error-prone by preventing users from clobbering existing config files unless this is done deliberately.

## :steam_locomotive: How could this be done?

Have a discussion about what commands (and options) should be exposed.

jemrobinson:
Some discussion on #1820:

> Perhaps we could replace `dsh config upload config.yaml` with something like `dsh deploy shm --config config.yaml` which would prompt for "are you sure?" if there's a remote config that differs from `config.yaml`.

> I think that sounds sensible. It would be easy (and probably useful) to keep the `$ dsh config upload ...` command. Both functions should just need to call a couple of `Config` methods.

> I'm happy to have more powerful advanced commands (or options to commands) but I'd like to keep the minimal required commands simple and easy to understand. If possible, they would just be: `dsh initialise; dsh deploy` but I realise we're not there at the moment!

> I think this is a philosophical argument.
>
> I would say that isn't simple or easy to understand. Having so few commands means they must be complex and handle a lot of processes without input from the user.
>
> I would say something more like a tree (`dsh shm {deploy, destroy}`, `dsh sre {deploy, destroy}` is simple and easy to understand because the commands are clear, limited and descriptive.

JimMadge:
[...]

PR #1848, opened 2024-04-30
File structuring

jemrobinson:

An idea of a possible file/inheritance restructuring.

n/a

### :microscope: Tests

Local tests passing


JimMadge (commented):
I like this structure,

- I like keeping context configuration and DSH configuration separate. They do feel like quite different things to me.
- We _could_ pass the information `AzureSeialisableModel.upload` needs instead of a `Context` instance to remove that import.
- Having the serialisers in their own package feels right.
- I still wonder if breaking the types down more would be useful. `data_safe_haven.enums.ANIMAL` might be clearer than `data_safe_haven.types.ANIMAL`.
- `utility` is a lot neater now. I don't mind having that or a `helpers` package.
- Breaking up the `functions` package might be good. In particular we could have
  ```python
   from data_safe_haven import validators

   validators.timezone(a)
   ```
   instead of 
   ```python
   from data_safe_haven import functions

   functions.validate_timezone(a)
   ```

JimMadge:
Also, I think we have some sensible nesting now. Like in the `infrastructure` package, where the top level `__init__.py` only exports what you need at a high level, the stack managers. The finer details are in sub-packages which are only needed in `infrastructure` anyway :+1:.

JimMadge:
This may also be an opportunity to rename `StackManager` to `ProjectManager` or similar.

JimMadge:
I'm going to try making some of those changes.

jemrobinson:
I would also like to move the AzureSDK deployment stuff (currently in `context > context_infra` into `infrastructure`) if you're comfortable with that.

JimMadge:
> I would also like to move the AzureSDK deployment stuff (currently in `context > context_infra` into `infrastructure`) if you're comfortable with that.

I think we should be consistent. `config` and `infrastructure` are separate so `context` and `context_infra` should be too.

I think maybe a `context_infrastructure` package/module might be better as `infrastructure` is currently all Pulumi stuff. (I think it used to be called `pulumi`?)

jemrobinson:
I can see the argument behind keeping AzureSDK and Pulumi deployment stuff separate, but perhaps this could look like:

```
data_safe_haven/infrastructure
├── __init__.py
├── sdk
│   ├── __init__.py
│   ├── some
│   ├── files/folders
│   └── here
└── pulumi
    ├── __init__.py
    ├── some
    ├── files/folders
    └── here
```

with

```
data_safe_haven/infrastructure/__init__.py

from .pulumi import SHMStackManager ...
from .sdk import ContextInfra ...

...
```

so you could do something like

```
from data_safe_haven.infrastructure import ContextInfra

context_infra = ContextInfra(context)
context_infra.create()
```

```
from data_safe_haven.infrastructure import SHMStackManager

stack = SHMStackManager(context, config, pulumi_project)
stack.deploy()
```

and we could possibly unify the names/function names at the same time.

JimMadge:
I think I will stop here.

Changing `AzureSerialisableConfig.upload` or breaking down `types` doesn't seem worth it right now.

I think it would be good to restructure the `commands` package, at least to be consistent with how it is laid out. But perhaps that can be part of #1836?

jemrobinson:
Yes, let's do `commands` separately. I feel like I prefer `dsh shm <verb>` over `dsh <verb> shm` but happy to be convinced otherwise.

JimMadge:
Yes, I prefer <object> <verb> too :100:.

JimMadge (approved):

PR #1854, opened 2024-05-03
Move encrypted_key to DSHPulumiConfig

### LINKED ISSUE #1849: Cannot deploy SRE

<!--
Before reporting a problem please check the following. Replace the empty checkboxes [ ] below with checked ones [x] accordingly.
-->

- **Operating System:** macOS
- **Data Safe Haven version:** develop @ c1d0adc

<!--
What happens now and (if relevant) an explanation of why this is incorrect

Optionally include specific tasks (if any) in the order in which they need to be done.
- [ ] Task 1
- [ ] Task 2
- [ ] Task 3

Include a clear and concise description of what the behaviour should be when this issue is closed if this is not obvious.
-->
I cannot deploy an SRE (after successfully deploying an SHM) as the encryption key is not copied from the SHM project into the SRE project.

Possibly encryption key should be a top-level property of the DSHPulumiConfig instead of belonging to each DSHPulumiProject? Alternatively, each DSHPulumiProject should generate its own encryption key

<details>

```none
2024-05-02 12:11:02 [   ERROR] Could not deploy Secure Research Environment apple.                                                                                                                                                                                   cli.py:99
2024-05-02 12:11:02 [   ERROR] Pulumi deployment failed.                                                                                                                                                                                                             cli.py:99
2024-05-02 12:11:02 [   ERROR] Applying Pulumi configuration options failed.                                                                                                                                                                                         cli.py:99
2024-05-02 12:11:02 [   ERROR] 1 validation error for DSHPulumiProject                                                                                                                                                                                               cli.py:99
2024-05-02 12:11:02 [   ERROR] encrypted_key                                                                                                                                                                                                                         cli.py:99
2024-05-02 12:11:02 [   ERROR]   Input should be a valid string                                                                                                                                                                                                      cli.py:99
2024-05-02 12:11:02 [   ERROR]     For further information visit https://errors.pydantic.dev/2.7/v/string_type.
```

</details>

## :recycle: To reproduce

<!--
How can the current behaviour be reproduced? Include some steps to reproduce the problem.

Example:
- Deploy an SRE with option X set to Y
- Attempt to perform action Z
- Observe the previously mentioned failure/error message
-->
- Deploy an SHM
- Attempt to deploy an SRE

JimMadge:

- Move `encrypted_key` from `DSHPulumiProject` to `DSHPulumiConfig` as it is shared across projects.
- Move `DSHPulumiProject` fetching into `ProjectManager`, avoiding duplicate code in Typer commands.

Closes #1849


data_safe_haven/infrastructure/project_manager.py
```diff
@@ -365,23 +375,33 @@ def update_dsh_pulumi_project(self) -> None:
...
-        ).encrypted_key
         self.pulumi_project.stack_config = all_config_dict
 
+    def update_dsh_pulumi_config(self) -> None:
+        """Update persistent data in the DSHPulumiProject object"""
+        if self.pulumi_config.encrypted_key is None:
+            self.pulumi_config.encrypted_key = self.stack.workspace.stack_settings(
+                stack_name=self.stack_name
+            ).encrypted_key
+
```
jemrobinson: Should this check if the stack encrypted key is different from the one in the Pulumi config? We expect them to be the same, but should we raise an Exception if they're not?

data_safe_haven/infrastructure/project_manager.py
```diff
@@ -105,10 +107,14 @@ def project_settings(self) -> automation.ProjectSettings:
...
             config=self.pulumi_project.stack_config,
-            encrypted_key=self.pulumi_project.encrypted_key,
+            encrypted_key=self.pulumi_config.encrypted_key,
             secrets_provider=self.context.pulumi_secrets_provider_url,
         )
 
+    @property
+    def pulumi_project(self) -> DSHPulumiProject:
+        return self.pulumi_config[self.pulumi_project_name]
+
```
jemrobinson: Would it be better to do the `create_if_new` here, so we're effectively lazy-loading the project?

data_safe_haven/commands/deploy.py
```diff
@@ -46,7 +47,12 @@ def shm(
         verification_record = graph_api.add_custom_domain(config.shm.fqdn)
 
         # Initialise Pulumi stack
-        stack = SHMProjectManager(context, config, pulumi_project)
+        stack = SHMProjectManager(
+            context=context,
+            config=config,
+            pulumi_config=pulumi_config,
+            shm_name=context.shm_name,
```
jemrobinson: Couldn't `SHMProjectManager` load `shm_name` from `context` itself?

data_safe_haven/config/pulumi.py
```diff
@@ -56,7 +57,6 @@ def project_names(self) -> list[str]:
         """Produce a list of known DSH Pulumi Project names"""
         return list(self.projects.keys())
 
-    def create_or_select_project(self, project_name: str) -> DSHPulumiProject:
+    def create_if_new(self, project_name: str) -> None:
         if project_name not in self.project_names:
-            self[project_name] = DSHPulumiProject(encrypted_key="", stack_config={})
-        return self[project_name]
+            self[project_name] = DSHPulumiProject(stack_config={})
```
jemrobinson: I think the previous `create_or_select` logic made more sense. Let the DSHPulumiConfig class deal with keeping track of whether the project already exists or not, and let the client classes just get a `DSHPulumiProject` when they want one.

jemrobinson (changes requested):
A couple of structure thoughts.

Do you think it would be worth running a full deployment? If it helps the "how to deploy" docs should be ~done, so you could work through those?

data_safe_haven/infrastructure/project_manager.py
JimMadge: Could do, that should never happen though.

data_safe_haven/infrastructure/project_manager.py
JimMadge: The trouble with that is, it is only when deploying that we would want to create a `DSHPulumiProject`. In all other cases we expect it to exist and should raise an exception if it doesn't.

data_safe_haven/config/pulumi.py
JimMadge: This is more or less the same, it is just there was no place left where the result was assigned to a name, only `_ = DSHPulumiConfig.create_or_select(...)`.

data_safe_haven/infrastructure/project_manager.py
jemrobinson: "Should never happen" feels like a great place to raise an exception tbh :)

data_safe_haven/infrastructure/project_manager.py
jemrobinson: OK, but are we relying on a default KeyError exception if it's not in the dictionary? Maybe having a `get_project` method or similar on DSHPulumiConfig would help us to raise a more understandable exception?

JimMadge: I was thinking over lunch. I think I can get this and the `create_or_select` both working together in a better way.

jemrobinson: Yes, I think this can all live in one function

JimMadge: Fixed in 2d186463

data_safe_haven/config/pulumi.py
JimMadge: Fixed in 2d186463

data_safe_haven/commands/deploy.py
JimMadge: Fixed in fc18093d

data_safe_haven/infrastructure/project_manager.py
JimMadge: Fixed in b8ac270c

data_safe_haven/commands/teardown.py
```diff
@@ -24,12 +24,15 @@ def shm() -> None:
...
 
     try:
         # Remove infrastructure deployed with Pulumi
         try:
-            stack = SHMProjectManager(context, config, pulumi_project)
+            stack = SHMProjectManager(
+                context=context,
+                config=config,
+                pulumi_config=pulumi_config,
+            )
```
jemrobinson, line 35: Do we need to be passing `pulumi_project_name` here?

data_safe_haven/commands/teardown.py
```diff
@@ -68,10 +70,10 @@ def sre(
...
             stack = SREProjectManager(
-                context,
-                config,
-                pulumi_project,
-                sre_name,
+                context=context,
+                config=config,
+                pulumi_config=pulumi_config,
+                sre_name=sre_name,
                 graph_api_token=graph_api.token,
```
jemrobinson, line 77: Do we need to be passing `pulumi_project_name` here?

data_safe_haven/infrastructure/project_manager.py
```diff
@@ -105,10 +113,23 @@ def project_settings(self) -> automation.ProjectSettings:
...
+    def pulumi_project(self) -> DSHPulumiProject:
+        if not self._pulumi_project:
+            # Create DSH Pulumi Project if it does not exist, otherwise use existing
+            if self.create_project:
+                self._pulumi_project = self.pulumi_config.create_or_select_project(
+                    self.pulumi_project_name
+                )
+            else:
+                self._pulumi_project = self.pulumi_config[self.pulumi_project_name]
+
```
jemrobinson: We should catch the KeyError if `self.pulumi_project_name` is not an existing project an raise an informative exception.

jemrobinson (changes requested):

data_safe_haven/commands/teardown.py
JimMadge: No, because it is `context.shm_name`.

data_safe_haven/commands/teardown.py
JimMadge: It is the same as `sre_name`.

data_safe_haven/infrastructure/project_manager.py
JimMadge: I thought this would handle that

https://github.com/alan-turing-institute/data-safe-haven/blob/063131becebb476c17bc08ecb08e05c32c620663/data_safe_haven/config/pulumi.py#L17-L26

Could we do something better here?
Like `"No SHM/SRE called {self.pulumi_project_name} is defined"`?

data_safe_haven/commands/teardown.py
jemrobinson: Haha, this is the thing I already pointed out in my last review right? Whoops.

data_safe_haven/infrastructure/project_manager.py
jemrobinson: Sorry - didn't notice that you'd overwritten `__getitem__`. In which case, yes, this should catch it.

jemrobinson: However, I think it's good practice to catch all exceptions that we think might happen and reraise them as some kind of `DataSafeHavenError` so that we can do an overall `try: ... catch DataSafeHavenError as exc: ...` in the main cli loop.

jemrobinson:
```suggestion
            else:
                try:
                    self._pulumi_project = self.pulumi_config[self.pulumi_project_name]
                except (KeyError, TypeError) as exc:
                    msg = "No SHM/SRE named {self.pulumi_project_name} is defined."
                    raise DataSafeHavenConfigError(msg) from exc
```

data_safe_haven/infrastructure/project_manager.py
```diff
@@ -105,10 +117,26 @@ def project_settings(self) -> automation.ProjectSettings:
...
+            # Create DSH Pulumi Project if it does not exist, otherwise use existing
+            if self.create_project:
+                self._pulumi_project = self.pulumi_config.create_or_select_project(
+                    self.pulumi_project_name
+                )
+            else:
+                try:
+                    self._pulumi_project = self.pulumi_config[self.pulumi_project_name]
+                except (KeyError, TypeError) as exc:
+                    msg = f"No SHM/SRE named {self.pulumi_project_name} is defined."
```
jemrobinson:
```suggestion
                    msg = f"No SHM/SRE named {self.pulumi_project_name} is defined.\n{exc}"
```

jemrobinson (approved):
LGTM. Merge whenever you're ready @JimMadge

PR #1853, opened 2024-05-02
Break circular dependency in Context

jemrobinson:
## :white_check_mark: Checklist

- [x] You have given your pull request a meaningful title (_e.g._ `Enable foobar integration` rather than `515 foobar`).
- [x] You are targeting the appropriate branch. If you're not certain which one this is, it should be **`develop`**.
- [x] Your branch is up-to-date with the **target branch** (it probably was when you started, but it may have changed since then).
- [x] You have marked this pull request as a **draft** and added `'[WIP]'` to the title if needed (if you're not yet ready to merge).

### :arrow_heading_up: Summary

Add ContextBase class to break circular dependency

### :closed_umbrella: Related issues

Closes #1852

### :microscope: Tests

Tested locally

PR #1855, opened 2024-05-03
Add test for help function

jemrobinson:

Addresses point from #1853 review

### :microscope: Tests

Tested locally

### DIFF
--- a/data_safe_haven/cli.py
+++ b/data_safe_haven/cli.py
@@ -17,6 +17,18 @@ from data_safe_haven.exceptions import DataSafeHavenError
 from data_safe_haven.utility import LoggingSingleton
 
 
+# Create the application
+application = typer.Typer(
+    context_settings={"help_option_names": ["-h", "--help"]},
+    invoke_without_command=True,
+    name="dsh",
+    no_args_is_help=True,
+)
+
+
+# Custom application callback
+# This is executed before
+@application.callback()
 def callback(
     output: Annotated[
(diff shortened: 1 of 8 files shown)


data_safe_haven/cli.py
```diff
@@ -90,9 +91,13 @@ def main() -> None:
...
 
-    # Start the application
+    return application
+
+
+def main() -> None:
+    """Run the application and log any exceptions"""
     try:
-        application()
+        application()()
```
JimMadge: Is there a reason to have this as a factory rather than a function?

JimMadge (changes requested):

data_safe_haven/cli.py
jemrobinson: If you want to make this change then you'll need to change the return type of `application` from `typer.Typer` to whatever a `typer.Typer()` is.

JimMadge (changes requested):
In general, I don't like how `cli.py` is structured. It isn't very idiomatic typer.

I think it would be much clearer and easier to use if it followed the normal model of a `typer.Typer` object and `typer.Typer.command` decorators.

jemrobinson:
> In general, I don't like how `cli.py` is structured. It isn't very idiomatic typer.
> 
> I think it would be much clearer and easier to use if it followed the normal model of a `typer.Typer` object and `typer.Typer.command` decorators.

Are you happy to make that change on this PR?

JimMadge (approved):

PR #1857, opened 2024-05-03
Fix to test setup

jemrobinson:

Fix hatch test:test which was not working locally

n/a

### :microscope: Tests

n/a

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -113,7 +113,7 @@ dependencies = [
 pre-install-commands = ["pip install -r requirements.txt"]
 
 [tool.hatch.envs.test.scripts]
-test = "pytest {args:-vvv tests}"
+test = "python -m pytest {args:-vvv tests}"
 
 [tool.black]
 target-version = ["py312"]


JimMadge (commented):
I can't reproduce this.

I wonder if something was wrong with your virtual env as `pytest` should be installed in your `PATH`.

JimMadge:
@craddm Can you reproduce this?

jemrobinson:
I can only reproduce this in one environment and I'm not sure how it got into the state it's in. However, I think `python -m pytest` is never **wrong** and should be guaranteed to give you the right `pytest` anyway.

craddm:
Similarly, I can't reproduce this locally, but happy with what should be a more robust way of running it.

JimMadge:
Feels odd to introduce a fix for a broken virtual env. It is almost a feature to point out that something is wrong.

I'm not sure if it is most robust. If `pytest` fails your `$PATH` is likely wrong, if `python -m pytest` fails your `PYTHON_PATH` is likely wrong 🤔.

jemrobinson:
Closing as unnecessary.

PR #1870, opened 2024-05-09
Restructure commands

### LINKED ISSUE #1640: Pulumi: Move command line arguments to configuration

JimMadge:

Closes #1640 #1836


JimMadge:
@jemrobinson @craddm I'm going to go ahead with rearranging the commands to be `dsh <noun> <verb>` here like `dsh shm deploy`.

Would we prefer to have each command group as a single module (_e.g._ `shm.py`, `config.py`) or as a package (_e.g. `shm/deploy.py`, `shm/teardown.py`, `config/show.py`)?

jemrobinson:
Mild preference for having these as subpackages as it makes it easier to find e.g. `dsh shm deploy` if you know it's in `commands > shm > deploy.py`.

jemrobinson:
NB. Should this be targeting `develop`?

JimMadge:
> Mild preference for having these as subpackages as it makes it easier to find e.g. `dsh shm deploy` if you know it's in `commands > shm > deploy.py`.

@jemrobinson 
Doing this is a little awkward,

```python
# __init__.py
import typer

command_group = typer.Typer()
```

```python
# command.py

@command_group.command()
def command(...):
    ...
```

`command.py` is never imported, so the command is not register.

You could add unused imports to the `__init__.py`.
That might be the best solution, otherwise you have to do something unidiomatic like

```python
command_group.command(command)
```

or contrived like 
```python
# __init__.py
from .command import actual_implementation

command_group = typer.Typer()

@command_group.command()
def command(...):
    actual_implementation(...)
```

jemrobinson:
I think the last one (your "contrived" example) might be the best pattern. It's what was previously done for e.g. the `admin` command group and allows you to separate the typer command structure (all kept in one short file) from the implementation (separated from typer boilerplate).

For example:

```python
# cli.py
from .admin import admin_command_group

application.add_typer(
   admin_command_group,
   ....
)
```

```python
# admin/__init__.py
from .add_users import add_users as impl_add_users

admin_command_group = typer.Typer()

@admin_command_group.command()
def add_users(...):
    impl_add_users(...)
```

```python
# admin/add_users.py

def add_users(...):
   ....
```

JimMadge:
I hated that structure, it was almost always needless boilerplate.

```python

def f(x):
    g(x)

def g(x):
    ...
```

Where `g` is never called anywhere else.

jemrobinson:
Yeah, it's not the cleanest structure. Can you explain your preferred option again?

JimMadge:
Like this,

https://github.com/alan-turing-institute/data-safe-haven/blob/a8746eef90a36337147b895986a047ff3da75972/data_safe_haven/commands/context/__init__.py
https://github.com/alan-turing-institute/data-safe-haven/blob/a8746eef90a36337147b895986a047ff3da75972/data_safe_haven/commands/context/add.py

jemrobinson:
I find the `from . import update` syntax a bit unsettling. I'd probably prefer to have everything in the `__init__.py` than that (which I think was your original suggestion).

JimMadge:
@jemrobinson That would be much more in keeping with the Typer docs.

Do you still prefer packages, like

```
commands
├── sre
│   └── __init__.py
└── shm
    └── __init__.py
```

Seems needlessly complicated over

```
commands
├── sre.py
└── shm.py
```

jemrobinson:
No, in that case `sre.py` and `shm.py` are fine. I would prefer if the content of the command functions themselves was as simple as possible so that these files don't end up too large.

data_safe_haven/commands/admin.py
```diff
@@ -5,11 +5,13 @@
...
-from .admin_unregister_users import admin_unregister_users
+from data_safe_haven.administration.users import UserHandler
+from data_safe_haven.config import Config, DSHPulumiConfig
+from data_safe_haven.context import ContextSettings
+from data_safe_haven.exceptions import DataSafeHavenError
+from data_safe_haven.external import GraphApi
+from data_safe_haven.functions import sanitise_sre_name
+from data_safe_haven.utility import LoggingSingleton
 
 admin_command_group = typer.Typer()
```
jemrobinson: All the `admin` commands are to do with user management. Do we envisage future commands that do other things or should we simplify e.g. `dsh admin list-users` to `dsh users list`?

data_safe_haven/commands/admin.py
```diff
@@ -23,12 +25,54 @@ def add_users(
...
+
+        # Add users to SHM
+        users = UserHandler(context, config, pulumi_config, graph_api)
+        users.add(csv)
+    except DataSafeHavenError as exc:
+        msg = f"Could not add users to Data Safe Haven '{shm_name}'.\n{exc}"
+        raise DataSafeHavenError(msg) from exc
 
 
 @admin_command_group.command(help="List users from a deployed Data Safe Haven.")
```
jemrobinson: I think you might not need the help string as a decorator argument if you have it as a docstring. Worth checking?

data_safe_haven/commands/admin.py
```diff
@@ -50,7 +94,46 @@ def register_users(
...
 ) -> None:
-    admin_register_users(usernames, sre)
+    """Register existing users with a deployed SRE"""
+    context = ContextSettings.from_file().assert_context()
+    config = Config.from_remote(context)
+    pulumi_config = DSHPulumiConfig.from_remote(context)
+
+    shm_name = context.shm_name
+    # Use a JSON-safe SRE name
+    sre_name = sanitise_sre_name(sre)
```
jemrobinson: Should we be verifying the sre_name instead of just sanitising it?

data_safe_haven/commands/admin.py
```diff
@@ -66,7 +149,27 @@ def remove_users(
         ),
     ],
 ) -> None:
-    admin_remove_users(usernames)
+    """Remove existing users from a deployed Data Safe Haven"""
+    context = ContextSettings.from_file().assert_context()
+    config = Config.from_remote(context)
+    pulumi_config = DSHPulumiConfig.from_remote(context)
+
+    shm_name = context.shm_name
```
jemrobinson: Is `shm_name` used?

data_safe_haven/commands/admin.py
```diff
@@ -86,4 +189,47 @@ def unregister_users(
...
     ],
 ) -> None:
-    admin_unregister_users(usernames, sre)
+    """Unregister existing users from a deployed SRE"""
+    context = ContextSettings.from_file().assert_context()
+    config = Config.from_remote(context)
+    pulumi_config = DSHPulumiConfig.from_remote(context)
+
+    shm_name = context.shm_name
+    sre_name = sanitise_sre_name(sre)
```
jemrobinson: As above: should we verify this?

data_safe_haven/commands/sre.py
```diff
@@ -1,97 +1,21 @@
-"""Command-line application for deploying a Data Safe Haven component, delegating the details to a subcommand"""
-
 from typing import Annotated, Optional
```
jemrobinson, line 3: Can we restore a module-level docstring?

data_safe_haven/commands/shm.py
```diff
@@ -0,0 +1,116 @@
+from typing import Annotated, Optional
```
jemrobinson, line 3: Can we restore a module-level docstring?

jemrobinson (changes requested):
Looks good - a couple of minor changes.

data_safe_haven/commands/admin.py
JimMadge: Let's move that to `users`  then. In the future there might be `dsh admin users ...` and `dsh admin something-else ...`

JimMadge: Done in 96f91b76

data_safe_haven/commands/admin.py
JimMadge: Fixed in f35a3b70

data_safe_haven/commands/admin.py
JimMadge: Yes, I think they all are.

data_safe_haven/commands/sre.py
JimMadge: Done in aeb6e6fa

data_safe_haven/commands/admin.py
JimMadge: That happens later, in the project manager. I think it is better to do that in one place than having to add it to every command.

jemrobinson: That's true for some commands but for this command there's no check before we run `users.register(sre_name, usernames_to_register)`.

data_safe_haven/commands/admin.py
jemrobinson: I don't think it's used anywhere in the remainder of this function.

jemrobinson (approved):
One outstanding question about whether we should verify sre_name but that's probably out-of-scope for this PR.

data_safe_haven/commands/admin.py
JimMadge: Good point, I've opened an issue https://github.com/alan-turing-institute/data-safe-haven/issues/1879

PR #1873, opened 2024-05-09
Add comments/docstrings to exceptions

### LINKED ISSUE #1518: Add descriptions/guidance for custom exceptions
We have defined a number of custom exceptions to help with debugging and aid users.

https://github.com/alan-turing-institute/data-safe-haven/blob/f0a890632725e2b6883c8997787294b824caa1cb/data_safe_haven/exceptions/__init__.py#L1-L14

Currently only the names indicate when they should be used.
It would be good to add some documentation (at least comments/docstrings) to explain when the exceptions are to be used (and when they shouldn't be used).

craddm:

Adds descriptive comments and/or docstrings to exceptions.

Also checks consistency of usage across the codebase.

Closes #1518


data_safe_haven/exceptions/__init__.py
```diff
@@ -9,7 +9,8 @@ class DataSafeHavenCloudError(DataSafeHavenError):
 class DataSafeHavenConfigError(DataSafeHavenError):
     pass
 
-
+# This is a custom exception class for handling errors when interacting with Entra ID.
+# For example, when adding users to an Entra group fails.
 class DataSafeHavenEntraIDError(DataSafeHavenCloudError):
     pass
```
jemrobinson:
```suggestion
class DataSafeHavenEntraIDError(DataSafeHavenCloudError):
    """
    This is a custom exception class for handling errors when interacting with Entra ID.
    For example, when adding users to an Entra group fails.
    """
    pass
```

If you make this a docstring instead of a comment then it will show up on e.g. tooltips in an IDE.

JimMadge: And in API documentation if we ever do that.

craddm:
Currently the custom exceptions use the following hierarchy: 
image

craddm:
I'm still trying to understand what we want the errors to achieve. At the moment they're a bit of a mishmash. 

For example, `DataSafeHavenParameterError` is used when a user supplies a key for a `context` that either does not exist or already exists (when trying to create), but also when a yaml config file cannot be validated.

On the other hand, we have `DataSafeHavenConfigError`, which is used when no context file exists, or when a user supplies an SRE name that does not exist in the config, but also when no context is selected, or when the config file is not correct yaml.

Then we have `DataSafeHavenInputError` which is used when a subscription cannot be found, when there are missing values for users, when a secret already exists in a key-vault...

Finding it difficult to nail a description for some of these error classes as it's not always clear what they are intended to capture

jemrobinson:
It's totally fine for you to redesign the error names and/or inheritance structure to make more sense. I think we want things to adhere to the following guidelines:

- everything should inherit from `DataSafeHavenError` eventually
- if `X` and `Y` both inherit from `A` it should be because they're conceptually both a type of `A` but that it's worth distinguishing between them (e.g. it might be nice to distinguish Azure errors from Graph errors even though they're both some kind of Microsoft/Cloud error).
- we (ideally) should be using each type of error in more than one place and, if possible, more than one file

Anything else you'd add to this @JimMadge ?

jemrobinson:
@craddm is this still being worked on?

craddm:
Yes - sorry, it's been bumped down the priority list a few times. I'll get back to it shortly.

jemrobinson:
No problem - just checking that you weren't waiting for review.

jemrobinson:
Um, what happened here @craddm ?

craddm:
I tried to update it inline with develop, and apparently it went horribly wrong

JimMadge:
@craddm I think 65b8888 fixes CI.

The issue was making `DataSafeHavenAzureAPIAuthenticationError` a child of `DataSafeHavenAzureError`. We want to distinguish between those so that we can catch API authentication problems and ask the user to run `az login`.

JimMadge:
@craddm Is this ready for review?

craddm:
Yes, I've decided I'm happy leaving `DSHUserHandlingErrors` alone for the time being, and happy to discuss whether to drop `DSHType/ValueErrors` here

data_safe_haven/exceptions/__init__.py
```diff
@@ -12,61 +18,106 @@ def __init__(self, message: str | bytes):
...
-class DataSafeHavenConfigError(DataSafeHavenError):
     pass
 
 
-class DataSafeHavenEntraIDError(DataSafeHavenCloudError):
+class DataSafeHavenAzureAPIAuthenticationError(DataSafeHavenError):
+    """
+    Exception class for handling errors when authenticating against the Azure API.
+
+    Used to capture exceptions generated when the user is not authenticated or authentication has expired
```
JimMadge:
```suggestion
    Used to capture exceptions generated when the user is not authenticated or authentication has expired
```
```suggestion
    Used to capture exceptions generated when the user is not authenticated or authentication has expired.
```

data_safe_haven/exceptions/__init__.py
```diff
@@ -12,61 +18,106 @@ def __init__(self, message: str | bytes):
...
+    """Exception raised when it is not possible to generate a valid IPv4 range."""
+
     pass
 
 
-class DataSafeHavenNotImplementedError(DataSafeHavenInternalError):
+class DataSafeHavenMicrosoftGraphError(DataSafeHavenAzureError):
+    """
+    Exception class for handling errors when interacting with the Microsoft Graph API.
+
```
JimMadge:
```suggestion
```

data_safe_haven/exceptions/__init__.py
```diff
@@ -12,61 +18,106 @@ def __init__(self, message: str | bytes):
...
     pass
 
 
-class DataSafeHavenParameterError(DataSafeHavenError):
+class DataSafeHavenPulumiError(DataSafeHavenError):
+    """
+    Exception class for handling errors when interacting with Pulumi.
+
+    For example, when a Pulumi operation such as a deployment fails
```
JimMadge:
```suggestion
    For example, when a Pulumi operation such as a deployment fails.
```

data_safe_haven/exceptions/__init__.py
```diff
@@ -12,61 +18,106 @@ def __init__(self, message: str | bytes):
...
+
     pass
 
 
 class DataSafeHavenSSLError(DataSafeHavenError):
-    pass
+    """
+    Exception class for handling errors related to administration of SSL certificates.
 
+    For example, errors refreshing or creating SSL certificates
```
JimMadge:
```suggestion
    For example, errors refreshing or creating SSL certificates.
```

data_safe_haven/exceptions/__init__.py
```diff
@@ -12,61 +18,106 @@ def __init__(self, message: str | bytes):
...
     pass
 
 
-class DataSafeHavenAzureAPIError(DataSafeHavenError):
-    pass
+class DataSafeHavenTypeError(DataSafeHavenError):
+    """
+    Exception class for handling errors related to type checking.
 
+    For example, when a function is called with an argument of the wrong type
```
JimMadge:
```suggestion
    For example, when a function is called with an argument of the wrong type.
```

data_safe_haven/exceptions/__init__.py
```diff
@@ -12,61 +18,106 @@ def __init__(self, message: str | bytes):
...
     pass
 
 
-class DataSafeHavenUserHandlingError(DataSafeHavenInternalError):
-    pass
+class DataSafeHavenUserHandlingError(DataSafeHavenError):
+    """
+    Exception class for handling errors related to user handling.
 
+    For example, when listing or registering users fails
```
JimMadge:
```suggestion
    For example, when listing or registering users fails.
```

data_safe_haven/exceptions/__init__.py
```diff
@@ -12,61 +18,106 @@ def __init__(self, message: str | bytes):
...
-class DataSafeHavenMicrosoftGraphError(DataSafeHavenAzureError):
     pass
 
 
-class DataSafeHavenPulumiError(DataSafeHavenCloudError):
+class DataSafeHavenValueError(DataSafeHavenError):
+    """
+    Exception class for handling errors related to value checking.
+
+    For example, when a function is called with an argument of the wrong value
```
JimMadge:
```suggestion
    For example, when a function is called with an argument of the wrong value.
```

JimMadge (changes requested):
Looks good.

I'm happy to keep `DataSafeHaven{Value,Type}Error`. Do we catch them anywhere?
If we don't I can see that we might want to in the future.

Looking at commands files, I think there are places where we need to replace `raise DataSafeHaven*` with

```
logger.critical(message)
Typer.Exit(1)
```

@craddm could you open an issue for that unless you want to fix it here?

JimMadge:
@craddm Are you looking to fix all of the exceptions in commands here?

craddm:
Possibly - am I on the right track?

data_safe_haven/commands/shm.py
```diff
@@ -114,4 +116,5 @@ def teardown() -> None:
         pulumi_config.upload(context)
     except DataSafeHavenError as exc:
         msg = "Could not teardown Safe Haven Management environment."
-        raise DataSafeHavenError(msg) from exc
+        logger.critical(msg)
```
JimMadge: Could probably put the little string straight in the `logger.critical()` call here.

The reason we often have

```python
msg = "..."
raise Exception(message)
```

is for linting.

JimMadge:
> Possibly - am I on the right track?

Yes, you just need to do something similar for every `@typer.command()`.

craddm:
I think I've got all the relevant ones now. There are some I've left alone as I think we don't want to exit at the point they occur, but could be wrong.

craddm:
One thing I was unsure about was when I needed to decode `response.content` to print it as part of the error messages, done in 5598759b8b5724b006fb2e760afab5586302bf56

data_safe_haven/external/api/graph_api.py
```diff
@@ -770,18 +769,17 @@ def http_delete(self, url: str, **kwargs: Any) -> requests.Response:
...
-        except Exception as exc:
+        except requests.exceptions.RequestException as exc:
             msg = f"Could not execute DELETE request to '{url}'."
             raise DataSafeHavenMicrosoftGraphError(msg) from exc
 
+        # We do not use response.ok as this allows 3xx codes
+        if requests.codes.OK <= response.status_code < requests.codes.MULTIPLE_CHOICES:
+            return response
+        else:
+            msg = f"Could not execute DELETE request to '{url}'. Response content received: '{response.content.decode()}'."
```
JimMadge, line 780:
```suggestion
            msg = f"Could not execute DELETE request to '{url}'. Response content received: '{response.text}'."
```

I think this is because `response.content` is a `bytes` object, but the exception expects a `str`. I think `Response.text` is a `str`.
If this works it could be applied across all the places.

JimMadge (changes requested):
@craddm Are there any more `raise`s in `data_safe_haven/commands/*.py` functions? I think all of those should be replaces with a message and `Typer.Exit` to gracefully end the program.

craddm:
> @craddm Are there any more `raise`s in `data_safe_haven/commands/*.py` functions? I think all of those should be replaces with a message and `Typer.Exit` to gracefully end the program.

There are some nested `try` blocks in `shm.py`:

https://github.com/craddm/data-safe-haven/blob/43a9698716d7e77a6badd32b6c834a02ce1e011c/data_safe_haven/commands/shm.py#L91-L119

I interpret this as the inner exception being caught by the outer exception, and thus there's no need to `typer.Exit` the inner exception. Or perhaps the inner exception is redundant?

JimMadge:
@craddm Yes those are OK because we catch them can raise `typer.Exit()`.
There were a couple of others that would terminate the program ungracefully though.
I think 8caebec should fix that.

JimMadge (approved):

JimMadge:
@craddm Ready to merge?

PR #1881, opened 2024-05-10
Protect against configuration changes

### LINKED ISSUE #1714: Pulumi: Context with multiple administrators

- **Operating System:** macOS 13.6.3
- **Data Safe Haven version:** `python-migration`

Previously, config files always lived in Azure storage and could only be interacted with programmatically. This meant that multiple administrators attempting to make a change (e.g. adding a new SRE) were always in sync with existing changes.

The new setup involves editing a local file and then uploading it manually. This introduces the possibility of something like the following:

AdminA: Deploys SHM (local file has SHM)
AdminA: Deploys SRE Alpha (local file has SHM and SRE Alpha)
AdminB: Downloads config file, deploys SRE Beta (local file has SHM, SRE Alpha and SRE Beta)
AdminA: Does not download updated config file but deploys SRE Gamma (local file has SHM, SRE Alpha and SRE Gamma)

This means that SRE Beta will be removed from the config version stored in Azure and it's also quite likely that neither AdminA nor AdminB will notice that this has happened.

## :recycle: To reproduce

Do the above

Some possibilities include:

1. switch back to requiring users to provide input at the command line rather than by editing YAML themselves
2. switch from the current method of directly editing the master config file to instead creating a new file (for just the SHM or just a new SRE) that the code will then merge into the canonical file that lives in Azure storage

I personally think there could be some useful mileage in (2) - it would simplify the thing we're asking users to do ("just write a config section for your SRE") and also remove the possibility that they might delete something important like the pulumi stack information.

JimMadge:

Add protection for configuration changes,

- Add method to print diff in YAML for `YamlSerialisableModel`
- Add method to print diff for local/remote for `AzureSerialisableModel`
- Improve upload commands to prompt user before making changes to config
  - displays diff to user, asks for y/n
- Tests for methods and commands

Closes #1714


tests/commands/test_config.py
```diff
@@ -25,22 +25,109 @@ def test_template_file(self, runner, tmp_path):
...
+    #     )
+    #     assert result.exit_code == 0
+
+    #     mock_method.assert_called_once_with(
+    #         config_yaml,
+    #         Config.filename,
+    #         context.resource_group_name,
+    #         context.storage_account_name,
+    #         context.storage_container_name,
+    #     )
```
jemrobinson: If we don't need this, can we delete it?

data_safe_haven/commands/config.py
```diff
@@ -34,9 +35,29 @@ def upload(
...
+
+    # Create configuration object from file
     with open(file) as config_file:
         config_yaml = config_file.read()
     config = Config.from_yaml(config_yaml)
+
+    # Present diff to user
+    if Config.remote_exists(context):
+        if diff := config.remote_yaml_diff(context):
+            print("".join(diff))
```
jemrobinson, line 48:
```suggestion
            for line in diff:
                logger.info(line)
```
I think it might be clearer to log each line separately

data_safe_haven/commands/config.py
```diff
@@ -34,9 +35,29 @@ def upload(
...
+            if not logger.confirm(
+                (
+                    "Configuration has changed, "
+                    "do you want to overwrite the remote configuration?"
+                ),
+                default_to_yes=False,
+            ):
+                raise typer.Exit()
+        else:
+            print("No changes, won't upload configuration.")
```
jemrobinson, line 58:
```suggestion
            logger.info("No changes, won't upload configuration.")
```

jemrobinson: I prefer logging over printing because:

- you can easily redirect it as needed
- you can enable/disable at different levels of info
- you can turn it off completely if desired

all of which is hard to do with `print` statements.

data_safe_haven/serialisers/yaml_serialisable_model.py
```diff
@@ -1,5 +1,6 @@
 """A pydantic BaseModel that can be serialised to and from YAML"""
 
+from difflib import unified_diff
```
jemrobinson, line 3: Is this part of the standard library or do we need to add a package?

data_safe_haven/serialisers/azure_serialisable_model.py
```diff
@@ -36,16 +36,31 @@ def from_remote_or_create(
         Construct an AzureSerialisableModel from a YAML file in Azure storage, or from
         default arguments if no such file exists.
         """
+        if cls.remote_exists(context):
+            return cls.from_remote(context)
+        else:
+            return cls(**default_args)
+
+    @classmethod
+    def remote_exists(cls: type[T], context: ContextBase) -> bool:
```
jemrobinson, line 45: Is this function used elsewhere?

jemrobinson (changes requested):
Looks good. A couple of questions.

data_safe_haven/serialisers/azure_serialisable_model.py
JimMadge: It is also used in the `config upload` command.

data_safe_haven/commands/config.py
JimMadge: I want to avoid using the logger here because,

- It makes the output very difficult to read
- It bypasses typer stdout so we can't use it in tests (it would break the tests here, and make them a bit pointless as they could only check the return code)

More generally, I think we are sometimes mixing up log information and stdout in commands.
Can we open an issue to sort this?

jemrobinson: Yes - please open an issue. I'd be interested to know what you mean by bypassing typer stdout - I thought typer didn't do its own I/O but delegated this to rich (which is also what the logger does).

In general, I think the only time we should be writing directly to stdout is if we're in a command that is intended to be used as part of a pipe/redirect.

JimMadge: > I'd be interested to know what you mean by bypassing typer stdout

I don't entirely understand. However, anything sent to the logger object isn't captured as stdout in tests. I don't know how to inspect any of that.

> In general, I think the only time we should be writing directly to stdout is if we're in a command that is intended to be used as part of a pipe/redirect.

I suppose that is the conversation we need to have, but I think that is incorrect and goes against convention.

I think of stdout as information for the user. Logs are for posterity, for the user/admin.
`pacman -Syu` for example uses stdout to show the user package being downloaded and installed, and hooks being run. You wouldn't pipe that to anything else. Logs of transactions are also written to logs in `/var/logs/pacman.log`.

Or, a more common/mundane example, you would rarely pipe `<command> --help` to another command, but it would be very odd to not send that to stdout.

data_safe_haven/serialisers/yaml_serialisable_model.py
JimMadge: stdlib

data_safe_haven/commands/config.py
JimMadge: #1891

craddm (approved):
LGTM

Issue #710, opened 2020-06-25
Support for Weka 3.8 or later


jack89roberts:
#641 added the version of [Weka](https://www.cs.waikato.ac.nz/ml/weka/) available with `apt` to the DSVM build image. This is currently Weka version 3.6.14, which is >4 years old. The latest stable version (3.8.4) and development version (3.9.4) are only available on [SourceForge](https://sourceforge.net/projects/weka/).

Below are relevant commands I played with to install and run Weka 3.8.4, should we need it in the future. Outstanding points are:
* How to correctly configure path, java classpath etc. to make it easy to use from the command-line?
* Is SourceForge an acceptable source for something being added to safe haven? They've cleaned up their image but have a sketchy past bundling junkware with software under previous ownership.

### Installation
```
apt-get install wget unzip
cd /tmp
wget https://prdownloads.sourceforge.net/weka/weka-3-8-4-azul-zulu-linux.zip
unzip weka-3-8-4-azul-zulu-linux.zip
rm weka-3-8-4-azul-zulu-linux.zip
```

### Start GUI
```
cd /tmp/weka-3-8-4
./weka.sh
```

### Run something from command line
```
cd /tmp/weka-3-8-4
java -classpath weka.jar weka.core.Version
```

### Install a package from downloaded zip
Assuming a model where people can request a package zip be ingressed to the environment if needed (rather than having access to a package mirror). Packages are here: https://weka.sourceforge.io/packageMetaData/
```
cd /tmp
wget http://prdownloads.sourceforge.net/weka/bayesianLogisticRegression1.0.5.zip
cd weka-3-8-4
java -classpath weka.jar weka.core.WekaPackageManager -install-package ../bayesianLogisticRegression1.0.5.zip
```
(or from GUI)

PR #1892, opened 2024-05-16
Add workspace packages

### LINKED ISSUE #1574: Minimal required workspace software

Currently the workspaces have very little installed: R, Python and libraries to interact with databases. What is a minimal useful workspace. Ideally small enough that we don't need to pre-build it (either with current scripts or bureau).

JimMadge:

- #1909 - Shouldn't dive too far into specifying packages/configuration until we have decided on the distro.
- #2034 

- Add ansible playbook for desired state
  - Install apt packages
  - Configures shell/prompts
- Remove FileUpload dynamic provider

I have moved some, but not all desired state here.
There were some files templated in cloud-init which need data from Pulumi/configuration.
We _can_ do that in Ansible, but we should think about how to we get the data to the workspaces.

- Template files locally and send them
- Pass variable to Ansible process?

There are some places where we could take advantage of the "official" Ansible community modules, or other packages from Ansible Galaxy. That will also require a bit of thought though as `ansible-galaxy` is really a package manager, so would be difficult to use in a locked down SRE.

#### Further Work

- Using Ansible galaxy collections/roles
- Testing (molecule?)
- Auditd configuration
- Clamav configuration
- Packages which require some manual work
  - RStudio
  - Nvidia drivers
  - Azure Data Studio and sqlcmd (possibly)
- Developer docs (how to change configuration and run using az CLI)
- Tightening networking rules
- Fix release lookup (and remove old releases)

Closes #1574

### :microscope: Tests

Tested in a new deployment,

- Desired state container mounted correctly
- Playbook runs
- Systemd module installed
- Systemd module works
- Systemd timer installed and active
- Manually checked modified files, systemd services, _etc._
- Manually checked requested packages were installed


JimMadge:
Looking at some options to get the files on workspace VMs. The constraints are,

- Needs to happen early after deployment
  - Will have Linux utils and some basics from cloud-init (curl, bash, sed, but not az)
- Should happen on all workspaces
  - Easier to avoid authentication if possible
- Data is non-sensitive
  - No need for authentication for read access

The good solutions I can see are,

1. Blob container with anonymous read access
  - Use the existing data_configuration storage account (is that appropriate @jemrobinson)
  - Could create a new storage account if we are worried about enabling anonymous access
  - Anonymous read access
  - Networking rules only allow access from inside SRE
  - Upload a manifest file (as you need list permissions to list all files)
  - Script to fetch manifest, then fetch all files in the manifest using curl
2. Blob container with sftp access
  - Storage account needs [hierarchical namespace](https://learn.microsoft.com/en-us/azure/storage/blobs/data-lake-storage-namespace)
  - Enable SFTP
  - Create local user with read/list permissions
    - With machine password for all workspaces or,
    - with ssh keypairs generated on each workspace
  - Script to scp from sftp using local account/password/pubkey

Non-options,

1. NFS: No easy way for us/admins to put files into NFS share
2. BlobFuse or SMB: more pre-requisites

Using anonymous access blob and http is nice for not needing authentication. However, you have to keep the manifest up to date which could be prone to error.
SFTP might be the best option, for the cost of storing the SFTP password and writing it to each workspace VM.

Thoughts @jemrobinson @craddm

jemrobinson:
Is the cloud-init Ansible support (https://cloudinit.readthedocs.io/en/latest/reference/modules.html#ansible) any help here?

Some thoughts on your suggestions:

1. `data_configuration` storage account
- Access is allowed from known `data_configuration_ip_addresses`
- Currently doesn't have any blob containers (only file shares)
- See e.g. the way we mount `ingress` and `egress` from blob storage over NFSv3

2. Enabling SFTP
- This is a bit of a mess, but @craddm got it working for a DSG in the past (hopefully notes on how to do it are somewhere)
- Not sure I would recommend it

How/when would the file get updated? Is it reasonable to imagine e.g. an Azure Function that would do the updating? If so, perhaps it could write directly to the VM and you wouldn't need a storage volume at all?

craddm:
I do have some notes on SFTP, yes. Wasn't enormously tricky, IIRC. Creating a local user was easy through the portal, and seems from Powershell it was easy to create one with a password rather than an SSH key.

I'm not seeing much about Azure Functions writing directly to VMs; you can trigger them on uploads to blob storage but so far I've only seen people suggest that the function copies files to a File Share

JimMadge:
@jemrobinson How much experience with key pairs in Azure/Pulumi do you have?

Looks like your can't generate a local user with password using Pulumi (at least, the password isn't an output).
You can specify an existing key pair though. That would mean we would need to be able to fetch the private key for each workspace.

jemrobinson:
What Pulumi resource are you using to generate a local user? I assume you don't mean in cloud-init?

JimMadge:
[azure_native.storage.LocalUser](https://www.pulumi.com/registry/packages/azure-native/api-docs/storage/localuser/)

Local user for SFTP, not for the workspaces.

JimMadge:
@jemrobinson Looks like Azure expects to keep the private key for keys in a keyvault.

I think the best option would be to use the [TLS package](https://www.pulumi.com/registry/packages/tls/) to generate a key, then store that as a secret.
(I'm now vaguely recalling using the TLS package before to do something like this).

jemrobinson:
I'm still a bit confused about what advantage we get for using SFTP over one of the other options. Is it just that we can do so earlier in the cloud-init order of operations? If so, why does this matter?

N.B. if you want to generate a Key in a Keyvault, I think you should probably use the [Key resource](https://www.pulumi.com/registry/packages/azure-native/api-docs/keyvault/key/)

JimMadge:
The SFTP way means,

- No requirements beyond what is already on a bare Ubuntu VM image
- No need to have a manifest (you can just `scp -r user@host:* /`)

The current downside is it means sharing a private key between workspaces (or creating a new local account with key pair for each workspace).

data_safe_haven/resources/workspace/workspace.cloud_init.mustache.yaml
```diff
@@ -48,10 +20,44 @@ write_files:
...
+
+  - path: "/etc/systemd/system/desired-state.timer"
+    permissions: "0644"
+    content: |
+      [Unit]
+      Description=Run desired state script daily
+
+      [Timer]
+      OnCalendar=daily
+      Persistent=true
```
jemrobinson, line 67: Is daily often enough? What time will this happen? It might be a while after first deployment.

JimMadge: Feels OK as a baseline. In the long term, I think we can trigger this manually.

I believe daily is midnight, plus whatever default random delay (we can specify that if we want to avoid running too much at once).

I was going to say the service is run at the end of cloud init. Then I realised I had forgotten to do that, so I've added it now 😅.

data_safe_haven/infrastructure/programs/sre/data.py
```diff
@@ -550,10 +550,14 @@ def __init__(
         ]
         # Upload file assets to desired state container
         for file in files_desired_state:
-            file_path = Path(file.path)
-            file_name = file_path.name
-            blob_name = file_path.relative_to(
-                (resources_path / "workspace" / "ansible").absolute()
+            file_path = Output.from_input(file).apply(lambda file: Path(file.path))
```
jemrobinson:
```suggestion
            file_path = Output.from_input(file).apply(Path)
```
might work here?

JimMadge:
## Problem

Blob storage accounts with NFSv3 enabled don't (seem to) use access tiers for blobs.
When uploading blobs

https://github.com/alan-turing-institute/data-safe-haven/blob/9ac1ee022f94a8ce6a147785f234f6321cb1e532/data_safe_haven/infrastructure/programs/sre/data.py#L554-L562

If [`access_tier`](https://www.pulumi.com/registry/packages/azure-native/api-docs/storage/blob/#inputs) is undefined, it defaults to `None`.
This causes an error `ValueError: '' is not a valid BlobAccessTier`.
Settings `access_tier` to any `storage.BlobAccessTier` enum value, or string causes an error

```
creating replacement (0s) ; error: updating access tier for blob
"hello.txt" (container "desiredstate" / account
"shdaisrehojdesiredstatec"): blobs.Client#SetTier: Failure responding to
request: StatusCode=400 -- Original Error: autorest/azure: Service
returned an error. Status=400 Code="InvalidQueryParameterValue"
Message="Value for one of the query parameters specified in the request
URI is
invalid.
```

Perhaps the Pulumi azure provider is always calling the Azure API method for setting access tier.
For this storage account/container, however, there is no valid value for access tier.

### Thoughts

This may be an azure-pulumi-native bug, or at least an edge case that isn't supported.
The Azure API, through az CLI, can handle uploading blobs to this kind of container. _I.e._ the setting of access tier isn't an inherent part of uploading a blob. We could create a custom Pulumi resource which uses the azure Python SDK to upload a blob.

JimMadge:
Importing a blob uploaded in the portal gives

```Python
requirementblob = azure_native.storage.Blob("requirementblob",
    access_tier=azure_native.storage.BlobAccessTier(""),
    account_name="shdaisrehojdesiredstatec",
    blob_name="requirements.txt",
    container_name="desiredstate",
    content_md5="b4s63roBt3c9gnki/yIG8A==",
    content_type="text/plain",
    metadata={},
    resource_group_name="shm-daimyo-sre-hojo-rg-data",
    type=azure_native.storage.BlobType.BLOCK,
    opts=pulumi.ResourceOptions(protect=True))
```

Using `access_tier=azure_native.storage.BlobAccessTier("")` gives the same error as when using `None` or providing no argument, `ValueError: '' is not a valid BlobAccessTier`.

data_safe_haven/infrastructure/programs/sre/data.py
```diff
@@ -456,6 +457,154 @@ def __init__(
...
+                ResourceOptions(parent=storage_account_data_desired_state),
+            ),
+        )
+        # Set storage container ACLs
+        BlobContainerAcl(
+            f"{container_desired_state._name}_acl",
+            BlobContainerAclProps(
+                acl_user="rwx",
+                acl_group="r-x",
+                acl_other="r-x",
```
jemrobinson: Why do we want these permissions?

data_safe_haven/infrastructure/programs/sre/networking.py
```diff
@@ -355,6 +355,18 @@ def __init__(
...
+                    description="Allow inbound connections from workspaces.",
+                    destination_address_prefix=SREIpRanges.data_configuration.prefix,
+                    destination_port_range="*",
+                    direction=network.SecurityRuleDirection.INBOUND,
+                    name="AllowWorkspacesInbound",
+                    priority=NetworkingPriorities.INTERNAL_SRE_WORKSPACES,
+                    protocol=network.SecurityRuleProtocol.ASTERISK,
+                    source_address_prefix=SREIpRanges.workspaces.prefix,
+                    source_port_range="*",
+                ),
```
jemrobinson, line 420: I think we don't want this. The configuration data is deliberately kept isolated from user space so that we can store secrets etc. there as required.

Since you've set up a new storage account container for desired state, you can attach the private endpoint to its own subnet.

data_safe_haven/infrastructure/programs/sre/networking.py
```diff
@@ -1282,6 +1294,18 @@ def __init__(
...
+                    description="Allow outbound connections to configuration data endpoints.",
+                    destination_address_prefix=SREIpRanges.data_configuration.prefix,
+                    destination_port_range="*",
+                    direction=network.SecurityRuleDirection.OUTBOUND,
+                    name="AllowDataConfigurationEndpointsOutbound",
+                    priority=NetworkingPriorities.INTERNAL_SRE_DATA_CONFIGURATION,
+                    protocol=network.SecurityRuleProtocol.ASTERISK,
+                    source_address_prefix=SREIpRanges.workspaces.prefix,
+                    source_port_range="*",
+                ),
```
jemrobinson: As above, this should be a link to a dedicated desired state endpoint.

data_safe_haven/infrastructure/programs/sre/workspaces.py
```diff
@@ -215,34 +186,13 @@ def __init__(
...
-                "ldap_user_filter": ldap_user_filter,
-                "ldap_user_search_base": ldap_user_search_base,
-                "software_repository_hostname": software_repository_hostname,
-                "storage_account_data_private_user_name": storage_account_data_private_user_name,
-                "storage_account_data_private_sensitive_name": storage_account_data_private_sensitive_name,
-            }
-            cloudinit = chevron.render(f_cloudinit, mustache_values)
-            return b64encode(cloudinit)
+            cloudinit = chevron.render(f_cloudinit, kwargs)
+            logger.debug(f"Generated cloud-init config:\n {cloudinit}")
```
jemrobinson: Are we worried about exposing secrets in the logs here?

data_safe_haven/infrastructure/programs/sre/workspaces.py
```diff
@@ -131,9 +132,13 @@ def __init__(
...
             ldap_user_search_base=props.ldap_user_search_base,
             software_repository_hostname=props.software_repository_hostname,
+            storage_account_data_desired_state_name=props.storage_account_data_desired_state_name,
             storage_account_data_private_user_name=props.storage_account_data_private_user_name,
             storage_account_data_private_sensitive_name=props.storage_account_data_private_sensitive_name,
-        ).apply(lambda kwargs: self.read_cloudinit(**kwargs))
+        ).apply(lambda kwargs: self.template_cloudinit(**kwargs))
+        b64cloudinit = Output.from_input(cloudinit).apply(
+            lambda cloudinit: b64encode(cloudinit)
+        )
```
jemrobinson: You shouldn't need `Output.from_input` as cloudinit should be an `Output[str]` already.

```suggestion
        b64cloudinit = cloudinit.apply(b64encode)
```

data_safe_haven/resources/workspace/ansible/host_vars/localhost.yaml
```diff
@@ -0,0 +1,122 @@
...
+  - category: database
+    common:
+      - libpq-dev     # interact with PostgreSQL databases
+      - msodbcsql17   # interact with Microsoft SQL databases
+      - unixodbc-dev  # interact with Microsoft SQL databases
+    focal:
+    jammy:
+    noble:
+  - category: development
+    common:
```
jemrobinson, line 33: Missing wrt 4.X.Y
- java
- rustc
- scala
- spark-shell

data_safe_haven/resources/workspace/ansible/host_vars/localhost.yaml
```diff
@@ -0,0 +1,122 @@
...
+      - python3.9
+    jammy:
+      - python3.10
+      - python3.10-venv
+      - python-is-python3
+    noble:
+      - python3.12
+      - python3.12-venv
+      - python-is-python3
+  - category: editors
```
jemrobinson, line 71: Missing wrt 4.X.Y
- code (this is what pretty much all our users use!)
- nano
- pycharm-community
- rstudio

data_safe_haven/resources/workspace/ansible/host_vars/localhost.yaml
```diff
@@ -0,0 +1,122 @@
...
+      - vim-airline-themes
+      - vim-fugitive
+      - vim-gtk3
+      - vim-julia
+      - vim-latexsuite
+      - vim-syntastic
+    focal:
+    jammy:
+    noble:
+  - category: office
```
jemrobinson, line 87: Missing wrt 4.X.Y
- latex
- xelatex

data_safe_haven/resources/workspace/ansible/host_vars/localhost.yaml
```diff
@@ -0,0 +1,122 @@
...
+    common:
+      - firefox
+    focal:
+    jammy:
+    noble:
+  - category: database
+    common:
+      - libpq-dev     # interact with PostgreSQL databases
+      - msodbcsql17   # interact with Microsoft SQL databases
+      - unixodbc-dev  # interact with Microsoft SQL databases
```
jemrobinson: Missing wrt 4.X.Y
- azuredatastudio
- dbeaver-ce
- psql
- sqlcmd

data_safe_haven/resources/workspace/ansible/host_vars/localhost.yaml
```diff
@@ -0,0 +1,122 @@
...
+      - bat
+      - fd-find
+      - fzf
+      - htop
+      - httpie
+      - nvtop
+      - ripgrep
+      - rsync
+      - tmux
+      - wget
```
jemrobinson, line 126: Missing wrt 4.X.Y
- git
- nvidia-smi
- weka

data_safe_haven/infrastructure/programs/sre/data.py
JimMadge: We probably actually want something like 550.

rx is needed for root to read and browse. Others shouldn't need it though.

jemrobinson (changes requested):

data_safe_haven/infrastructure/programs/sre/workspaces.py
JimMadge: I'm not particularly worried. There aren't any secrets there (at the moment). It would also only expose these to a user who can read all of the secrets already.

jemrobinson: OK, and do we want a multiline log message here? I thought we were trying to avoid those.

data_safe_haven/resources/workspace/ansible/host_vars/localhost.yaml
JimMadge: - Do we want to keep Java, scala? I would assume they are much less common
- rustc will be installed as part of cargo (we could add it explicitly though)
- Is spark-shell any use given we don't have a spark server?

jemrobinson: I think supporting as many programming languages as possible is part of our aim. Java and scala should be in Ubuntu's repos, so I imagine that's quite easy to support (`maven` has been a bit of a problem with scala projects in the past, but we can cross that bridge if we come to it again). Agreed that rustc should be installed by cargo, but let's be explicit. We can drop spark-shell if it's useless without a cluster (I don't know whether it can be run standalone).

Are there any other languages we should be considering here? e.g. dotnet/mono support?

data_safe_haven/resources/workspace/ansible/host_vars/localhost.yaml
JimMadge: - Added codium and pycharm-community as snaps.
- Nano is part of the base image but I've added explicitly

Rstudio is more difficult as posit host debs on their own site and don't distribute on the canonical repos or as a snap.
Can we make move that to a new issue?

data_safe_haven/resources/workspace/ansible/host_vars/localhost.yaml
JimMadge: Added texlive-full.

data_safe_haven/resources/workspace/ansible/host_vars/localhost.yaml
jemrobinson: Yes, any/all of these can be put into new issues, together or separately as you prefer. This was just to ensure that anything we're removing is done deliberately.

jemrobinson:
Deployments from this branch are failing with the following error:

```
 +  azure-native:network:VirtualNetwork sre_networking_virtual_network **creating failed** error: Code="InvalidResourceReference" Message="Resource
/subscriptions/3f1a8e26-eae2-4539-952a-0a6184ec248a/resourceGroups/SHM-GREEN-SRE-FERN-RG-NETWORKING/providers/Microsoft.Network/routeTables/SHM-GREEN-SRE-FERN-ROUTE-TABLE referenced by resource
/subscriptions/3f1a8e26-eae2-4539-952a-0a6184ec248a/resourceGroups/shm-green-sre-fern-rg-networking/providers/Microsoft.Network/virtualNetworks/shm-green-sre-fern-vnet was not found. Please make sure that the referenced resource exists, and that both resources are in
the same region." Details=[{"code":"NotFound","message":"Resource /subscriptions/3f1a8e26-eae2-4539-952a-0a6184ec248a/resourceGroups/SHM-GREEN-SRE-FERN-RG-NETWORKING/providers/Microsoft.Network/routeTables/SHM-GREEN-SRE-FERN-ROUTE-TABLE not found."}]
```

@craddm @JimMadge have you seen this?

data_safe_haven/resources/workspace/ansible/host_vars/localhost.yaml
JimMadge: Added dbeaver-ce as a snap and postgreql-client-common (which provides the psql command).

Azure Data Studio and sqlcmd are more difficult, requiring custom PPAs or fetching debs.
We could open a new issue, I feel like they overlap with existing tools though. I would lean towards not including them.

data_safe_haven/resources/workspace/ansible/host_vars/localhost.yaml
jemrobinson: re: `weka` there was an issue [a couple of years ago](https://github.com/alan-turing-institute/data-safe-haven/issues/710) with the version available through Ubuntu's repos being quite old. Worth checking whether this is still an issue.

JimMadge: Git was already there, I've added `git-all` though to include other modules (lfs and so on). I've added weka.

nvidia-smi should probably be installed as part of `nvidia-utils-???`. You need to specify a version, which really requires knowing which driver version your hardware needs. I think we should open an issue for that.

JimMadge:
@jemrobinson I've not seen that :eyes:.

data_safe_haven/resources/workspace/ansible/host_vars/localhost.yaml
JimMadge: Agreed. I think in practice Python > R >> everything else, but it shouldn't be too difficult to support more.

Didn't we find dotnet had some daemon that wanted to phone home?

JimMadge: We are missing functional languages, LISP dialects, Haskell, F#.

jemrobinson: F# would be covered by dotnet (I think - check with e.g. Evelina). If you can find a LISP/Haskell practitioner and find out what they need installed then yes, let's do that.

jemrobinson: > Didn't we find dotnet had some daemon that wanted to phone home?

I think you can (a) turn off dotnet telemetry and (b) ignore it, since failure to phone home doesn't break anything.

Maven was more of a problem, since scala programs wouldn't run until they'd updated all their packages.

data_safe_haven/resources/workspace/ansible/host_vars/localhost.yaml
```diff
@@ -31,15 +31,18 @@ package_categories:
...
       - cargo
       - cmake
       - default-jre
+      - dotnet-sdk-8.0
       - g++
       - gcc
       - gfortran
       - git-all
       - golang
+      - mono-devel
```
jemrobinson: Do we want dotnet AND mono? It's been a while since I had to think about this, but aren't they doing the same thing?

JimMadge: I'm not familiar with them. We could be opinionated at say mono only, but both are available so it seems like a low cost. I'm not sure how compatible they are.

This is a bit different to Matlab/Octave where there would also be licensing problems.

jemrobinson: IIRC Mono was an open-source implementation of DotNet when it didn't support Linux. Subsequently, Mono was bought by Microsoft and DotNet core is available for Linux. I think we probably don't need Mono here - can you check if that's right?

jemrobinson:
@JimMadge I'm getting the `python inline source runtime error: '' is not a valid BlobAccessTier` message. Do we need an updated version of `pulumi` or `pulumi-azure-native` to get the fix? If so, can you add the required change to this PR?

data_safe_haven/resources/workspace/ansible/host_vars/localhost.yaml
JimMadge: > IIRC Mono was an open-source implementation of DotNet when it didn't support Linux.

That might be true but I think there are still an important difference due to licenses and that .Net core doesn't implement all .Net functions (_e.g._ silverlight/moonlight).

I would suspect they are very similar but there would be some edge cases which might require one implementation in particular.
If we were to remove one, I would lean towards keeping on Mono as it is FOSS.

JimMadge:
@jemrobinson Which versions do you have?

I think `pulumi-azure-native` should be up to date in the requirements, and I _think_ that is all we need.

I also found that if the stack already had a blob tier defined I would still get the error. Removing that from the stack or deleting and recreating the resources fixed it.

data_safe_haven/resources/workspace/ansible/host_vars/localhost.yaml
jemrobinson: Wasn't Silverlight deprecated in the 2010s?

No strong preference between the two - can you ask someone who actually uses a .NET language which is better?

JimMadge: Can you point me in the direction of someone who uses .NET :laughing:? (I'm honestly not aware of anyone).

jemrobinson: @evelinag?

JimMadge: Looks like .NET is the way to go. It is MIT now and more actively developed.

We can worry about edge cases where Mono is needed over .NET when they happen :+1:

data_safe_haven/infrastructure/programs/sre/networking.py
```diff
@@ -1282,6 +1341,18 @@ def __init__(
...
```
jemrobinson:
```suggestion
                network.SecurityRuleArgs(
                    access=network.SecurityRuleAccess.ALLOW,
                    description="Allow outbound connections to desired state data endpoints.",
                    destination_address_prefix=SREIpRanges.data_desired_state.prefix,
                    destination_port_range="*",
                    direction=network.SecurityRuleDirection.OUTBOUND,
                    name="AllowDataDesiredStateEndpointsOutbound",
                    priority=NetworkingPriorities.INTERNAL_SRE_DATA_DESIRED_STATE,
                    protocol=network.SecurityRuleProtocol.ASTERISK,
                    source_address_prefix=SREIpRanges.workspaces.prefix,
                    source_port_range="*",
                ),
```

N.B. This will also need a new entry in the `NetworkingPriorities` enum.

By the way, do we not know which destination port this connection will use? If we do, we should add that to this rule.

jemrobinson (changes requested):

data_safe_haven/infrastructure/programs/sre/networking.py
JimMadge: I knew I would miss something :smile:, quite a few places to make changes.

JimMadge: > The NFS 3.0 protocol uses ports 111 and 2049. If you're connecting from an on-premises network, make sure that your client allows outgoing communication through those ports. If you have granted access to specific VNets, make sure that any network security groups associated with those VNets don't contain security rules that block incoming communication through those ports.

https://learn.microsoft.com/en-us/azure/storage/blobs/network-file-system-protocol-support#supported-network-connections

JimMadge: Fixed in ed11e31 and 69674c2

data_safe_haven/infrastructure/programs/sre/networking.py
```diff
@@ -1274,14 +1333,26 @@ def __init__(
                     access=network.SecurityRuleAccess.ALLOW,
                     description="Allow outbound connections to private data endpoints.",
                     destination_address_prefix=SREIpRanges.data_private.prefix,
-                    destination_port_range="*",
+                    destination_port_range=[Ports.NFS_V3_1, Ports.NFS_V3_2],
```
jemrobinson:
```suggestion
                    destination_port_ranges=[Ports.NFS_V3_1, Ports.NFS_V3_2],
```

data_safe_haven/infrastructure/programs/sre/networking.py
```diff
@@ -396,6 +396,65 @@ def __init__(
...
+            f"{self._name}_nsg_data_desired_state",
+            network_security_group_name=f"{stack_name}-nsg-data-desired-state",
+            resource_group_name=resource_group.name,
+            security_rules=[
+                # Inbound
+                network.SecurityRuleArgs(
+                    access=network.SecurityRuleAccess.ALLOW,
+                    description="Allow inbound connections from workspaces.",
+                    destination_address_prefix=SREIpRanges.data_configuration.prefix,
+                    destination_port_range=[Ports.NFS_V3_1, Ports.NFS_V3_2],
```
jemrobinson:
```suggestion
                    destination_port_ranges=[Ports.NFS_V3_1, Ports.NFS_V3_2],
```

data_safe_haven/infrastructure/programs/sre/networking.py
```diff
@@ -406,7 +465,7 @@ def __init__(
                     access=network.SecurityRuleAccess.ALLOW,
                     description="Allow inbound connections from SRE workspaces.",
                     destination_address_prefix=SREIpRanges.data_private.prefix,
-                    destination_port_range="*",
+                    destination_port_range=[Ports.NFS_V3_1, Ports.NFS_V3_2],
```
jemrobinson:
```suggestion
                    destination_port_ranges=[Ports.NFS_V3_1, Ports.NFS_V3_2],
```

jemrobinson:
Currently hitting a problem with an incorrect type being passed to Pulumi:

```
  pulumi:pulumi:Stack (data-safe-haven-shm-green-sre-fern):
    error: python inline source runtime error: Unexpected type. Expected 'list' got '<class 'str'>'
    Traceback (most recent call last):
      File "/Users/jrobinson/Library/Application Support/hatch/env/virtual/data-safe-haven/Svgq2p76/data-safe-haven/lib/python3.12/site-packages/pulumi/automation/_server.py", line 80, in run
        loop.run_until_complete(run_in_stack(self.program))
      File "/opt/homebrew/Cellar/python@3.12/3.12.4/Frameworks/Python.framework/Versions/3.12/lib/python3.12/asyncio/base_events.py", line 687, in run_until_complete
        return future.result()
               ^^^^^^^^^^^^^^^
      File "/Users/jrobinson/Library/Application Support/hatch/env/virtual/data-safe-haven/Svgq2p76/data-safe-haven/lib/python3.12/site-packages/pulumi/runtime/stack.py", line 138, in run_in_stack
        await run_pulumi_func(run)
      File "/Users/jrobinson/Library/Application Support/hatch/env/virtual/data-safe-haven/Svgq2p76/data-safe-haven/lib/python3.12/site-packages/pulumi/runtime/stack.py", line 52, in run_pulumi_func
        await wait_for_rpcs()
      File "/Users/jrobinson/Library/Application Support/hatch/env/virtual/data-safe-haven/Svgq2p76/data-safe-haven/lib/python3.12/site-packages/pulumi/runtime/stack.py", line 114, in wait_for_rpcs
        await task
      File "/Users/jrobinson/Library/Application Support/hatch/env/virtual/data-safe-haven/Svgq2p76/data-safe-haven/lib/python3.12/site-packages/pulumi/runtime/resource.py", line 909, in do_register
        resolver = await prepare_resource(res, ty, custom, remote, props, opts, typ)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/Users/jrobinson/Library/Application Support/hatch/env/virtual/data-safe-haven/Svgq2p76/data-safe-haven/lib/python3.12/site-packages/pulumi/runtime/resource.py", line 196, in prepare_resource
        serialized_props = await rpc.serialize_properties(
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/Users/jrobinson/Library/Application Support/hatch/env/virtual/data-safe-haven/Svgq2p76/data-safe-haven/lib/python3.12/site-packages/pulumi/runtime/rpc.py", line 215, in serialize_properties
        result = await serialize_property(
                 ^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/Users/jrobinson/Library/Application Support/hatch/env/virtual/data-safe-haven/Svgq2p76/data-safe-haven/lib/python3.12/site-packages/pulumi/runtime/rpc.py", line 359, in serialize_property
        await serialize_property(
      File "/Users/jrobinson/Library/Application Support/hatch/env/virtual/data-safe-haven/Svgq2p76/data-safe-haven/lib/python3.12/site-packages/pulumi/runtime/rpc.py", line 521, in serialize_property
        k: await serialize_property(
           ^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/Users/jrobinson/Library/Application Support/hatch/env/virtual/data-safe-haven/Svgq2p76/data-safe-haven/lib/python3.12/site-packages/pulumi/runtime/rpc.py", line 355, in serialize_property
        element_type = _get_list_element_type(typ)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/Users/jrobinson/Library/Application Support/hatch/env/virtual/data-safe-haven/Svgq2p76/data-safe-haven/lib/python3.12/site-packages/pulumi/runtime/rpc.py", line 168, in _get_list_element_type
        raise AssertionError(f"Unexpected type. Expected 'list' got '{typ}'")
    AssertionError: Unexpected type. Expected 'list' got '<class 'str'>'
    error: Exception calling application: invalid state
```

can you see where this is happening @JimMadge ?

JimMadge:
@jemrobinson I didn't see that before, so it is likely it was introduce during the networking changes,

- Adding desired state subnet
- Changing ports in NSG rules

Is there no more output? When I had similar errors I think there was often a clue to which line in our code had a problem.

jemrobinson:
I think I've fixed it with 39aa441.

JimMadge:
Sorry, I didn't catch the argument for a port string or list of strings was different 😅.

data_safe_haven/infrastructure/programs/sre/networking.py
```diff
@@ -399,6 +399,65 @@ def __init__(
             opts=child_opts,
             tags=child_tags,
         )
+        nsg_data_desired_state = network.NetworkSecurityGroup(
+            f"{self._name}_nsg_data_desired_state",
+            network_security_group_name=f"{stack_name}-nsg-data-desired-state",
```
jemrobinson, line 405:
```suggestion
            location=props.location,
            network_security_group_name=f"{stack_name}-nsg-data-desired-state",
```

jemrobinson:
OK, I've now got this deployed, but I can't connect to the Workspace from the Guacamole dashboard. Is this working for you @JimMadge ?

data_safe_haven/infrastructure/programs/sre/data.py
```diff
@@ -456,6 +461,154 @@ def __init__(
...
+                blob_name=file_path,
+                container_name=container_desired_state.name,
+                resource_group_name=resource_group.name,
+                source=file_asset,
+            )
+        # Set up a private endpoint for the desired state storage account
+        storage_account_data_desired_state_endpoint = network.PrivateEndpoint(
+            f"{storage_account_data_desired_state._name}_private_endpoint",
+            location=props.location,
+            private_endpoint_name=f"{stack_name}-pep-storage-account-desired-state",
```
jemrobinson:
```suggestion
            private_endpoint_name=f"{stack_name}-pep-storage-account-data-desired-state",
```

jemrobinson:
N.B. This is caused by a failure to mount the `/desiredstate` volume. Could be due to overly-restrictive NSG rules (e.g. the recent change to which ports are allowed)? Can you take a look @JimMadge ?

```
root@shm-green-sre-fern-vm-workspace-01:~# mount -fav
/                        : ignored
/boot/efi                : already mounted
mount.nfs: timeout set for Thu Jul 18 13:30:34 2024
mount.nfs: trying text-based options 'sec=sys,vers=3,nolock,proto=tcp,addr=10.0.1.20'
mount.nfs: prog 100003, trying vers=3, prot=6
mount.nfs: trying 10.0.1.20 prog 100003 vers 3 prot TCP port 2048
```

```
root@shm-green-sre-fern-vm-workspace-01:~# nslookup shgresreferdesiredstatec.blob.core.windows.net
Server:         127.0.0.53
Address:        127.0.0.53#53

Non-authoritative answer:
shgresreferdesiredstatec.blob.core.windows.net  canonical name = shgresreferdesiredstatec.privatelink.blob.core.windows.net.
Name:   shgresreferdesiredstatec.privatelink.blob.core.windows.net
Address: 10.0.1.20
```

data_safe_haven/resources/workspace/workspace.cloud_init.mustache.yaml
```diff
@@ -75,57 +107,17 @@ apt:
...
-
-  # Enable xrdp and session manager at startup
-  # ------------------------------------------
-  - echo ">=== Enable xrdp... ===<"
-  - systemctl enable xrdp
-  - systemctl enable xrdp-sesman
-  - systemctl start xrdp
-  - systemctl start xrdp-sesman
 
   # Mount all external volumes
```
jemrobinson, line 123: Add a test for /desiredstate here too.

N.B. Is there a better place we can put this than as a new root-level directory? There must be an established location for config files that would be a better mount point.

jemrobinson:
Looks like you may have added the wrong NFS ports. [This doc](https://learn.microsoft.com/en-us/azure/storage/blobs/network-file-system-protocol-support-how-to) says we need 111 and 2048.

JimMadge:
@jemrobinson 🤦 I think we are trying to debug at the same time. Saw the port change from 2049 to 2048 before my eyes 😆.

data_safe_haven/resources/workspace/workspace.cloud_init.mustache.yaml
JimMadge: I would probably put all of the mounts in `/mnt` when is conventionally used for non-system disks at network drives. (`/media` or more recently `/run/media` for removable media).

Do you want to open an issue?

jemrobinson: Captured in #2027 and marked for rc3.

data_safe_haven/resources/workspace/workspace.cloud_init.mustache.yaml
```diff
@@ -75,65 +107,34 @@ apt:
...
   # --------------------------
   - echo ">=== Mounting all external volumes... ===<"
   - grep -v -e '^[[:space:]]*$' /etc/fstab | sed 's|^|  /etc/fstab  |'
   - mount -fav
   - while (! mountpoint -q /data); do sleep 5; mount /data; done
+  - while (! mountpoint -q /desired_state); do sleep 5; mount /data; done
   - while (! mountpoint -q /home); do sleep 5; mount /home; done
   - while (! mountpoint -q /output); do sleep 5; mount /output; done
   - while (! mountpoint -q /shared); do sleep 5; mount /shared; done
-  - findmnt
```
jemrobinson, line 139: What's the reason for dropping `findmnt`? Is it because we think sysadmins will need/want to look at these log files? Or is this done better somewhere else?

jemrobinson:
The mount is still failing. I'm going to try with `other: r-x` permissions as I'm not sure which user the storage account is mounted with.

jemrobinson:
Container mounting is fixed but GUI login is still not working. Running commands on the serial console indicates ansible probably hasn't run:

```
dshadmin@shm-green-sre-fern-vm-workspace-01:~$ which rustc
dshadmin@shm-green-sre-fern-vm-workspace-01:~$ which cargo
dshadmin@shm-green-sre-fern-vm-workspace-01:~$ 
dshadmin@shm-green-sre-fern-vm-workspace-01:~$ systemctl status desired-state
○ desired-state.service - Run desired state script
     Loaded: loaded (/etc/systemd/system/desired-state.service; static)
     Active: inactive (dead)
```

any thoughts on how to debug this @JimMadge ?

data_safe_haven/resources/workspace/workspace.cloud_init.mustache.yaml
JimMadge: An accident 😅.

JimMadge:
> Container mounting is fixed but GUI login is still not working. Running commands on the serial console indicates ansible probably hasn't run:
> 
> ```
> dshadmin@shm-green-sre-fern-vm-workspace-01:~$ which rustc
> dshadmin@shm-green-sre-fern-vm-workspace-01:~$ which cargo
> dshadmin@shm-green-sre-fern-vm-workspace-01:~$ 
> dshadmin@shm-green-sre-fern-vm-workspace-01:~$ systemctl status desired-state
> ○ desired-state.service - Run desired state script
>      Loaded: loaded (/etc/systemd/system/desired-state.service; static)
>      Active: inactive (dead)
> ```
> 
> any thoughts on how to debug this @JimMadge ?

The `systemctl` output is what I would expect. It is a one-shot service so one the command has finished the unit won't be active.

Looking at `journalctl -u desired-state.service` should show the output. You can run the script at `/root/desired_state.sh`.
It's also quite easy to make changes because you can use `az` to update the ansible project on a deployed environment.

```command
az storage blob upload-batch --account-name <storageaccount> -d <container> -s ./ --overwrite
```

JimMadge:
@jemrobinson It is working now (and probably _was_ before although there was a non-zero exit code).

The problem fixed in 1713ae5 fixed the exit code.

Was the output you posted above from a VM that had been deployed while the mount was broken?
I logged into a deployed VM and rustc and cargo were installed. Running Ansible made no changes.
I suspect at deployment the NFS share wasn't mounted, so the initial run of `desired-state.service` didn't work. But the timer ran it overnight, so when I looked this morning it was all configured.

jemrobinson:
@JimMadge: I think we need to EITHER wait until the mount is definitely available before running the one-shot service (I *thought* this should be what the existing cloud-init code is doing but it doesn't seem to be) OR add an explicit dependency in the service that the mount exists (I haven't tried this, but I think it's something like `After=mnt-whatever.mount`)

JimMadge:
Yes it should do.

Looking at a fresh deployment, the timer isn't active although `systemctl enable desired-state.service` should have run. It might be that `systemctl daemon-reload` wasn't run after writing unit files.

JimMadge:
@jemrobinson it looks like cloud-init isn't getting past the mount stage.

```
>=== Mounting all external volumes... ===<
  /etc/fstab  # CLOUD_IMG: This file was created/modified by the Cloud Image build process
  /etc/fstab  UUID=fd0e4ffa-6c5f-49cc-9a56-46113512bb1a /        ext4   discard,errors=remount-ro       0 1
  /etc/fstab  UUID=6C4C-3E9C    /boot/efi       vfat    umask=0077      0 1
  /etc/fstab  shdaisreshidesiredstatec.blob.core.windows.net:/shdaisreshidesiredstatec/desiredstate     /desired_state  nfs     ro,_netdev,sec=sys,vers=3,nolock,proto=tcp,comment=cloudconfig       0       2
  /etc/fstab  shdaisreshisensitivedata.blob.core.windows.net:/shdaisreshisensitivedata/ingress  /data   nfs     ro,_netdev,sec=sys,vers=3,nolock,proto=tcp,comment=cloudconfig  0   2
  /etc/fstab  shdaisreshisensitivedata.blob.core.windows.net:/shdaisreshisensitivedata/egress   /output nfs     rw,_netdev,sec=sys,vers=3,nolock,proto=tcp,comment=cloudconfig  0   2
  /etc/fstab  shmdaimysreshimauserdata.file.core.windows.net:/shmdaimysreshimauserdata/shared   /shared nfs     _netdev,sec=sys,nconnect=4,comment=cloudconfig  0       2
  /etc/fstab  shmdaimysreshimauserdata.file.core.windows.net:/shmdaimysreshimauserdata/home     /home   nfs     _netdev,sec=sys,nconnect=4,comment=cloudconfig  0       2
mount.nfs: trying 10.0.1.20 prog 100003 vers 3 prot TCP port 2048
mount.nfs: trying 10.0.1.20 prog 100005 vers 3 prot TCP port 2048
mount.nfs: timeout set for Fri Jul 19 09:41:30 2024
mount.nfs: trying text-based options 'sec=sys,vers=3,nolock,proto=tcp,addr=10.0.1.20'
mount.nfs: prog 100003, trying vers=3, prot=6
mount.nfs: prog 100005, trying vers=3, prot=6
mount.nfs: trying 10.0.1.29 prog 100003 vers 3 prot TCP port 2048
mount.nfs: trying 10.0.1.29 prog 100005 vers 3 prot TCP port 2048
mount.nfs: timeout set for Fri Jul 19 09:41:30 2024
mount.nfs: trying text-based options 'sec=sys,vers=3,nolock,proto=tcp,addr=10.0.1.29'
mount.nfs: prog 100003, trying vers=3, prot=6
mount.nfs: prog 100005, trying vers=3, prot=6
mount.nfs: trying 10.0.1.29 prog 100003 vers 3 prot TCP port 2048
mount.nfs: trying 10.0.1.29 prog 100005 vers 3 prot TCP port 2048
mount.nfs: timeout set for Fri Jul 19 09:41:30 2024
mount.nfs: trying text-based options 'sec=sys,vers=3,nolock,proto=tcp,addr=10.0.1.29'
mount.nfs: prog 100003, trying vers=3, prot=6
mount.nfs: prog 100005, trying vers=3, prot=6
mount.nfs: timeout set for Fri Jul 19 09:41:30 2024
mount.nfs: trying text-based options 'sec=sys,nconnect=4,vers=4.2,addr=10.0.1.28,clientaddr=10.0.2.4'
mount.nfs: timeout set for Fri Jul 19 09:41:30 2024
mount.nfs: trying text-based options 'sec=sys,nconnect=4,vers=4.2,addr=10.0.1.28,clientaddr=10.0.2.4'
/                        : ignored
/boot/efi                : already mounted
/desired_state           : successfully mounted
/data                    : successfully mounted
/output                  : successfully mounted
/shared                  : successfully mounted
/home                    : successfully mounted
mount.nfs: Connection timed out
mount.nfs: Connection timed out
mount.nfs: Connection timed out
```

Maybe because of the NFS changes introduced here?
Can we try rolling those back and open an issue about a stricter configuration (if it works)?.

JimMadge:
~~Or, is it only ingress that is failing?~~

jemrobinson:
Sure - is it just the port change we need to revert? That seems fine, esp. if we capture it in another issue. BTW, running `mount -fav` manually *does* mount the drives so I'm not sure why this isn't working in the script.

JimMadge:
Yes, it is strange, the output it saying timeout. I also saw that on the SRE you deployed when running Ansible, occasional timeout messages.

I don't remember seeing that before, so possibly related to networking changes.

JimMadge:
cloudinit still seems to be stalling. Doesn't seem to get past the `mountpoint` loops.

JimMadge:
Found a bug 🤞

JimMadge:
Ok, it moves on to running desired state now.

JimMadge:
```
[ 3485.382261] desired_state.sh[4934]: PLAY RECAP *********************************************************************
[ 3485.382793] desired_state.sh[4934]: localhost                  : ok=16   changed=11   unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
[  OK  ] Finished Run desired state script.
Jul 19 15:02:11 shm-daimyo-sre-oda-vm-workspace-01 systemd[1]: desired-state.service: Deactivated successfully.
Jul 19 15:02:11 shm-daimyo-sre-oda-vm-workspace-01 systemd[1]: Finished Run desired state script.
Jul 19 15:02:11 shm-daimyo-sre-oda-vm-workspace-01 systemd[1]: desired-state.service: Consumed 7min 45.817s CPU time.
[  OK  ] Finished Execute cloud user/final scripts.
[  OK  ] Reached target Cloud-init target.
```

JimMadge:
```command
root@shm-daimyo-sre-oda-vm-workspace-01:/# systemctl status desired-state.timer
● desired-state.timer - Run desired state script daily
     Loaded: loaded (/etc/systemd/system/desired-state.timer; enabled; vendor p>
     Active: active (waiting) since Fri 2024-07-19 14:45:36 UTC; 20min ago
    Trigger: Sat 2024-07-20 00:00:00 UTC; 8h left
   Triggers: ● desired-state.service
```

data_safe_haven/resources/workspace/workspace.cloud_init.mustache.yaml
```diff
@@ -126,7 +126,7 @@ runcmd:
   - grep -v -e '^[[:space:]]*$' /etc/fstab | sed 's|^|  /etc/fstab  |'
   - mount -fav
   - while (! mountpoint -q /data); do sleep 5; mount /data; done
-  - while (! mountpoint -q /desired_state); do sleep 5; mount /data; done
+  - while (! mountpoint -q /desired_state); do sleep 5; mount /desired_state; done
```
jemrobinson, line 129: :)

JimMadge: 🤦

jemrobinson:
1. Cannot log in via GUI with user credentials (grey error box that usually indicates incorrect username/password)
Screenshot 2024-07-21 at 13 19 29

2. Cannot log in via serial console with admin credentials (not recognised)
Screenshot 2024-07-21 at 13 19 57

jemrobinson:
Also blob creation in the desired state account is failing on a fresh deployment:

```
  azure-native:storage:Blob (sre_data_blob_desired_state_blob_run_all_tests.bats):
    error: retrieving blob properties "files/home/dshadmin/tests/run_all_tests.bats" (container "desiredstate" / account "shgresrebotdesiredstatec"): blobs.Client#GetProperties: Failure responding to request: StatusCode=403 -- Original Error:
autorest/azure: error response cannot be parsed: {"" '\x00' '\x00'} error: EOF
```

JimMadge:
I did see that serial console problem. I did a fresh deployment and it didn't occur again. I assumed it was some problem unrelated to the changes here.

Haven't seen the azure-native Blob error. Looks like an authentication problem.

My output above was from a fresh sre deployment at 226bebd.

jemrobinson:
I think the blob issue is actually a Storage Account issue, where sometimes the network permissions on the storage account ("IP address A.B.C.D can access this account") don't translate into *actually* being able to access the account from that IP address. Worth moving this into its own issue as I don't think it's related to this PR.

JimMadge:
New deployment at [46af571](https://github.com/alan-turing-institute/data-safe-haven/pull/1892/commits/46af57159d2f0405e3c91bfcac99682ee05e2c79), Pulumi ran without error, Ansible is now running on the workspace VM.

JimMadge:
Replicated serial console and guacamole login problems.

#2028 is possibly causing guacamole login problems (redeployed Entra ID App doesn't have the correct token to authenticate with LDAP).

```command
# ldapsearch -H "ldap://identity.oda.daimyo.develop.turingsafehaven.ac.uk:1389"
ldap_sasl_interactive_bind: Protocol error (2)
        additional info: LDAP search request failed. Failed to fetch bearer token from OAuth endpoint.
(unauthorized_client) AADSTS700016: Application with identifier 'd13c364d-992e-4ff2-8911-40c6646c82fd' was not found in the directory 'Turing Data Safe Haven (Green)'. This can happen if the application has not been installed by the administrator of the tenant or consented to by any user in the tenant. You may have sent your authentication request to the wrong tenant. Trace ID: 02613c3e-331e-4e2c-b034-1b317be65200 Correlation ID: 302e5952-a067-4456-95d8-4500577d0602 Timestamp: 2024-07-22 10:49:52Z

**Comment by JimMadge** [comment]
For serial console issue, it may be a sporadic problem.

I will try,

- Adding workspaces
- Running deployment without cloudinit and Ansible

I would be surprised if it were cloudinit or Ansible (although I can't think what it _is_) as they shouldn't modify that user and using the portal password reset (either to change `dshadmin`'s password, or create a new user) doesn't work.

**Comment by JimMadge** [comment]
Some progress,

I can log in to the serial console using the `dshadmin` account _before_ cloudinit has finished running. That sessions works as long as it is active. If you log out after cloudinit has finished, wait until it has finished before logging in, restart the VM, you get the login incorrect error.

That would suggest there is something in cloud init/ansible that breaks the account.
(Still strange that resetting doesn't appear to work).

**Comment by JimMadge** [comment]
In a VM where I had a session,

- `# su dshadmin` works (mimic user without password)
- `login dshadmin` does not work, login incorrect error
- `# passwd dshadmin` lets you change the password, login still fails.

**Comment by JimMadge** [comment]
Suspect it may be a PAM configuration issue.

**Comment by JimMadge** [comment]
Before and after correcting `/etc/pam.d/common-session`,

```
Jul 22 14:51:57 shm-daimyo-sre-oda-vm-workspace-01 login[100118]: PAM (other) illegal module type: pam_mkhomedir.so
Jul 22 14:51:57 shm-daimyo-sre-oda-vm-workspace-01 login[100118]: PAM pam_parse: expecting return value; [...skel=/etc/skel]
Jul 22 14:52:01 shm-daimyo-sre-oda-vm-workspace-01 login[100118]: FAILED LOGIN (1) on '/dev/pts/0' FOR 'dshadmin', Permission denied
Jul 22 14:52:09 shm-daimyo-sre-oda-vm-workspace-01 login[100118]: FAILED LOGIN (2) on '/dev/pts/0' FOR 'dshadmin', Permission denied
Jul 22 14:56:49 shm-daimyo-sre-oda-vm-workspace-01 login[101098]: pam_unix(login:session): session opened for user dshadmin(uid=1000) by dshadmin(uid=0)
Jul 22 14:56:54 shm-daimyo-sre-oda-vm-workspace-01 login[101098]: pam_unix(login:session): session closed for user dshadmin
Jul 22 15:02:17 shm-daimyo-sre-oda-vm-workspace-01 login[102510]: pam_unix(login:session): session opened for user dshadmin(uid=1000) by dshadmin(uid=0)
Jul 22 15:02:22 shm-daimyo-sre-oda-vm-workspace-01 login[102510]: pam_unix(login:session): session closed for user dshadmin
```

**Comment by jemrobinson** [comment]
For reference, here's `/etc/pam.d/common-session` from an SRE deployed from `develop`

```
#
# /etc/pam.d/common-session - session-related modules common to all services
#
# This file is included from other service-specific PAM config files,
# and should contain a list of modules that define tasks to be performed
# at the start and end of interactive sessions.
#
# As of pam 1.0.1-6, this file is managed by pam-auth-update by default.
# To take advantage of this, it is recommended that you configure any
# local modules either before or after the default block, and use
# pam-auth-update to manage selection of other modules.  See
# pam-auth-update(8) for details.

# here are the per-package modules (the "Primary" block)
session [default=1]                     pam_permit.so
# here's the fallback if no module succeeds
session requisite                       pam_deny.so
# prime the stack with a positive return value if there isn't one already;
# this avoids us returning an error just because nothing sets a success code
# since the modules above will each just jump around
session required                        pam_permit.so
# The pam_umask module will set the umask according to the system default in
# /etc/login.defs and user settings, solving the problem of different
# umask settings with different shells, display managers, remote sessions etc.
# See "man pam_umask".
session optional                        pam_umask.so
# and here are more per-package modules (the "Additional" block)
session required                        pam_unix.so 
session [success=ok default=ignore]     pam_ldap.so minimum_uid=1000
session optional                        pam_systemd.so 
session optional                        pam_mkhomedir.so skel=/etc/skel umask=0022
# end of pam-auth-update config
```

Can you compare to the one deployed from this branch?

JimMadge:
@jemrobinson The same (now) 👍
Before there was a line starting with `pam_mkhomedir.so`.

jemrobinson:
I can log into the serial console with these changes and #2034 fixes the Entra application issue. Looking into a problem with the apricot identity server.

jemrobinson:
@JimMadge : Interactive login is now working (with the fix from #2034 applied) but I'm getting a prompt that needs sudo approval. Can we get ansible to do this or alternatively, remove it if it's not needed?

Screenshot 2024-07-23 at 08 29 55

There's a potential solution here: https://www.reddit.com/r/Ubuntu/comments/15stmwn/how_do_i_suppress_authentication_is_required_to/ but I haven't tested it.

Captured in #2035

JimMadge:
@jemrobinson Excellent, thanks for testing.

I'm sure we had that error before and fixed in in v3-4. We should have a look in the old cloud-init, issues/PRs.
I assume you can close the prompt and ignore it safely?

Can we,

1. Merge this and #2034
2. Open an issue for the prompt?

jemrobinson:
Screenshot 2024-07-23 at 08 50 47

looks like we don't have `python-is-python3` installed.

JimMadge:
I've moved `python-is-python3` to the common list in dfda85d

jemrobinson (approved):
Happy with this. There are some remaining problems that have been pulled out into their own issues.

PR #1893, opened 2024-05-16
Fix next_occurrence function

### LINKED ISSUE #1895: Tests for next_occurrence break at certain times of day

- **Operating System:** macOS
- **Data Safe Haven version:** develop @ 9e5b72a

## :package: Packages

<details>

```none
acme==2.6.0
annotated-types==0.6.0
appdirs==1.4.4
Arpeggio==2.0.2
attrs==23.2.0
azure-common==1.1.28
azure-core==1.30.1
azure-identity==1.16.0
azure-keyvault-certificates==4.8.0
azure-keyvault-keys==4.9.0
azure-keyvault-secrets==4.8.0
azure-mgmt-automation==1.0.0
azure-mgmt-compute==30.6.0
azure-mgmt-containerinstance==10.1.0
azure-mgmt-core==1.4.0
azure-mgmt-dns==8.1.0
azure-mgmt-keyvault==10.3.0
azure-mgmt-msi==7.0.0
azure-mgmt-network==25.3.0
azure-mgmt-rdbms==10.1.0
azure-mgmt-resource==23.0.1
azure-mgmt-storage==21.1.0
azure-storage-blob==12.19.1
azure-storage-file-datalake==12.14.0
azure-storage-file-share==12.15.0
certifi==2024.2.2
cffi==1.16.0
charset-normalizer==3.3.2
chevron==0.14.0
click==8.1.7
cryptography==42.0.6
-e git+ssh://git@github.com/jemrobinson/data-safe-haven.git@cb3d09511884bff82ab33b39e18b1db74dad7959#egg=data_safe_haven
dill==0.3.8
dnspython==2.4.2
fqdn==1.5.1
grpcio==1.60.1
idna==3.7
iniconfig==2.0.0
isodate==0.6.1
josepy==1.14.0
markdown-it-py==3.0.0
mdurl==0.1.2
msal==1.28.0
msal-extensions==1.1.0
msrest==0.7.1
oauthlib==3.2.2
packaging==24.0
parver==0.5
pluggy==1.5.0
portalocker==2.8.2
protobuf==4.25.3
psycopg==3.1.18
pulumi==3.115.0
pulumi_azure_native==2.39.0
pulumi_random==4.16.1
pycparser==2.22
pydantic==2.7.1
pydantic_core==2.18.2
Pygments==2.18.0
PyJWT==2.8.0
pyOpenSSL==24.1.0
pyRFC3339==1.1
pytest==8.2.0
pytest-asyncio==0.23.6
pytz==2023.4
PyYAML==6.0.1
requests==2.31.0
requests-oauthlib==2.0.0
rich==13.7.1
semver==2.13.0
setuptools==69.5.1
shellingham==1.5.4
simple-acme-dns==3.0.0
six==1.16.0
typer==0.12.3
typing_extensions==4.11.0
urllib3==2.2.1
validators==0.22.0
websocket-client==1.8.0
```

</details>

At certain times of day `next_occurrence` can choose a datetime more than 1 day in the future, causing the tests to break.

<details>

```none
___________________ test_next_occurrence_is_within_next_day ____________________

    def test_next_occurrence_is_within_next_day():
        next_time = next_occurrence(5, 13, "Australia/Perth")
        dt_next_time = datetime.datetime.fromisoformat(next_time)
        dt_utc_now = datetime.datetime.now(datetime.UTC)
        assert dt_next_time > dt_utc_now
>       assert dt_next_time < dt_utc_now + datetime.timedelta(days=1)
E       AssertionError: assert datetime.datetime(2024, 5, 17, 21, 13, tzinfo=datetime.timezone.utc) < (datetime.datetime(2024, 5, 16, 16, 52, 28, 540053, tzinfo=datetime.timezone.utc) + datetime.timedelta(days=1))
E        +  where datetime.timedelta(days=1) = <class 'datetime.timedelta'>(days=1)
E        +    where <class 'datetime.timedelta'> = datetime.timedelta

tests/functions/test_strings.py:16: AssertionError
```

</details>

## :recycle: To reproduce

Run `hatch run test:test` at around 16:15 UTC.

jemrobinson:

n/a

Ensure that next_occurrence only adds one day when necessary, to avoid it choosing a datetime more than 1 day in the future

Closes #1895

### :microscope: Tests

Check CI

### DIFF
--- a/data_safe_haven/functions/strings.py
+++ b/data_safe_haven/functions/strings.py
@@ -48,9 +48,11 @@ def next_occurrence(
             microsecond=0,
         )
         utc_dt = local_dt.astimezone(pytz.utc)
-        # Add one day until this datetime is in the future
-        utc_now = datetime.datetime.now(pytz.utc) + datetime.timedelta(minutes=1)
-        while utc_dt < utc_now:
+        # Add one day until this datetime is at least 1 hour in the future.
+        # This ensures that any Azure functions which depend on this datetime being in
+        # the future should treat it as valid.
+        utc_near_future = datetime.datetime.now(pytz.utc) + datetime.timedelta(hours=1)
+        while utc_dt < utc_near_future:
             utc_dt += datetime.timedelta(days=1)
         if time_format == "iso":
             return utc_dt.isoformat()
--- a/data_safe_haven/functions/strings.py
+++ b/data_safe_haven/functions/strings.py
@@ -46,8 +46,12 @@ def next_occurrence(
             minute=minute,
             second=0,
             microsecond=0,
-        ) + datetime.timedelta(days=1)
+        )
         utc_dt = local_dt.astimezone(pytz.utc)
+        # Add one day until this datetime is in the future
+        utc_now = datetime.datetime.now(pytz.utc) + datetime.timedelta(minutes=1)
+        while utc_dt < utc_now:
+            utc_dt += datetime.timedelta(days=1)
         if time_format == "iso":
             return utc_dt.isoformat()
         elif time_format == "iso_minute":
--- a/tests/functions/test_strings.py
+++ b/tests/functions/test_strings.py
@@ -7,46 +7,52 @@ from data_safe_haven.exceptions import DataSafeHavenInputError
 from data_safe_haven.functions import next_occurrence, sanitise_sre_name
 
 
-@freeze_time(datetime.datetime(2024, 1, 2, 1, 0, tzinfo=datetime.UTC))
-def test_next_occurrence():
-    next_time = next_occurrence(5, 13, "Australia/Perth")
-    assert next_time == "2024-01-02T21:13:00+00:00"
+class TestNextOccurance:
+    @pytest.mark.parametrize("hour,minute,timezone,expected", [
+        (5, 13, "Australia/Perth", "2024-01-02T21:13:00+00:00"),
+        (0, 13, "Australia/Perth", "2024-01-02T16:13:00+00:00"),
+        (20, 13, "Australia/Perth", "2024-01-02T12:13:00+00:00"),
+        (20, 13, "Europe/London", "2024-01-02T20:13:00+00:00"),
+    ])
+    @freeze_time(datetime.datetime(2024, 1, 2, 1, 0, tzinfo=datetime.UTC))
+    def test_next_occurrence(self, hour, minute, timezone, expected):
+        next_time = next_occurrence(hour, minute, timezone)
+        assert next_time == expected
 
+    @freeze_time(datetime.datetime(2024, 7, 2, 1, 0, tzinfo=datetime.UTC))
+    def test_dst(self):
+        next_time = next_occurrence(13, 5, "Europe/London")
+        assert next_time == "2024-07-02T12:05:00+00:00"
 
-@freeze_time(datetime.datetime(2024, 1, 2, 1, 0, tzinfo=datetime.UTC))
-def test_next_occurrence_timeformat():
-    next_time = next_occurrence(5, 13, "Australia/Perth", time_format="iso_minute")
-    assert next_time == "2024-01-02 21:13"
+    @freeze_time(datetime.datetime(2024, 1, 2, 1, 0, tzinfo=datetime.UTC))
+    def test_timeformat(self):
+        next_time = next_occurrence(5, 13, "Australia/Perth", time_format="iso_minute")
+        assert next_time == "2024-01-02 21:13"
 
+    @freeze_time(datetime.datetime(2024, 1, 2, 23, 0, tzinfo=datetime.UTC))
+    def test_is_tomorrow(self):
+        next_time = next_occurrence(5, 13, "Australia/Perth")
+        assert next_time == "2024-01-03T21:13:00+00:00"
 
-@freeze_time(datetime.datetime(2024, 1, 2, 23, 0, tzinfo=datetime.UTC))
-def test_next_occurrence_is_tomorrow():
-    next_time = next_occurrence(5, 13, "Australia/Perth")
-    assert next_time == "2024-01-03T21:13:00+00:00"
+    def test_invalid_hour(self):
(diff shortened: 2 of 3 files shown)


JimMadge (changes requested):
I don't like this.

I've lost track of what `next_occurrence` is trying to do and adding minutes in odd places feels like fudge factors.

```python
utc_now = datetime.datetime.now(pytz.utc) + datetime.timedelta(minutes=1)
```

This isn't true, for example, that is now plus one minute.

If the problem is the test isn't true sometimes, it would be better to make the tests deterministic by not using `datetime.now()` but a set of known times instead.

jemrobinson:
`next_occurrence` is supposed to give the next time that a desired local time happens in UTC. For some Azure applications the time that you pass **cannot** be in the past.

Previously, we took local-time-in-UTC + 1 day which is always in the future (over the lifetime of a deploy operation). However, this can break the test which assumes that the output of `next_occurrence` will always be within one day of current-time-in-UTC.

Possible solutions are:

1. relax the test so that `next_occurrence` must be within 2 days of now
2. update the function so that it only adds 1 day if the suggested time is in the past
  - checking for in-the-past uses now-plus-1-minute to ensure that we don't end up with a time that is currently valid, but ends up being invalid by the time it gets given to an Azure function
  - possibly this should be longer (e.g. 5 minutes) and have a better explanation as to why
  
I'm happy to switch to solution (1) if you prefer.

JimMadge:
Hmm, some thoughts that might help,

- Why does the test assume `next_occurrence` will be within one day?
  Does it have to be? Will Azure reject times not within one day of now?
  Could `next_occurrence` simply being in the future be a better test?
- The fudge sounds more reasonable given how the Azure API works (and operations are not instantaneous). Adding a comment to explain it would be good.
- Should this really be >1 function? `next_occurrence` and `ensure_within_1d`?

jemrobinson:
> Why does the test assume `next_occurrence` will be within one day?

Just because that's the intention of the function (to generate the UTC timestamp corresponding to the next time it is e.g. 3am in Perth)

> Does it have to be? Will Azure reject times not within one day of now?

No, it doesn't have to be, but some of these schedulers won't start until the time occurs, so it would bad bad if this was e.g. 10 days in the future.

> Could next_occurrence simply being in the future be a better test?

Potentially, but I think there should be a maximum window on how far in the future it could be (see above)

> The fudge sounds more reasonable given how the Azure API works (and operations are not instantaneous). Adding a comment to explain it would be good.

Happy to remove this fudge (esp. given your comments above)

> Should this really be >1 function? next_occurrence and ensure_within_1d?

I think it only needs to be one function - the important thing is to come up with a UTC value that represents when your thing will happen in the future.

data_safe_haven/functions/strings.py
```diff
@@ -48,9 +48,11 @@ def next_occurrence(
...
         utc_dt = local_dt.astimezone(pytz.utc)
-        # Add one day until this datetime is in the future
-        utc_now = datetime.datetime.now(pytz.utc) + datetime.timedelta(minutes=1)
-        while utc_dt < utc_now:
+        # Add one day until this datetime is at least 1 hour in the future.
+        # This ensures that any Azure functions which depend on this datetime being in
+        # the future should treat it as valid.
+        utc_near_future = datetime.datetime.now(pytz.utc) + datetime.timedelta(hours=1)
+        while utc_dt < utc_near_future:
             utc_dt += datetime.timedelta(days=1)
```
JimMadge: I think this is a lot clearer :+1:.

JimMadge: Does the while loop ever cycle more than once?

tests/functions/test_strings.py
```diff
@@ -8,10 +8,22 @@
 from data_safe_haven.functions import next_occurrence, sanitise_sre_name
 
 
-def test_next_occurrence_is_within_next_day():
+def test_next_occurrence_will_occur_soon():
     dt_utc_now = datetime.datetime.now(datetime.UTC)
     next_time = next_occurrence(5, 13, "Australia/Perth")
-    dt_utc_tomorrow = dt_utc_now + datetime.timedelta(days=1, minutes=5)
+    dt_utc_tomorrow = dt_utc_now + datetime.timedelta(days=1, hours=2)
```
JimMadge: If you ran this at 23:00 wouldn't `dt_utc_tomorrow` be the day after tomorrow?

I think it is probably better to parameterise the tests over some set values, including edge cases near day boundaries.
That way the results and catching bugs doesn't depend on when the tests are run.

JimMadge (changes requested):

data_safe_haven/functions/strings.py
jemrobinson: I think not, but dates/times are hard! I was taking a "better safe than sorry" approach here. Happy to change to an `if` if you think that's better?

tests/functions/test_strings.py
jemrobinson: I guess that would require mocking `datetime.datetime.now()` inside the test? Or is there another way?

data_safe_haven/functions/strings.py
JimMadge: I know right :smile:.
I was trying to think of an edge case. Datetimes are always confusing. Let's leave it :+1:.

tests/functions/test_strings.py
JimMadge: Yes that sounds right.

I think the clever way to do it would be to make a fixture to mock `datetime.datetime.now()` which takes an argument that is the datetime to return. Then you could loop over that with `pytest.parametrize`.

jemrobinson:
Looks like [you can't mock builtin methods](https://stackoverflow.com/questions/23988853/how-to-mock-set-system-date-in-pytest). I'll have a go with `freezegun`.

tests/functions/test_strings.py
jemrobinson: Done with `freezegun` in 09af7930c

JimMadge (approved):
Added some extra tests with `@parametrize` and a test for DST.

PR #1894, opened 2024-05-16
Add pulumi tests

### LINKED ISSUE #1835: Implement Pulumi tests

With our infrastructure deployed by Pulumi, we should be able to write some infrastructure tests.

Look at the Pulumi testing docs: https://www.pulumi.com/docs/using-pulumi/testing/

Note [this issue](https://github.com/pulumi/examples/issues/1007) which implies it won't be trivial to add unit tests to an inline program.

jemrobinson:

n/a

Adds some proof-of-concept Pulumi unit tests for the SRE ApplicationGateway component.

**Note that much of what is described in the Pulumi documentation about Mocking does not seem to work with inline programs.**

Important notes:

- Pulumi `Outputs` should be tested within an `apply` loop with `run_with_unknowns` enabled, otherwise they will silently succeed on encountering any unknown values
- Value-based assertions need to be defined separately (currently in `resource_assertions.py`) as raw `assert` cannot be used inside an `apply` loop
- All tests must be decorated with `pulumi.runtime.test` or they will silently succeed
- Any sub-resource that we want to test needs to be exposed as an attribute of the Component class (e.g. `SREApplicationGateway` now has `self.application_gateway = ...` in its constructor so that we can test the ApplicationGateway resource.

Closes #1835

### DIFF
--- a/data_safe_haven/infrastructure/common/transformations.py
+++ b/data_safe_haven/infrastructure/common/transformations.py
@@ -13,7 +13,7 @@ def get_address_prefixes_from_subnet(subnet: network.GetSubnetResult) -> list[st
         return [str(p) for p in address_prefixes]
     if address_prefix := subnet.address_prefix:
         return [address_prefix]
-    msg = f"Subnet '{subnet.name}' has no address prefix."
+    msg = f"Subnet '{subnet}' has no address prefix."
     raise DataSafeHavenPulumiError(msg)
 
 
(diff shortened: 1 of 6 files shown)


JimMadge (approved):
Looks good.

Helpful as an example when we write more Pulumi tests in the future.

Having to `.apply()` does make tests a bit untidy but I suppose that is just what happens when working with promises.
If we wanted to do more "abstract" tests like

```
assert "Internet" not in sre_nsg.rules.allowed
```

we might have to write bespoke functions or lambdas in test functions.

jemrobinson:
@JimMadge I think the kind of abstract test you're thinking of could be better covered by [property tests](https://www.pulumi.com/docs/using-pulumi/testing/property-testing/) on deployed infrastructure.

PR #1913, opened 2024-05-21
Enable Azure api tests

JimMadge:

These tests were not running as the test file was not named like `test*.py`


jemrobinson (approved):
LGTM

PR #1920, opened 2024-05-29
Suppress warnings in config template

### LINKED ISSUE #1838: Pydantic warnings raised in dsh config template
When running `dsh config template` Pydantic raises warnings as some "help strings" are not expected types,

```
❯ dsh config template > config.yaml
/Users/jmadge/Library/Application Support/hatch/env/virtual/data-safe-haven/R3_ZGVTw/data-safe-haven/lib/python3.12/site-packages/pydantic/main.py:328: UserWarning: Pydantic serializer warnings:
  Expected `enum` but got `str` - serialized value may not be as expected
  Expected `bool` but got `str` - serialized value may not be as expected
  Expected `bool` but got `str` - serialized value may not be as expected
  return self.__pydantic_serializer__.to_python(
```

We could,

- suppress the warnings
- use the expected types then add comments another way

JimMadge:

Closes #1838

### DIFF
--- a/data_safe_haven/commands/config.py
+++ b/data_safe_haven/commands/config.py
@@ -22,6 +22,9 @@ def template(
 ) -> None:
     """Write a template Data Safe Haven configuration."""
     config = Config.template()
+    # The template uses explanatory strings in place of the expected types.
+    # Serialisation warnings are therefore suppressed to avoid misleading the users into
+    # thinking there is a problem and contaminating the output.
     config_yaml = config.to_yaml(warnings=False)
     if file:
         with open(file, "w") as outfile:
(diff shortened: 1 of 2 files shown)


data_safe_haven/commands/config.py
```diff
@@ -22,11 +22,12 @@ def template(
 ) -> None:
     """Write a template Data Safe Haven configuration."""
     config = Config.template()
+    config_yaml = config.to_yaml(warnings=False)
```
jemrobinson, line 28: Possibly worth commenting on *why* we want to disable warnings here?

jemrobinson (approved):

data_safe_haven/commands/config.py
JimMadge: Fixed in 8b61e7b

PR #1921, opened 2024-05-30
Validate SHM/SRE for user commands

### LINKED ISSUE #1879: Validate SRE name for non-pulumi commands
The `ProjectManager` class ensures that user-provided names for SHMs and SREs are valid (they exist if they need to). There are functions which use SRE names without checking, for example user management commands.

JimMadge:

Add checks that Pulumi configuration for SHM/SREs exists.
This is a proxy for ensuring the infrastructure exists.
It should be more reliable that just checking if the names are used in the DSH configuration, and much lighter weight than checking for particular pieces of infrastructure (either using Azure API or looking through the Pulumi stack).

Closes #1879

### DIFF
--- a/data_safe_haven/commands/users.py
+++ b/data_safe_haven/commands/users.py
@@ -32,6 +32,11 @@ def add(
 
     shm_name = context.shm_name
 
+    logger = LoggingSingleton()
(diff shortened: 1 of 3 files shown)


craddm (approved):
LGTM

PR #1936, opened 2024-06-14
Logging

### LINKED ISSUE #1657: Restructure entrypoint and exception/log interaction

Currently, the entrypoint catches `DataSafeHaven*` exceptions at the top level

https://github.com/alan-turing-institute/data-safe-haven/blob/653e86d4294a9290b7b168e12441c753347736c0/data_safe_haven/cli.py#L88-L94

There are a few problems with this,

- Errors are not written to stdout/stderr until the exceptions are raised back to this level. We cannot test the output in any Typer instance
- This try block stops the CLI from returning a non-zero exit code for any `DataSafeHaven*` exception

### LINKED ISSUE #1891: Stdout, stderr and log usage
From [PR discussion](https://github.com/alan-turing-institute/data-safe-haven/pull/1881#discussion_r1600036932).

We should decide how and when to use stdout and logging.
Logging may (or may not) be directed to stdout by default, but could also be direct to file.

Historically logging has been used heavily and printing avoided.

> In general, I think the only time we should be writing directly to stdout is if we're in a command that is intended to be used as part of a pipe/redirect.

> I think that is incorrect and goes against convention.
> 
> I think of stdout as information for the user. Logs are for posterity, for the user/admin.
`pacman -Syu` for example uses stdout to show the user package being downloaded and installed, and hooks being run. You wouldn't pipe that to anything else. Logs of transactions are also written to logs in `/var/logs/pacman.log`.
> 
> Or, a more common/mundane example, you would rarely pipe `<command> --help` to another command, but it would be very odd to not send that to stdout.

My suggestion would be,

- Information the user should know goes to stdout.
  - What the program is doing (at a high level)
  - Prompts for user input
  - Output of CLI commands
- stderr
  - Specifically for error messages (for convention and to help split from stdout)
- Information useful for auditing and debugging goes to logs.
  - What the program is doing at different levels (INFO to DEBUG)
  - Record of decisions
- Some things will go to both logs and stdout/stderr, although output for the user doesn't need to be in log format.
  - What the program is doing at a high level
  - Warnings
  - Errors

JimMadge:

- Logging changes
  - Simplify handler/formatter construction
  - Remove log level, file, line number from console handler
- Move unrelated methods from logging class
- Remove top level try/except
- More Pythonic use of logging module (makes configuration neater, particularly in unit tests)

The approach here is

- ConsoleHandler: Logging to console, information for the user, INFO and above
- PlainFileHandler: Logging to a file with rich markup removed, information for auditing and debugging, all levels

with logging levels

- rich.print: When output is only of interest to the user. Used for commands which send their output to stdout.
- logger.info: Used for notifications of actions taken and completed successfully. stdout which is also useful to log.
- logger.warning: Used for unexpected events or potential problems that won't cause the program to stop. User should take notice even if the program continues.
- logger.error: Used for unexpected events or problems that may cause the program to stop.
- logger.fatal: Used for problems that will stop the program, immediately before an exit call.

Closes #1891 
Closes #1657


jemrobinson:
Something to consider (no problem if you decide not to) - should we log all runs to `$appdirs / $date-$user-something.log` and potentially also upload these to Azure for future reference?

data_safe_haven/commands/cli.py
```diff
@@ -96,9 +97,4 @@ def callback(
 
 def main() -> None:
     """Run the application and log any exceptions"""
```
jemrobinson: ... and don't log exceptions? :)

data_safe_haven/exceptions/__init__.py
```diff
@@ -1,5 +1,12 @@
+from data_safe_haven.logging import get_logger
+
+
 class DataSafeHavenError(Exception):
-    pass
+    def __init__(self, message: str | bytes):
+        super().__init__(message)
+
+        logger = get_logger()
+        logger.error(message)
```
jemrobinson: What does this do?

data_safe_haven/logging/logger.py
```diff
@@ -0,0 +1,93 @@
+"""Standalone logging class implemented as a singleton"""
```
jemrobinson: This doesn't describe this file any more.

data_safe_haven/utility/console.py
```diff
@@ -0,0 +1,25 @@
+from rich import print as rprint
```
jemrobinson, line 1: As this is still to do with printing information to screen, I think it could/should live in the logging module. Happy if you think that means we need to change the module name to something broader.

data_safe_haven/utility/prompts.py
```diff
@@ -0,0 +1,14 @@
+from rich.prompt import Confirm
```
jemrobinson, line 1: As above, I'd like to see this live with the other logging commands.

data_safe_haven/commands/cli.py
```diff
@@ -1,13 +1,12 @@
 """Command line entrypoint for Data Safe Haven application"""
 
-import pathlib
 from typing import Annotated, Optional
 
 import typer
+from rich import print as rprint
```
jemrobinson, line 6: I think we should replace this with a `console.print` function that passes its args through to `rich.print`. That would simplify our include structure and unify the things that write straight to the console.

jemrobinson (approved):
Overall, LGTM. Can you post some examples of what the output looks like before and after for places where it's changed?

data_safe_haven/utility/console.py
JimMadge: I could see a module for "console stuff" like formatting text and prompts.

I think that is quite different from logging though and we should keep them separate.

I will leave it as is for this PR.

data_safe_haven/commands/cli.py
JimMadge: Do you mean [`rich.console.Console.print`](https://rich.readthedocs.io/en/stable/reference/console.html#rich.console.Console.print) or making something like `data_safe_haven.console.print`?

That could work with the other two comments here, introducing a `data_safe_haven.{console,output,terminal,text}` module to hold these things.

JimMadge:
Merged through docs link check failing as that wasn't modified here. Does look like a genuine failure though.

JimMadge:
Screenshots,

- New with `--verbose --show-level`
- New (default)
- Old

Screenshot 2024-06-20 at 11 15 26

JimMadge:
And the log file,

```
2024-06-20 11:13:47,048 - DEBUG - Reading project settings from '/Users/jmadge/Library/Application Support/data_safe_haven/contexts.yaml'.
2024-06-20 11:13:47,067 - INFO - Switched context to 'elements'.
2024-06-20 11:13:47,073 - DEBUG - Saved context settings to '/Users/jmadge/Library/Application Support/data_safe_haven/contexts.yaml'.
2024-06-20 11:14:06,576 - DEBUG - Reading project settings from '/Users/jmadge/Library/Application Support/data_safe_haven/contexts.yaml'.
2024-06-20 11:14:06,586 - INFO - Switched context to 'elements'.
2024-06-20 11:14:06,595 - DEBUG - Saved context settings to '/Users/jmadge/Library/Application Support/data_safe_haven/contexts.yaml'.
```

jemrobinson:
Looks great. I like the idea of keeping timestamps in the log file but not in the console output (by default). Do you think there's any benefit in adding an option to include timestamps in the console output?

JimMadge:
> Looks great. I like the idea of keeping timestamps in the log file but not in the console output (by default). Do you think there's any benefit in adding an option to include timestamps in the console output?

That did occur to me when I posted the pictures. Could definitely do that too, should be an easy addition.

PR #1948, opened 2024-06-19
Add console module

### LINKED ISSUE #1945: Console module
See https://github.com/alan-turing-institute/data-safe-haven/pull/1936#discussion_r1644948923, https://github.com/alan-turing-institute/data-safe-haven/pull/1936#discussion_r1644952814

Add a package to hold functions related to writing to and formatting for stdout,

- pretty printing (using rich)
- formatting tables, lists, _etc._
- user prompts

This is currently in modules in the utility package and handled by a `from rich import print as rprint` in each file that needs it.

JimMadge:
Closes #1945


jemrobinson:
I agree that logging and writing to console are distinct actions, but I was wondering whether there's a sufficiently broad module name that could cover them both. They're more similar to one another than e.g. logging is to deploying Pulumi infrastructure. What do you think @JimMadge?

JimMadge:
I don't like the idea of both in a module at the same level, like `output.logger.info()` and `output.pretty_print`. Having nested modules, like `output.logging.logger` and `output.console.pretty_print`, seems over complicated.

jemrobinson:
I guess I was imagining something like:

```
from data_safe_haven.output import get_logger, console

logger = get_logger()
logger.info("message")
console.print("message")
```

to replace

```
from data_safe_haven.logging import get_logger
from data_safe_haven.console import console

logger = get_logger()
logger.info("message")
console.print("message")
```

Not a big difference, but might make the code a bit easier to understand for new developers.

data_safe_haven/commands/cli.py
```diff
@@ -60,7 +60,7 @@ def callback(
         show_console_level()
 
     if version:
-        rprint(f"Data Safe Haven {__version__}")
+        pretty_print(f"Data Safe Haven {__version__}")
```
jemrobinson: As noted above I think I prefer something like:

```
from data_safe_haven.output import console
...
console.pretty_print(f"Data Safe Haven {__version__}")
```

which makes it clear at the point-of-use that this is being sent to the console.

data_safe_haven/commands/pulumi.py
```diff
@@ -76,4 +76,4 @@ def run(
         )
 
     stdout = project.run_pulumi_command(command)
-    rprint(stdout)
+    pretty_print(stdout)
```
jemrobinson: Can we stream this output or are we blocked until the command finishes?

data_safe_haven/console/pretty.py
```diff
@@ -0,0 +1,42 @@
...
+    overflow: OverflowMethod | None = None,
+    no_wrap: bool | None = None,
+    emoji: bool | None = None,
+    markup: bool | None = None,
+    highlight: bool | None = None,
+    width: int | None = None,
+    height: int | None = None,
+    crop: bool = True,
+    soft_wrap: bool | None = None,
+    new_line_start: bool = False,
```
jemrobinson: Do we actually want to use any of these options in our code? If not, why are we exposing them as arguments to `pretty_print`?

jemrobinson (changes requested):

data_safe_haven/commands/pulumi.py
JimMadge: I can have a look, I would guess we shouldn't need to worry about that because the OS(?) handles flushing to stdout. I hope `print` returns whether or not the output has been written.

data_safe_haven/console/pretty.py
JimMadge: We do use `sep`.

Seemed like I might as well make it API compatible. Though, if you are doing that you might as well just import the `Console` instance from this module and use `console.print` directly.

I was initially making _our_ pretty printing function, so that if we move away from rich the internal API doesn't change. I don't think we would want to implement all of those features if we did move from rich though.

JimMadge:
I also should add tests for the parts that aren't covered given that I have had a look at them here.

JimMadge:
@jemrobinson I'm still not keen on collecting logging and printing in one module.

Would something like

```python
from data_safe_haven.logging import get_logger
from data_safe_haven.console import console, prompts

logger = get_logger()
logger.info("hello")
console.print("hello")  # console is a rich.console.Console
prompts.confirm("yes or no?")
```

be better?

jemrobinson:
I think that looks OK (did you mean to import `prompts`? I think you wouldn't need to in this case). Essentially, I think it would be helpful to have `console.method` calls which look like `logger.method` calls so you can easily see "ah, this is being logged but this is going straight to the console".

N.B. I don't mean that we need `console.info` etc., `console.print` is fine.

JimMadge:
Oh, I made a mistake, I've updated the code block.

jemrobinson:
OK, what do you think of:

```python
from data_safe_haven.logging import get_logger
from data_safe_haven.console import console

logger = get_logger()
logger.info("hello")
console.print("hello")
console.confirm("yes or no?")
```

instead?

JimMadge:
Yes we can do that. The only thing that is a little annoying about that is we are making a wrapper for `rich.Console.print` instead of just exposing a `rich.Console`. Not a big deal though and it does make the API much cleaner.

jemrobinson:
I agree, it's a bit annoying - maybe we can use this chance to provide an adaptor for the `console.print` arguments that fits our needs better (e.g. we probably don't use all the arguments at the moment)?

jemrobinson (approved):
LGTM. Does it work? :)

JimMadge:
Yes, as far as I can tell.

I'm still not sure what is happening in #1954. I can't reproduce it with `dsh config upload ...`.

PR #1949, opened 2024-06-19
Remove broken link to git cheat sheet

JimMadge:

Removes a broken link, pointing to a Git cheat sheet.
There is another link to a similar resource so I think this should be fine.

I also tidied up the sentence, it looked like there may have been a third example in the past.


jemrobinson (approved):
LGTM

PR #1963, opened 2024-06-25
Merge SHM and context resource deployment

### LINKED ISSUE #1850: SHM and SRE should not be allowed to share a name

- **Operating System:** macOS
- **Data Safe Haven version:** develop @ c1d0adc

If an SHM and SRE share a name then the keys in `pulumi.yaml` will be clobbered.

This happens because the YAML file does not distinguish between SHM and SRE projects, e.g.

```
projects:
  apple:
    encrypted_key: <key>
    stack_config:
      azure-native:location: uksouth
      ...
  green:
    encrypted_key: <key>
    stack_config:
      azure-native:location: uksouth
      ...
```

<details>

```none
Your log details here
```

</details>

## :recycle: To reproduce

- Deploy an SHM
- Attempt to deploy an SRE with the same name

### LINKED ISSUE #1896: Improve deployment workflow usability

We should ensure that our deployment workflow maximises usability for people who are new to the project.

Our current deployment workflow looks like this:

- `dsh context add <KEY> --admin-group <group name> --location <location> --name <human friendly name> --subscription <Azure subscription name>`
  - this creates a local context file (but nothing else)
  - I'm not sure it's clear that's all this command is doing
  - Are command line switches the best way to do this, or should it use a template (see #1720)?
- `dsh context create`
  - it's slightly unclear whether a Context is the textfile that describes which Pulumi backend you're using or the infrastructure deployed to use it. Perhaps it's both, but we should either separate the two or make it clear that they're the same thing
  - why `create` not `deploy` here (as for `sre` and `shm`)?
- `dsh config template --file config.yaml`
  - consider using e.g. copier (see #1720)
  - consider separating SHM and SRE configs (see #1777)
- `dsh config upload config.yaml`
  - the reason why the config needs to be uploaded is a bit opaque
  - it might be better to combine this with the next command by adding an optional `--config` argument to it
- `dsh shm deploy`
  - this command is simple but it's the first time we've mentioned the `shm` (all the previous steps were about a generic `config`)
  - as above, I suggest allowing an optional config argument: `dsh shm deploy --config config.yaml`
- `dsh sre deploy <name of your SRE>`
  - as above, I suggest allowing an optional config argument: `dsh shm deploy --config config.yaml`

There are a lot of steps in the above workflow, some of which are relatively mechanical and some of which we could give more help/guidance with.

Suggestion:

- `dsh context add` -> `dsh context add/create` but using e.g. copier to provide better templating
- ~`dsh context create`~ -> add a check in the deployment commands so that this is lazily created "on demand"
- `dsh config template --file config.yaml` use e.g. copier for templating
  - potentially this could also be the place where a context is created?
- ~`dsh config upload config.yaml`~ -> drop this
- `dsh shm deploy` -> `dsh shm deploy --config config.yaml`
- `dsh sre deploy <name>` -> `dsh shm deploy <name> --config config.yaml`

### LINKED ISSUE #1898: Explain difference between Context key and name
In `dsh context add` we should explain the difference between key and name and whether you'd like them to be the same or different.

```
╭─ Arguments ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ *    key      TEXT  Key of the context to add. [default: None] [required]                                                                                                                                                                                                  │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Options ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ *  --admin-group           TEXT  The ID of an Azure group containing all administrators. [default: None] [required]                                                                                                                                                        │
│ *  --location              TEXT  The Azure location to deploy resources into. [default: None] [required]                                                                                                                                                                   │
│ *  --name                  TEXT  The human friendly name to give this Data Safe Haven deployment. [default: None] [required]                                                                                                                                               │
│ *  --subscription          TEXT  The name of an Azure subscription to deploy resources into. [default: None] [required]                                                                                                                                                    │
│    --help          -h            Show this message and exit.                                                                                                                                                                                                               │
╰──────────────────────────────────────────────────────────────────────────────
```

### LINKED ISSUE #1900: Remove SHM

There is very little currently left in the SHM and we should consider removing it completely in the interest of simplicity.

Migrate remaining components to the SRE and/or Context.

### LINKED ISSUE #1962: Replace admin_group_id with admin_group_name

If possible, it would be nice to replace the opaque `admin_group_id` parameter with a more human-readable `admin_group_name` parameter.

This is likely to involve some AzureCLI/GraphAPI calls.

### LINKED ISSUE #1965: How to handle config names/ids

Several of our config files share a pattern where they need a human-readable/writeable name and a YAML/JSON-safe ID. We should standardise how this is done.

### Requirements

- **Human name**
  - Supports a reasonably-size character set (at least numbers, letters, punctuation and whitespace)
  - Can be reasonably long if desired
  - Can be written into YAML/JSON as a value (not necessarily as a key)
  - Use cases:
    - log messages
    - tags on Azure resources

- **Config ID**
  - Safe for use as a YAML/JSON key (restricted character set probably without whitespace or punctuation)
  - Should be short (probably max 8 characters)
  - (Ideally) easily generated from the human name
  - Collisions between names unlikely
  - Use cases:
    - keys in YAML/JSON config files
    - providing unique names for Azure resources

### Conclusion
- Our existing `json_safe` function fails to meet the requirement to avoid collisions
- If we want to stick with generating the ID from the human name we should probably use some kind of compact hash-function (or a truncation of a hash)
- If we want the ID to be user-provided we should explain why

### Outstanding questions
- Which characters should be allowed in human-name?
- What maximum length should human-name have?
- Which hash function should we use?

jemrobinson:

n/a

This PR moves SHM deployment out of Pulumi and into AzureSDK commands. It also refactors the CLI interface so only the following commands are needed to deploy and SHM and SREs

```bash
$ dsh context add --admin-group-name $GROUP_NAME --description $HUMAN_FRIENDLY_NAME --name $SAFE_NAME --subscription $SUBSCRIPTION_NAME
$ dsh shm deploy --entra-tenant-id $TENANT_ID  --fqdn $FQDN  --location $LOCATION
$ dsh config template --file $YAML_FILE_PATH
$ dsh config upload $YAML_FILE_PATH
$ dsh sre deploy $SRE_NAME
```

It moves some config values between `Context` and `SHMConfig` so that fewer things depend on the (local-only) `Context` and more things use the (shared) `SHMConfig`. The things remaining in `Context` are all human-readable (no GUIDs)

#### Config examples

```
$ dsh context show
Current context: contoso
        Admin group name: Safe Haven Test Admins
        Description: DSG 2024 - Contoso
        Name: contoso
        Subscription name: Data Safe Haven Development
```

```
$ dsh config show-shm
azure:
  location: uksouth
  subscription_id: <removed GUID>
  tenant_id: <removed GUID>
shm:
  admin_group_id: <removed GUID>
  entra_tenant_id: <removed GUID>
  fqdn: green.develop.turingsafehaven.ac.uk
```

```
$ dsh config show apple
azure:
  location: uksouth
  subscription_id: <removed GUID>
  tenant_id: <removed GUID>
description: Apple Project
name: apple
sre:
  admin_email_address: safehavendevs@turing.ac.uk
  admin_ip_addresses:
  - <removed IP address>
  - <removed IP address>
  - <removed IP address>
  - <removed IP address>
  data_provider_ip_addresses:
  - <removed IP address>
  - <removed IP address>
  - <removed IP address>
  databases:
  - postgresql
  remote_desktop:
    allow_copy: true
    allow_paste: true
  research_user_ip_addresses:
  - <removed IP address>
  - <removed IP address>
  - <removed IP address>
  software_packages: pre-approved
  timezone: Europe/London
  workspace_skus:
  - Standard_D4s_v5
```

Closes #1850
Closes #1896
Closes #1898
Closes #1900
Closes #1962
Closes #1965 

### :microscope: Tests

Tested locally


data_safe_haven/config/shm_config.py
```diff
@@ -15,15 +16,44 @@ class SHMConfig(AzureSerialisableModel):
     azure: ConfigSectionAzure
     shm: ConfigSectionSHM
 
+    @classmethod
+    def from_local(
```
JimMadge:
```suggestion
    def from_args(
```

data_safe_haven/config/shm_config.py
```diff
@@ -15,15 +16,44 @@ class SHMConfig(AzureSerialisableModel):
...
+        cls: type[Self],
+        context: ContextBase,
+        *,
+        entra_tenant_id: str,
+        fqdn: str,
+        location: str,
+    ) -> SHMConfig:
+        """Construct an AzureSerialisableModel from a YAML file in Azure storage."""
```
JimMadge:
```suggestion
        """Construct an SHMConfig from arguments."""
```

data_safe_haven/infrastructure/programs/imperative_shm.py
```diff
@@ -80,7 +85,59 @@ def create(self) -> None:
                 key_vault_name=keyvault.name,
             )
         except DataSafeHavenAzureError as exc:
-            msg = "Failed to create context resources."
+            msg = f"Failed to deploy resources needed by Pulumi.\n{exc}"
```
JimMadge:
```suggestion
            msg = f"Failed to deploy resources needed by Pulumi."
```

data_safe_haven/infrastructure/programs/imperative_shm.py
```diff
@@ -80,7 +85,59 @@ def create(self) -> None:
...
+                record_flags=0,
+                record_name="@",
+                record_tag="issue",
+                record_value="letsencrypt.org",
+                resource_group_name=resource_group.name,
+                ttl=3600,
+                zone_name=self.config.shm.fqdn,
+            )
+        except DataSafeHavenAzureError as exc:
+            msg = f"Failed to create SHM resources.\n{exc}"
```
JimMadge:
```suggestion
            msg = f"Failed to create SHM resources."
```

data_safe_haven/config/context_manager.py
```diff
@@ -91,50 +90,49 @@ def available(self) -> list[str]:
...
         *,
-        key: str,
+        admin_group_name: str,
         name: str,
-        admin_group_id: str,
-        location: str,
         subscription_name: str,
     ) -> None:
         # Ensure context is not already present
+        key = json_safe(name)
```
JimMadge: I don't like taking user input and then modifying it opaquely where two different inputs can give the same output, like here.

If we want to only have one, I think it would be better to have a validator ensuring that `name == json_safe(name)`.

data_safe_haven/external/api/azure_cli.py
```diff
@@ -81,3 +81,25 @@ def confirm(self) -> None:
...
+
+        try:
+            result_dict = json.loads(result)
+            return str(result_dict[0]["id"])
+        except json.JSONDecodeError as exc:
+            msg = f"Unable to parse Azure CLI output as JSON.\n{result}"
+            raise DataSafeHavenAzureError(msg) from exc
+        except (IndexError, KeyError) as exc:
+            msg = f"Group '{group_name}' was not found in Azure CLI."
+            raise DataSafeHavenAzureError(msg) from exc
```
JimMadge, line 105: This feels fairly complex/fragile.

I'm not sure the gain of giving a name rather than id is worth it.

data_safe_haven/commands/shm.py
```diff
@@ -4,114 +4,144 @@
...
+from data_safe_haven.validators import (
+    typer_aad_guid,
+    typer_fqdn,
+)
 
 shm_command_group = typer.Typer()
 
 
 @shm_command_group.command()
 def deploy(
```
JimMadge, line 25: This command is doing a lot of work,

- Create or fetch SHM config
- Update SHM config with write protection
- Deploy SHM

It breaks the convention we had before of `context add/config upload` and `deploy`.

Do I understand correctly that before `dsh shm deploy` there is no where to store config?
How likely is it that someone will want to modify an already deployed SHM?

JimMadge (changes requested):
Looks good.

I have some questions, and there are a few pieces I'm not very keen on.

I think there is scope to improve the coverage in `data_safe_haven/commands/` and `data_safe_haven/config/`.

data_safe_haven/commands/shm.py
jemrobinson: Yes, it's doing all those things. This was in the interest of simplification - there's very little user-editable content in the SHMConfig (most of it is GUIDs which we've often derived from the human-readable version) so the model of "create template" -> "edit template" -> "upload template" -> "deploy from remote config" seems like overkill here.

Would you prefer to see this command broken up? If so, how?

data_safe_haven/config/context_manager.py
jemrobinson: Sure - that makes sense. I think we could expand the definition of JSON-safe to include other characters (underscores and hyphens would make sense - perhaps others too?).

data_safe_haven/commands/shm.py
```diff
@@ -4,114 +4,144 @@
...
+            logger.critical(
+                "You must provide the --fqdn argument when first deploying an SHM."
+            )
+            raise typer.Exit(1)
+        if not location:
+            logger.critical(
+                "You must provide the --location argument when first deploying an SHM."
+            )
+            raise typer.Exit(1)
+        config = SHMConfig.from_local(
```
jemrobinson:
```suggestion
        config = SHMConfig.from_args(
```

data_safe_haven/external/api/azure_cli.py
jemrobinson: I'm not sure it's as fragile as it looks. I don't think the JSONDecodeError is very likely, for example, so could be merged with the IndexError and KeyError catch.

IMO, seeing "DSH Test Admins" in the Context file is more readable/understandable than seeing the GUID and having to lookup whether it's the correct one.

data_safe_haven/config/context_manager.py
jemrobinson: This is actually more involved than I thought at first. I suggest leaving this PR as-is and discussing further on #1965.

data_safe_haven/external/api/azure_cli.py
jemrobinson:
```suggestion
    def group_id_from_name(self, group_name: str) -> str:
        """Get ID for an Entra ID group that this user is permitted to view."""
        try:
            result = subprocess.check_output(
                [self.path, "ad", "group", "list", "--display-name", group_name],
                stderr=subprocess.PIPE,
                encoding="utf8",
            )
            result_dict = json.loads(result)
            return str(result_dict[0]["id"])
        except (subprocess.CalledProcessError, json.JSONDecodeError, IndexError, KeyError) as exc:
            msg = f"Group '{group_name}' was not found in Azure CLI."
            raise DataSafeHavenAzureError(msg) from exc
```

data_safe_haven/config/context_manager.py
JimMadge: We can't easily undo the changes here?

data_safe_haven/external/api/azure_cli.py
JimMadge: I think I prefer the original, distinguishes between "Azure CLI didn't give me JSON" and "The thing you are looking for doesn't exist".

We can leave it if we think the benefit is there.

In general, I don't like the subprocess calls as we have to support ABI compatibility with Azure CLI. This feels like a place where we don't need one because the group name and id have the same meaning.

data_safe_haven/commands/shm.py
JimMadge: I think what I was worried about was, if you need to redeploy an SHM, is there a bigger risk of inadvertently overwriting settings or needing to share information between admins.

However, I think the check for an existing config should cover that. Is that true?

Given there is a class for `SHMConfig` that supports serialisation, we could do something like

```
dsh config shm-template
# edit template
dsh shm deploy shm.yaml
```

As I understand it, until you deploy there isn't anywhere to upload the config to anyway.

JimMadge:
@craddm Would appreciate your thoughts on the conversations here.

data_safe_haven/external/api/azure_cli.py
craddm: I personally prefer the name rather than the GUID, as it's much more obvious to the user if it's incorrect

data_safe_haven/commands/shm.py
jemrobinson: > we could do something like
> ```
> dsh config shm-template
> # edit template
> dsh shm deploy shm.yaml
> ```

The main issue with this is that SHMConfig now has 6 fields, 4 of which we don't want the user to edit:

```yaml
azure:
  location: uksouth  # user-editable
  subscription_id: <GUID from lookup>
  tenant_id: <GUID from lookup>
shm:
  admin_group_id: <GUID from lookup>
  entra_tenant_id: <GUID from lookup>
  fqdn: green.develop.turingsafehaven.ac.uk  # user-editable
```

I'm not sure how much the user will benefit from editing these two fields in a config file (while remembering not to edit the other ones) rather than providing them at the command line.

If you feel strongly that this is better, I'm happy to implement it.

> As I understand it, until you deploy there isn't anywhere to upload the config to anyway.

Yes, that's right. What happens at present is that the `SHMConfig` is held in-memory until the storage account has been deployed and is then uploaded.

data_safe_haven/external/api/azure_cli.py
jemrobinson: OK, let's leave this as it is then

data_safe_haven/config/context_manager.py
jemrobinson: I can undo the changes which removed the user-provided "key" argument when storing Contexts in YAML, but the wider issue of #1965 is that both the SREConfig and the SHMConfig share a pattern where they want a human-readable name in some circumstances and a YAML/JSON-friendly one in others.

Undoing the changes here still leaves that problem to be solved.

data_safe_haven/commands/shm.py
JimMadge: :+1:

Sounds good. Happy to leave this. We can always change later if we feel it is a problem.

data_safe_haven/config/context_manager.py
jemrobinson: Done in 0f5e6c4

jemrobinson:
@JimMadge : Anything else you want to see here?

JimMadge (approved):

PR #1970, opened 2024-06-27
Remove unused code

jemrobinson:

- Remove unused `ConfigSection::update()` functions
- Remove unused `ConfigName` validation and tests
- Add tests for `ConfigSection` and `SafeString`
- Simplify `dsh sre deploy/teardown` logic
- BUGFIX: create DSHPulumiConfig in sre deploy or this will not be done

n/a

### :microscope: Tests

Tested locally


data_safe_haven/commands/sre.py
```diff
@@ -89,30 +93,27 @@ def deploy(
         )
 
         # Deploy Azure infrastructure with Pulumi
-        if force is None:
-            stack.deploy()
-        else:
+        try:
             stack.deploy(force=force)
```
JimMadge, line 97: I think this is neater.

```python
force: bool = False
```
Rather than
```python
force: bool | None = None
```

data_safe_haven/commands/sre.py
```diff
@@ -122,30 +123,28 @@ def teardown(
...
-        # part of a Pulumi declarative command
+        # Load GraphAPI as this may require user-interaction
         graph_api = GraphApi(
             tenant_id=shm_config.shm.entra_tenant_id,
             default_scopes=["Application.ReadWrite.All", "Group.ReadWrite.All"],
         )
 
+        # Load Pulumi and SRE configs
+        pulumi_config = DSHPulumiConfig.from_remote(context)
+        sre_config = SREConfig.from_remote_by_name(context, name)
```
JimMadge, line 138: I'm confused, did this stuff get removed somewhere?
It must have been there at some point 😕.

tests/config/test_config_sections.py
```diff
@@ -20,41 +20,64 @@ def test_constructor(self) -> None:
 
     def test_invalid_location(self):
         with pytest.raises(
-            ValidationError, match="Value error, Expected valid Azure location"
+            ValidationError, match=r"Value error, Expected valid Azure location"
```
JimMadge: Why does this need to be a raw string?

JimMadge (approved):
LGTM

Clears out a lot of code we don't need :tada:.

tests/config/test_config_sections.py
jemrobinson: It doesn't need to be - it just looks nicer in my syntax highlighter :). I'll change it back.

data_safe_haven/commands/sre.py
jemrobinson: It's just moved down from a few lines up.

data_safe_haven/commands/users.py
```diff
@@ -130,7 +130,7 @@ def register(
 
         # Load SREConfig
         sre_config = SREConfig.from_remote_by_name(context, sre)
-        sre_name = sre_config.safe_name
+        sre_name = sre_config.name
         if sre_name not in pulumi_config.project_names:
```
jemrobinson, line 134:
```suggestion
        if sre_config.name not in pulumi_config.project_names:
```

data_safe_haven/commands/users.py
```diff
@@ -130,7 +130,7 @@ def register(
...
             msg = f"Could not load Pulumi settings for '{sre_name}'. Have you deployed the SRE?"
```
jemrobinson, line 135:
```suggestion
            msg = f"Could not load Pulumi settings for '{sre_config.name}'. Have you deployed the SRE?"
```

tests/config/test_config_sections.py
jemrobinson:
```suggestion
            ValidationError, match="Value error, Expected valid Azure location"
```

PR #2002, opened 2024-07-08
Refactor Azure authentication classes

### LINKED ISSUE #1980: CLI commands hanging

- **Operating System:** macOS
- **Data Safe Haven version:** `develop @ 025f298`

## :package: Packages

<details>

```none
Paste list of packages here
```

</details>

Occasionally, running a CLI command will result in the app hanging without returning. This seems to be related to a non-returning `subprocess.check_output` command in AzureCliCredential.

<details>

```none
Your log details here
```

</details>

## :recycle: To reproduce

Raise an exception in `sre deploy` and the Pulumi upload step should hang.

### LINKED ISSUE #2001: Simplify classes that interact with Azure

Currently all of `AzureApi`, `AzureCliSingleton`, `GraphApi` and `AzureAuthenticator` interact with/authenticate against Azure in various ways. Additionally we have helper classes like `AzureCliAccount` and `LocalTokenCache` which are only used by these classes. We should simplify this interface.

jemrobinson:

n/a

- Reduce Azure-connected classes to two: `AzureSdk` and `GraphApi`
- Use our own `DeferredCredential` class which can cache tokens to reduce external API calls
  - use of cached tokens avoids the "stalling bug" which was otherwise triggered by making an AzureCLI call during the `finally` clause of a try statement
- `AzureSdkCredential` provides authentication for `AzureSdk` using the Azure CLI
- `GraphApiCredential` provides authentication for `GraphApi` using a browser device-code
- Fixes a bug in `http_get` where we were ignoring additional pages for paged responses.
- N.B. this removes the `confirm` dialogue which was applied for AzureCLI logins but not for Browser logins. If we want this, I can easily add it to both.

Closes #2001 
Closes #1980


data_safe_haven/external/api/azure_sdk.py
```diff
@@ -685,6 +717,19 @@ def get_storage_account_keys(
             msg = f"Keys could not be loaded for {msg_sa} in {msg_rg}."
             raise DataSafeHavenAzureError(msg) from exc
 
+    def get_subscription(self, subscription_name: str) -> Subscription:
+        """Get the current Azure subscription."""
```
jemrobinson:
```suggestion
        """Get an Azure subscription by name."""
```

data_safe_haven/external/api/credentials.py
```diff
@@ -0,0 +1,166 @@
...
+            self.logger.info(
+                "You are currently logged into the [blue]Azure CLI[/] with the following details:"
+            )
+            self.logger.info(
+                f"... user: [green]{decoded['name']}[/] ({decoded['oid']})"
+            )
+            self.logger.info(
+                f"... tenant: [green]{decoded['upn'].split('@')[1]}[/] ({decoded['tid']})"
+            )
+        except CredentialUnavailableError as exc:
```
jemrobinson, line 92:
```suggestion
        except (CredentialUnavailableError, jwt.exceptions.DecodeError, KeyError) as exc:
```

pyproject.toml
```diff
@@ -165,7 +167,7 @@ addopts = [
     "-vvv",
     "--import-mode=importlib",
     "--disable-warnings",
-    "--ignore=tests/*",
+    "--ignore=data_safe_haven/resources/*",
```
JimMadge, line 174:
```suggestion
    "--ignore=data_safe_haven/resources/*",
    "--omit=tests/*",
    "--omit=data_safe_haven/resources/*",
```

I think the idea here was to omit tests from the coverage.

data_safe_haven/external/api/credentials.py
```diff
@@ -0,0 +1,177 @@
...
+        tenant_id: str,
+        scopes: Sequence[str] = [],
+    ) -> None:
+        super().__init__(scopes=scopes, tenant_id=tenant_id)
+
+    def get_credential(self) -> TokenCredential:
+        """Get a new DeviceCodeCredential, using cached credentials if they are available"""
+        cache_name = f"dsh-{self.tenant_id}"
+        authentication_record_path = (
+            pathlib.Path.home() / f".msal-authentication-cache-{cache_name}"
```
JimMadge: Will this be the correct path for all OSs?

data_safe_haven/external/api/credentials.py
```diff
@@ -0,0 +1,177 @@
...
+        kwargs = {}
+        try:
+            with open(authentication_record_path) as f_auth:
+                existing_auth_record = AuthenticationRecord.deserialize(f_auth.read())
+                kwargs["authentication_record"] = existing_auth_record
+        except FileNotFoundError:
+            kwargs["authority"] = "https://login.microsoftonline.com/"
+            # Use the Microsoft Graph Command Line Tools client ID
+            kwargs["client_id"] = "14d82eec-204b-4c2f-b7e8-296a70dab67e"
+            kwargs["tenant_id"] = self.tenant_id
```
JimMadge:
```suggestion
if authentication_record_path.is_file():
    with open(authentication_record_path) as f_auth:
        existing_auth_record = AuthenticationRecord.deserialize(f_auth.read())
        kwargs["authentication_record"] = existing_auth_record
else:
    kwargs["authority"] = "https://login.microsoftonline.com/"
    # Use the Microsoft Graph Command Line Tools client ID
    kwargs["client_id"] = "14d82eec-204b-4c2f-b7e8-296a70dab67e"
    kwargs["tenant_id"] = self.tenant_id
```

data_safe_haven/external/api/credentials.py
```diff
@@ -0,0 +1,177 @@
...
+        new_auth_record = credential.authenticate(scopes=self.scopes)
+        with open(authentication_record_path, "w") as f_auth:
+            f_auth.write(new_auth_record.serialize())
+
+        # Write confirmation details about this login
+        self.logger.info(
+            "You are currently logged into the [blue]Microsoft Graph API[/] with the following details:"
+        )
+        self.logger.info(
+            f"... user: [green]{new_auth_record.username}[/] ({new_auth_record._home_account_id.split('.')[0]})"
```
JimMadge:
```suggestion
            f"\tuser: [green]{new_auth_record.username}[/] ({new_auth_record._home_account_id.split('.')[0]})"
```

data_safe_haven/external/api/credentials.py
```diff
@@ -0,0 +1,177 @@
...
+        )
+        self.logger.info(
+            f"... tenant: [green]{new_auth_record._username.split('@')[1]}[/] ({new_auth_record._tenant_id})"
```
JimMadge:
```suggestion
            f"\ttenant: [green]{new_auth_record._username.split('@')[1]}[/] ({new_auth_record._tenant_id})"
```

data_safe_haven/external/api/credentials.py
```diff
@@ -0,0 +1,177 @@
...
+        """Get a new AzureCliCredential."""
+        credential = AzureCliCredential(additionally_allowed_tenants=["*"])
+        # Check that we are logged into Azure
+        try:
+            decoded = self.decode_token(credential.get_token(*self.scopes).token)
+            self.logger.info(
+                "You are currently logged into the [blue]Azure CLI[/] with the following details:"
+            )
+            self.logger.info(
+                f"... user: [green]{decoded['name']}[/] ({decoded['oid']})"
```
JimMadge:
```suggestion
                f"\tuser: [green]{decoded['name']}[/] ({decoded['oid']})"
```

data_safe_haven/external/api/credentials.py
```diff
@@ -0,0 +1,177 @@
...
+            )
+            self.logger.info(
+                f"... tenant: [green]{decoded['upn'].split('@')[1]}[/] ({decoded['tid']})"
```
JimMadge:
```suggestion
                f"\ttenant: [green]{decoded['upn'].split('@')[1]}[/] ({decoded['tid']})"
```

data_safe_haven/external/api/credentials.py
```diff
@@ -0,0 +1,177 @@
...
+        # Require at least 10 minutes of remaining validity
+        validity_cutoff = datetime.now(tz=UTC).timestamp() + 10 * 60
+        if not DeferredCredential.token_ or (
+            DeferredCredential.token_.expires_on < validity_cutoff
+        ):
+            # Generate a new token and store it at class-level token
+            DeferredCredential.token_ = self.get_credential().get_token(
+                *scopes, **kwargs
+            )
+        return DeferredCredential.token_
```
JimMadge: We could use a [cached property](https://docs.python.org/3/library/functools.html#functools.cached_property) but perhaps this is better.

It will generate a new credential if it needs to?
Is the cut off 10 minutes or 10 hours?

data_safe_haven/external/api/credentials.py
```diff
@@ -0,0 +1,177 @@
...
+    def get_token(
+        self,
+        *scopes: str,
+        **kwargs: Any,
+    ) -> AccessToken:
+        # Require at least 10 minutes of remaining validity
+        validity_cutoff = datetime.now(tz=UTC).timestamp() + 10 * 60
+        if not DeferredCredential.token_ or (
+            DeferredCredential.token_.expires_on < validity_cutoff
+        ):
```
JimMadge, line 71: Is `expires_on` in UTC?

JimMadge (dismissed):
Looks good.

It isn't _simple_ but I think this makes better use of the classes the Azure SDK provides and simplifies the classes we manage.

I think we need tests for the new code in `external/api`

pyproject.toml
jemrobinson: Maybe I'm misunderstanding how `coverage` and `pytest` interact. We already have `coverage run --omit=tests/*` in the `test:test` command - I think these options are to `pytest` itself. What I'd like to do is exclude `data_safe_haven/resources/*` from the list of places to look for tests. Do you know the correct syntax for that?

data_safe_haven/external/api/credentials.py
jemrobinson: Good point. Let's use `config_dir` here.

data_safe_haven/external/api/credentials.py
jemrobinson: Yes, it's a UTC Unix timestamp in seconds. Some tokens also have "expiresOn" which isn't, but let's not worry about that!

data_safe_haven/external/api/credentials.py
```diff
@@ -0,0 +1,177 @@
...
+    def get_credential(self) -> TokenCredential:
+        """Get a credential provider from the child class."""
+        pass
+
+    def get_token(
+        self,
+        *scopes: str,
+        **kwargs: Any,
+    ) -> AccessToken:
+        # Require at least 10 minutes of remaining validity
```
jemrobinson, line 66:
```suggestion
        # Require at least 10 minutes of remaining validity
        # The 'expires_on' property is a Unix timestamp integer in seconds
```

data_safe_haven/external/api/credentials.py
jemrobinson: The underlying credentials themselves (`AzureCliCredential` and `DeviceCodeCredential` should be able to refresh themselves if needed). `DeviceCodeCredential` actually has token-caching built-in, but `AzureCliCredential` doesn't, and there's a note in the class definition that you should implement this yourself!

jemrobinson: The cut-off is 10 minutes (see explanatory comment I've added to the code).

data_safe_haven/external/api/credentials.py
jemrobinson: Done in e95a0d7a0

jemrobinson:
> I think we need tests for the new code in `external/api`

I have added tests for `entra_directory`, `subscription_id` and `tenant_id`. Which others would you like to see?

pyproject.toml
JimMadge: > We already have coverage run --omit=tests/* in the test:test command

Ah, maybe we should move it to the options here for consistency.
My comment was assuming removing the ignore for tests would mean they are no longer omitted. I didn't see the other omit argument.

AFAIK, ignore stops pytest from searching matching paths/files for test files (like `test*.py` and omit prevents pytest-cov from counting files in the coverage stats.
I think before everything in `tests/` was being included in the coverage and making it seem lower than it really was.

I think also, if you run `hatch run test:test` you will only run tests in `tests/` as that is the default argument. So "tests" in `data_safe_haven/resources/` should only be found if you run something like `hatch run test:test ./`

JimMadge:
> > I think we need tests for the new code in `external/api`
> 
> I have added tests for `entra_directory`, `subscription_id` and `tenant_id`. Which others would you like to see?

I think tests for the new/modified classes like `DeferredCredential`, `AzureSdkCredential` and `AzureApiCredential`. I think it is worth writing unit tests for their behaviour as relying on them being imported by other classes.

The tests are failing now so it is hard to see where the coverage is.

pyproject.toml
jemrobinson: I'm pretty sure that `--omit` is an argument to `coverage` (see [here](https://coverage.readthedocs.io/en/7.5.4/cmd.html#cmd-run)) and it's telling `coverage` to ignore coverage in the `tests` directory.

Similarly the `pytest` documentation [says that](https://docs.pytest.org/en/7.1.x/example/pythoncollection.html?highlight=ignore) `--ignore` is the correct way to indicate directories that you don't want to be searched when looking for tests.

I think this change is correct.

JimMadge: Yes, I think we are saying the same thing.

I'm suggesting that we have all ignores and omits in the same place,

```toml
addopts=[
    ...,
    "--ignore=data_safe_haven/resources/*",
    "--omit=tests/*",
    "--omit=data_safe_haven/resources/*",
]
```

Rather than some in the test script and some in the toml options.

JimMadge (commented):
Still a few gaps in [the coverage](https://github.com/alan-turing-institute/data-safe-haven/pull/2002#issuecomment-2214309234) in `credentials.py`, `graph_api.py` and `azure_sdk.py`.

Things like getting credentials from files and exception handling.

pyproject.toml
jemrobinson: Right, but the things in `addopts` are options for `pytest`. We need `--omit=tests/*` to be an option to `coverage` (so that coverage of the test folder doesn't count towards overall stats) and `--ignore=data_safe_haven/resources/*` to be an option to `pytest` (so that we don't try to run the smoke tests with pytest).

JimMadge: 🤦

JimMadge: I'll push changes to the correct place.

jemrobinson:
@JimMadge : I think this now tests all the functions under `external/api`. Are you happy to merge with these changes?

tests/conftest.py
```diff
@@ -33,12 +34,20 @@
 from data_safe_haven.logging import init_logging
 
 
+def pytest_configure():
+    """Define constants for use across multiple tests"""
+    pytest.guid_admin = "00edec65-b071-4d26-8779-a9fe791c6e14"
+    pytest.guid_entra = "48b2425b-5f2c-4cbd-9458-0441daa8994c"
+    pytest.guid_subscription = "35ebced1-4e7a-4c1f-b634-c0886937085d"
+    pytest.guid_tenant = "d5c5c439-1115-4cb6-ab50-b8e547b6c8dd"
```
JimMadge: I'd not seen `pytest_configure` before :rocket:.

It might not be a problem, but adding attributes to pytest feels a little odd.
Is there a reason to not use a fixture or fixtures?

JimMadge (commented):
Just one comment, could leave for another PR if we open an issue.

tests/conftest.py
jemrobinson: This line `.replace("guid_subscription", pytest.guid_subscription)` doesn't work with a fixture. The `replace` function complains that `guid_subscription` is a function, but `guid_subscription()` isn't treated correctly by `pytest`.

This is a pattern that I've seen recommended in a few places e.g. here (https://pytest-with-eric.com/hooks/pytest-configure/#Share-Global-Variables-with-pytest-configure).

I agree - it would be nicer to simply use test-suite-level variables, but these don't seem to work as expected.

JimMadge: 😆 I was looking at the exact same site. I almost put in my comment that adding the attributes to `config` rather than `pytest` feels _a bit_ more safe.

JimMadge (approved):

Issue #1997, opened 2024-07-08
Gitea mirror creation integration


JimMadge:
Integrate mirror creation in a user friendly way.

This could be,

- Minimal webapp for users to request repositories
- Minimal admin dashboard to view/approve requests
- Notification system (emails?)

jemrobinson:
Email is hard from inside Azure - you typically need a SendGrid subscription which you can then link to Azure.

JimMadge:
We could use Azure functions to run a simple script to create mirrors.

A HTTP trigger allows [passing variables as HTTP query parameters or in the body of a request](https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-http-webhook-trigger?tabs=python-v2%2Cisolated-process%2Cnodejs-v4%2Cfunctionsv2&pivots=programming-language-python#example).

JimMadge:
Looking at the Azure Functions docs. It looks like not all plans [support virtual networking](https://learn.microsoft.com/en-us/azure/azure-functions/functions-scale) which I think will be essential for us.

The Consumption plan does not support virtual networking, so I think that is out of the question.
Flex Consumption does, and has the nice feature of Consumption that it can scale to 0 instances and bills by usage, but is in preview.
Functions Premium and App Service (Dedicated Plan) both mean that at least 1 instance must always be deployed.
App Service would likely make sense if we also want to host webapps in the future as they can be hosted on the same resources.

App Service might be the best solution long term then, as we may want to make a little webapp for requesting/approving mirrors.

[Billing docs](https://learn.microsoft.com/en-us/azure/azure-functions/functions-scale#billing)

PR #2008, opened 2024-07-10
WIP: Add mirror creation script

### LINKED ISSUE #1996: Automate Gitea mirror creation
Create functions to automate our code ingress process, probably using [Gitea API](https://docs.gitea.com/api/1.22/)

JimMadge:

Add Azure function apps to create read only mirrors in Gitea.

Closes #1996
Related to #1997


JimMadge:
Running this locally

```command
❯ http localhost:7071/api/create-mirror address=https://github.com/JimMadge/dotfiles name=dotfiles username=jim password=password
HTTP/1.1 200 OK
Content-Type: text/plain; charset=utf-8
Date: Wed, 10 Jul 2024 14:24:09 GMT
Server: Kestrel
Transfer-Encoding: chunked

Mirror successfully created
```

And the function environment output

```
[2024-07-10T14:24:07.542Z] Executing 'Functions.create_mirror' (Reason='This function was programmatically called via the host APIs.', Id=20e523e8-9ad1-4368-889a-0b463d0c39ed)
[2024-07-10T14:24:07.549Z] Request received.
[2024-07-10T14:24:07.549Z] parameters: address=https://github.com/JimMadge/dotfiles, name=dotfiles, password=password, username=jim
[2024-07-10T14:24:07.550Z] Sending request to create mirror.
[2024-07-10T14:24:10.173Z] Response status code: 201.
[2024-07-10T14:24:10.173Z] Sending request to configure mirror repo.
[2024-07-10T14:24:10.293Z] Response status code: 200.
[2024-07-10T14:24:10.297Z] Executed 'Functions.create_mirror' (Succeeded, Id=20e523e8-9ad1-4368-889a-0b463d0c39ed, Duration=2755ms)
```

JimMadge:
[this](https://github.com/pulumi/examples/blob/master/azure-ts-functions/index.ts) looks like a good reference for setting up a function app using Pulumi.
Note that the the app deployment needs some information you might need to construct from the `Outputs` of the storage account.

data_safe_haven/infrastructure/programs/sre/apps.py
```diff
@@ -107,4 +117,65 @@ def __init__(
...
+            server_farm_id=app_service_plan.id,
+            site_config=web.SiteConfig(
+                app_settings=[
+                    {"name": "runtime", "value": "python"},
+                    {"name": "FUNCTIONS_WORKER_RUNTIME", "value": "python"},
+                    {"name": "WEBSITE_RUN_FROM_PACKAGE", "value": blob_url},
+                    {"name": "FUNCTIONS_EXTENSION_VERSION", "value": "~4"},
+                ],
+            )
+        )
```
jemrobinson: Is there a reason we prefer the gitea mirror to run as a webapp rather than an ACI? Is it cheaper/easier/better in some other way to do this?

JimMadge: This part isn't the Gitea instance, it is a webapp that uses the Gitea API to create/delete mirrors.

I am hoping that using the Azure Functions framework we can avoid the complication of managing flask/fastapi/_etc._ and use Entra ID authentication.

JimMadge: Unless there is a reason not to, I think we should deploy all Gitea instances the same way.

JimMadge: Looking that that now, there might be a fair bit of work to either generalise the Gitea configuration or make a new Pulumi ComponentResource.

I think we will want the mirror instance to be fairly different,

- No login
- Browseable without login
- No registration
- No LDAP authentication

JimMadge:
Resources deploy. The webapp service is not able to fetch the app from storage though.

Using the connection url + SAS token gives,

```
error: ContentDecodingError: ('Received response with content-encoding: deflate, but failed to decode it.', error('Error -3 while decompressing data: invalid stored block lengths'))
```

Possibly a problem with the construction of that string, or the way data is uploaded to storage.

JimMadge:
Although, `wget` is able to fetch the zip for the SAS URL 😕.

JimMadge:
The connection string was needed to fetch and unpack the Zip from blob storage.
Pulumi ts examples showed how to do this.

Now there seems to be some trouble identifying the endpoints, so I'm taking a step back and looking at examples to see what needs to be done.

JimMadge:
The problem was likely using the v2 Python programming model, where as examples use the v1 model.

I couldn't find clear documentation on how to use the v2 model outside of deploying functions using Azure CLI, or examples of using the v2 model with Pulumi. Using the v1 model might be the most sensible thing for now.

JimMadge:
Following the v1 model directory structure means the functions get recognised and registered.

PR #2009, opened 2024-07-12
Clean Pulumi state during cleanup

### LINKED ISSUE #1976: Cleanup partially-constructed resources

If a Pulumi operation is interrupted (e.g. with `Ctrl+C`) and then reattempted, it will fail, due to partially constructed resources in the Pulumi state file.

Cleaning these up is complicated, especially since we can't run interactive Pulumi commands now that we do not have a local directory.

We could add a call to `pulumi refresh --clear-pending-creates` when we have provided the `--force` option. This would clear the pending state from the file and then, if any of it actually *has* been created, it should be easy enough to import it or to delete it in the Azure portal and redeploy.

jemrobinson:

n/a

Remove pending Pulumi actions during cleanup (e.g. when run with `--force`). This may leave resources that have been deployed but aren't in the Pulumi state but (a) 'pending' resources need manual intervention any way and (b) manually importing resources requires more knowledge of Pulumi internals than we expect from our users.

Closes #1976

### :microscope: Tests

Tested on SRE deploy and teardown.

### DIFF
--- a/data_safe_haven/infrastructure/project_manager.py
+++ b/data_safe_haven/infrastructure/project_manager.py
@@ -167,6 +167,13 @@ class ProjectManager:
                 f"Cancelling ongoing Pulumi operation for stack [green]{self.stack.name}[/]."
             )
             self.stack.cancel()
+            self.logger.warning(
+                f"Removing any ambiguous Pulumi resources from stack [green]{self.stack.name}[/]."
+            )
+            self.run_pulumi_command("refresh --clear-pending-creates --yes")
+            self.logger.warning(
+                "If you see '[bold]cannot create already existing resource[/]' errors, please manually delete these resources from Azure."
+            )
         except automation.CommandError:
             self.logger.error(
                 f"No ongoing Pulumi operation found for stack [green]{self.stack.name}[/]."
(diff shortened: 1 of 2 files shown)


data_safe_haven/infrastructure/project_manager.py
```diff
@@ -170,19 +167,18 @@ def cancel(self) -> None:
                 f"Cancelling ongoing Pulumi operation for stack [green]{self.stack.name}[/]."
             )
             self.stack.cancel()
+            self.logger.warning(
+                f"Removing any ambiguous Pulumi resources from stack [green]{self.stack.name}[/]."
+            )
+            self.run_pulumi_command("refresh --clear-pending-creates --yes")
```
JimMadge, line 173: We can't use `self.stack.refresh()`?

JimMadge (commented):
Looks good.

I'm not sure what ideal looks like where we depend so much on external processes. This feels helpful though as it should clean up _more_ resources.

data_safe_haven/infrastructure/project_manager.py
jemrobinson: 1. I don't think we want the default behaviour of refresh to include `--clear-pending-creates` 
2. There's no option to `automation.Stack.refresh` that adds this option so we'd have to write our own `refresh` function.

JimMadge (approved):

PR #2023, opened 2024-07-17
Add location specifiers to network resources

### LINKED ISSUE #2022: Networking creation failing
## Problem

SRE deployments from are failing with the following error:

```
 +  azure-native:network:VirtualNetwork sre_networking_virtual_network **creating failed** error: Code="InvalidResourceReference" Message="Resource
/subscriptions/3f1a8e26-eae2-4539-952a-0a6184ec248a/resourceGroups/SHM-GREEN-SRE-FERN-RG-NETWORKING/providers/Microsoft.Network/routeTables/SHM-GREEN-SRE-FERN-ROUTE-TABLE referenced by resource
/subscriptions/3f1a8e26-eae2-4539-952a-0a6184ec248a/resourceGroups/shm-green-sre-fern-rg-networking/providers/Microsoft.Network/virtualNetworks/shm-green-sre-fern-vnet was not found. Please make sure that the referenced resource exists, and that both resources are in
the same region." Details=[{"code":"NotFound","message":"Resource /subscriptions/3f1a8e26-eae2-4539-952a-0a6184ec248a/resourceGroups/SHM-GREEN-SRE-FERN-RG-NETWORKING/providers/Microsoft.Network/routeTables/SHM-GREEN-SRE-FERN-ROUTE-TABLE not found."}]
```

_Originally posted by @jemrobinson in https://github.com/alan-turing-institute/data-safe-haven/issues/1892#issuecomment-2232906789_
            
## Summary

The `pulumi-azure-native` has a default location set at initialisation time, which is stored in the state file.

```json
"location": "uksouth",
"subscriptionId": <redacted>,
"tenantId": <redacted>,
"version": "2.50.1"
```                        

These are set during `install_plugins` using stack options. These *should* be in-sync with what's in the config file. The problem that this PR fixes only happens if you change the "location" parameter in the config. The data in the Pulumi state file doesn't get updated (it seems to be "baked in").

I encountered the problem because I tried to deploy into "ukwest" - found that some resources aren't available there, so ran a teardown and redeploy into "uksouth". This left "ukwest" defined in the state file, which was used whenever "location" wasn't explicitly set on a resource.

jemrobinson:

n/a

Add location specifiers to virtual networks and NSGs as these will otherwise use the default location (which may not be correct).

Closes #2022 

### :microscope: Tests

Tested on previously-broken SRE deployment

### DIFF
--- a/data_safe_haven/infrastructure/components/composite/microsoft_sql_database.py
+++ b/data_safe_haven/infrastructure/components/composite/microsoft_sql_database.py
@@ -82,6 +82,7 @@ class MicrosoftSQLDatabaseComponent(ComponentResource):
         # Deploy a private endpoint for the PostgreSQL server
         private_endpoint = network.PrivateEndpoint(
             f"{self._name}_private_endpoint",
+            location=props.location,
             private_endpoint_name=Output.concat(
                 props.database_server_name, "-endpoint"
             ),
--- a/data_safe_haven/infrastructure/components/composite/postgresql_database.py
+++ b/data_safe_haven/infrastructure/components/composite/postgresql_database.py
@@ -95,6 +95,7 @@ class PostgresqlDatabaseComponent(ComponentResource):
         # Deploy a private endpoint for the PostgreSQL server
         private_endpoint = network.PrivateEndpoint(
             f"{self._name}_private_endpoint",
+            location=props.location,
             private_endpoint_name=Output.concat(
                 props.database_server_name, "-endpoint"
             ),
--- a/data_safe_haven/infrastructure/components/composite/virtual_machine.py
+++ b/data_safe_haven/infrastructure/components/composite/virtual_machine.py
@@ -190,6 +190,7 @@ class VMComponent(ComponentResource):
                     **network_interface_ip_params,
                 )
             ],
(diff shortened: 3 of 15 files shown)


JimMadge (approved):
LGTM

It seems to be good practice to always specify the location.

Do we not have a location specified in the stack configuration? `dev.pulumi.yaml` equivalent.
Or, does that not guarantee resources will use that location?

It might be there is future work here to make this more robust.

jemrobinson:
The `pulumi-azure-native` has a default location set at initialisation time, which is stored in the state file.

```json
"location": "uksouth",
"subscriptionId": <redacted>,
"tenantId": <redacted>,
"version": "2.50.1"
```                        

These are set during `install_plugins` using stack options. These *should* be in-sync with what's in the config file. The problem that this PR fixes only happens if you change the "location" parameter in the config. The data in the Pulumi state file doesn't get updated (it seems to be "baked in").

I encountered the problem because I tried to deploy into "ukwest" - found that some resources aren't available there, so ran a teardown and redeploy into "uksouth". This left "ukwest" defined in the state file, which was used whenever "location" wasn't explicitly set on a resource.

PR #2030, opened 2024-07-22
Exclude doi.org from lychee link check

JimMadge:


jemrobinson (approved):
LGTM. Does this apply to 4.X.Y too?

PR #2031, opened 2024-07-22
User register docs

JimMadge:


jemrobinson (approved):
LGTM

N.B. Are we sure that we want to enforce `<Given Name>.<Surname>` in future or should we use another form or make this something that can be specified during setup? It's something that's easy to make a mistake with esp. with users who come from cultures which use `<Surname> <Given Name>` to specify names.

JimMadge:
@jemrobinson We could do, I didn't know that was something we could control (assumed Entra forced that format for a canonical username).

PR #2032, opened 2024-07-22
Allow unencrypted storage for Azure credential tokens

craddm:
Allows token credential caching on machines without `libsecrets` library (e.g. docker containers)

Allows Azure credentials to be cached on display-less systems (like our devcontainer)


jemrobinson (approved):
LGTM (although our linter might complain).

PR #2052, opened 2024-07-25
Fix construction of apt package lists

### LINKED ISSUE #2046: Fix distribution release selection
The desired state Ansible attempts to select apt packages based on the Ansible fact `distribution_release` (_e.g._ `jammy`, `focal`).

This isn't working as desired at the moment,

https://github.com/alan-turing-institute/data-safe-haven/blob/2fd913486bd31edcb261a12e090e49a655936fd1/data_safe_haven/resources/workspace/ansible/desired_state.yaml#L14

Here the problem is likely that we are trying to expand `ansible_facts['distribution_release']` inside a Jinja2 expression. I think this is not expanding so we are looking up that literal string in the `apt_packages` dict.

JimMadge:

Corrects how apt package lists are constructed so that it include both the common and release specific packages.

Closes #2046

### :microscope: Tests

Tested on a fresh SRE.

### DIFF
--- a/data_safe_haven/resources/workspace/ansible/desired_state.yaml
+++ b/data_safe_haven/resources/workspace/ansible/desired_state.yaml
@@ -5,13 +5,20 @@
 
   tasks:
     - name: Update package cache
+      tags: apt
       ansible.builtin.apt:
         update_cache: true
         cache_valid_time: 600
 
+    - name: List apt packages to install
+      tags: apt
+      ansible.builtin.debug:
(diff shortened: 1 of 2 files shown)


jemrobinson (approved):
LGTM. Was the problem that blank entries weren't equivalent to empty lists?

JimMadge:
> LGTM. Was the problem that blank entries weren't equivalent to empty lists?

Yes 🤦.

(Or, at least `flatten([[1,2,3], Null]) == Null` where as `flatten([[1,2,3], []) == [1,2,3])`)

PR #2064, opened 2024-07-31
Improve credential logging and choice

jemrobinson:

n/a

Currently credential confirmation is written to the logs in unnecessary places (e.g. in the Pulumi output) and the user is not given a chance to check that the credentials are correct.

This adds:
- a confirmation prompt the first time a set of credentials are used
- suppresses credential information
  - after approval has been given
  - inside Pulumi (which is run as a separate process, so the previous class-variable solution was not working)

n/a

### :microscope: Tests

Tested on a fresh SRE deployment and on update

### DIFF
--- a/data_safe_haven/external/api/credentials.py
+++ b/data_safe_haven/external/api/credentials.py
@@ -120,9 +120,9 @@ class AzureSdkCredential(DeferredCredential):
 
     def __init__(
         self,
-        scope: AzureSdkCredentialScope,
+        scope: AzureSdkCredentialScope = AzureSdkCredentialScope.DEFAULT,
         *,
-        skip_confirmation: bool,
+        skip_confirmation: bool = False,
     ) -> None:
         super().__init__(scopes=[scope.value], skip_confirmation=skip_confirmation)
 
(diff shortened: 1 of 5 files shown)


JimMadge:
We recently removed the prompt for users to confirm the credentials, was that because it was difficult to keep in restructuring or because we decided it wasn't a good idea?

craddm:
Was maybe because it was super annoying to have to reconfirm every time? Doing it the first time makes sense

jemrobinson:
Because it was prompting during non-interactive sessions (i.e. during Pulumi calls) which caused lock-ups. With this PR we can set the `skip_confirmation` flag in the Pulumi code which should mean that the prompt (and associated messages) only happen in the interactive part of the code.

data_safe_haven/external/api/credentials.py
```diff
@@ -58,6 +63,35 @@ def decode_token(cls, auth_token: str) -> dict[str, Any]:
...
+            return
+        if (user_id, tenant_id) in DeferredCredential.cache_:
+            return
+        DeferredCredential.cache_.add((user_id, tenant_id))
+        self.logger.info(f"You are logged into the [blue]{target_name}[/] as:")
+        self.logger.info(f"\tuser: [green]{user_name}[/] ({user_id})")
+        self.logger.info(f"\ttenant: [green]{tenant_name}[/] ({tenant_id})")
+        if not console.confirm("Are these details correct?", default_to_yes=True):
+            msg = "Selected credentials are incorrect."
+            raise DataSafeHavenValueError(msg)
```
JimMadge, line 93: Let's make sure we test this method, there are quite a few paths it can take.

JimMadge (commented):
Looks good but I think we should make sure to test the confirmation behaviour.

data_safe_haven/external/api/credentials.py
jemrobinson: Done in 8092290ea

JimMadge (approved):

Issue #1998, opened 2024-07-08
Gitea mirror policy


JimMadge:
Set out which tiers the Gitea mirror feature will be enabled for.

Is there an override in the config?

craddm:
There is nothing about Gitea (or Hedgedoc for that matter) in the config as it stands. 

~~What are we intending to do? We have an existing Gitea resource that is always deployed and that users can use to version control within an SRE.~~

~~Is it possible to configure that to allow the same Gitea resource to do read-only mirroring of external repos? Or do we need a separate Gitea instance to do that? If so do we then need to allow the sysadmin to independently switch off the separate Gitea instance, if they are separate, or to switch off the capability of the main Gitea instance to provide read-only mirroring?~~

Edit: never mind the above, I went back and reread the other relevant issue

When it comes to Tiering, clearly for Tier 1 Gitea should be allowed to mirror anything at all and it doesn't even need to be read-only. 

If we mimic the policy for R/Python packages, then for Tier 2, read-only mirroring of any repo is allowed, while for Tier 3, it would be mirroring of approved repos only (perhaps then request/review would be required).

Issue #1995, opened 2024-07-08
Gitea mirroring security requirements 


JimMadge:
We are aiming to use the internal Gitea instance to mirror external git repositories in a read only manner.

We need to investigate how to do this in a secure way that,

- Does not allow data egreess (_e.g._ allowing pushing to external git repositories)
- Enables admins to review/approve/revoke mirrors (_e.g._ we are probably not comfortable with users being able to add access to arbitrary repositories themselves)

We should design an architecture/configuration that supports these requirements.

JimMadge:
Gitea can set up [pull repositories](https://docs.gitea.com/next/usage/repo-mirror#pulling-from-a-remote-repository) where you are not able to push back to the "original" repository.

The repository is periodically synced and can be synced ad-hoc

- How often is periodic sync? Can we configure that?
- _Who_ can start an ad-hoc sync?

JimMadge:
With Gitea is is possible to enable browsing repositories without signing in (this is enabled by default).

This means we could have a Gitea instance which users are not expected to authenticate with, to browse the read only mirrors.
(I assume that would mean an additional instance, which might be a little confusing, but we could use sensible domain names/theming to differentiate them)

An authenticated user can create mirrors

<img width="851" alt="Screenshot 2024-07-08 at 15 22 46" src="https://github.com/alan-turing-institute/data-safe-haven/assets/23616154/416a98e9-1e02-43fb-83c0-294bd90126c8">

When not authenticated, you can only browse

<img width="851" alt="Screenshot 2024-07-08 at 15 21 31" src="https://github.com/alan-turing-institute/data-safe-haven/assets/23616154/f2f63f79-2435-4c5e-8269-65ac48b688e5">

Configuration,

- Disable self-registration: True
- Show register button: False
- Enable OpenID self-regirstration/authentication: False
- Require sign-in to view pages: False

JimMadge:
@craddm @jemrobinson does a two Gitea instance solution (one for internal development, one for mirrors) sounds good to you?

craddm:
Yes, and giving them sensible domain names sounds good (could be as simple as `external.*` and `internal.*`)

jemrobinson:
Can you stop a non-authenticated user from registering? I wonder whether a simpler solution is to do what we used to do with package mirrors:
- external server: able to access the internet, not accessible by users
- internal server: not able to access the internet, accessible by users

[...]

PR #2071, opened 2024-08-01
WIP Adding a toggleable additional Gitea external mirror

craddm:

Adds an optionally deployable second Gitea instance intended for use as an external mirror that can then be accessed via the internal Gitea instance.

Relates to milestone for release 5.1.0 an all Gitea issues e.g. #1998 

### :microscope: Tests

Ongoing


data_safe_haven/config/sre_config.py
```diff
@@ -64,6 +64,7 @@ def template(cls: type[Self]) -> SREConfig:
                 data_provider_ip_addresses=[
                     "List of IP addresses belonging to data providers"
                 ],
+                gitea_servers="both/internal: whether to deploy both external and internal Gitea servers, or only an internal server",
```
jemrobinson: I suggest that the config option is called something more generic like `git_servers` in case we want to change which server we use in future.

data_safe_haven/infrastructure/programs/sre/gitea_external_server.py
```diff
@@ -0,0 +1,349 @@
...
+        self.ldap_user_search_base = ldap_user_search_base
+        self.location = location
+        self.resource_group_name = resource_group_name
+        self.sre_fqdn = sre_fqdn
+        self.storage_account_key = storage_account_key
+        self.storage_account_name = storage_account_name
+
+
+class SREGiteaExternalServerComponent(ComponentResource):
+    """Deploy Gitea server with Pulumi"""
```
jemrobinson: There's a lot of overlap between this and the `SREGiteaServerComponent`. Can we combine both into a single reusable component?

craddm: Yes, that's what I'm actually doing, I just haven't pushed yet

data_safe_haven/infrastructure/programs/sre/user_services.py
```diff
@@ -93,30 +95,36 @@ def __init__(
...
-                storage_account_name=props.storage_account_name,
-            ),
-            opts=child_opts,
-            tags=child_tags,
-        )
+        # Deploy the Gitea servers
+        if props.gitea_servers == GiteaServers.BOTH:
+            gitea_servers = ["external", "internal"]
+        else:
+            gitea_servers = ["internal"]
```
jemrobinson:
```suggestion
        gitea_servers = ["external", "internal"] if (props.gitea_servers == GiteaServers.BOTH) else ["internal"]
```
not sure this is clearer though

data_safe_haven/infrastructure/programs/sre/user_services.py
```diff
@@ -93,30 +95,36 @@ def __init__(
...
+                    database_subnet_id=props.subnet_containers_support_id,
+                    database_password=props.gitea_database_password,
+                    dns_server_ip=props.dns_server_ip,
+                    dockerhub_credentials=props.dockerhub_credentials,
+                    gitea_server=gitea_server,
+                    ldap_server_hostname=props.ldap_server_hostname,
+                    ldap_server_port=props.ldap_server_port,
+                    ldap_username_attribute=props.ldap_username_attribute,
+                    ldap_user_filter=props.ldap_user_filter,
+                    ldap_user_search_base=props.ldap_user_search_base,
```
jemrobinson, line 129: We ~probably don't want~ don't need the external server to plug in to our LDAP directory.

data_safe_haven/infrastructure/programs/sre/user_services.py
craddm: I'm also switching to using a boolean True/False for deploying the external mirror anyway, as we'd (presumably) always want the internal one anyway

jemrobinson: Sounds like I should stop commenting on your WIP PR 😆

craddm: Already some good points to nip some things in the bud 😆

jemrobinson:
One other thing to consider - you'll probably want the external mirror deployed into its own subnet so that we can set the networking rules accordingly (external one will be allowed internet access but no access from workspaces; internal one will have no internet access but access allowed from workspaces).

data_safe_haven/infrastructure/programs/sre/user_services.py
JimMadge: Some notes on Gitea configuration [here](https://github.com/alan-turing-institute/data-safe-haven/issues/1995)

JimMadge:
Closing this for now as it not on our plan.

PR #2076, opened 2024-08-02
Fix IP address in list

jemrobinson:

n/a

Fix logic about determining whether current IP address is allowed.

n/a

### :microscope: Tests

Tested from an IP that should have been working but wasn't. Added an additional test to catch this.

### DIFF
--- a/data_safe_haven/commands/context.py
+++ b/data_safe_haven/commands/context.py
@@ -6,9 +6,7 @@ import typer
 
 from data_safe_haven import console, validators
 from data_safe_haven.config import ContextManager
-from data_safe_haven.exceptions import (
-    DataSafeHavenConfigError,
-)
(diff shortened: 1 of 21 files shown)


JimMadge (commented):
As the IP check is duplicated could we put that behaviour in a single function.

jemrobinson:
> As the IP check is duplicated could we put that behaviour in a single function.

It *is* in a single function. That's what `ip_address_in_list()` is. The only other bit is the logic that does:

```
if not ip_address_in_list(my_list):
   raise Exception
```

which is slightly different in the two cases.

JimMadge:
Ah, I didn't spot the difference. I think that is fine. Could do a `func(verb="deploy")` and `func(verb="teardown")` if we want to be very DRY.

JimMadge (approved):

PR #2093, opened 2024-08-06
Add RStudio

### LINKED ISSUE #2041: Add RStudio
In previous versions we included RStudio, a popular R IDE.

Supporting RStudio is a little tricky as there is no deb in the canonical repositories, no PPA, and no snap.
To support this we would need to find way to fetch a deb from the Posit site.

We have coverage for R IDEs/editors (codium, vim, emacs).
However, many R developers have a strong preference RStudio so supporting it would likely help researchers.

jemrobinson:

n/a

Install a hard-coded version of RStudio.

Closes #2041 

### :microscope: Tests

Tested on a new SRE deployment

### DIFF
--- a/data_safe_haven/infrastructure/programs/sre/firewall.py
+++ b/data_safe_haven/infrastructure/programs/sre/firewall.py
@@ -268,6 +268,18 @@ class SREFirewallComponent(ComponentResource):
                             source_addresses=props.subnet_workspaces_prefixes,
                             target_fqdns=PermittedDomains.UBUNTU_SNAPCRAFT,
                         ),
+                        network.AzureFirewallApplicationRuleArgs(
+                            description="Allow external RStudio deb downloads",
+                            name="AllowRStudioDeb",
+                            protocols=[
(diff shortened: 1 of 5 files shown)


jemrobinson:
@JimMadge: Not the most elegant solution, but it works for now. Ideas for a better way to do this in #2094.

JimMadge (approved):
LGTM.

It would be better to do something like 

```yaml
    - name: Install deb packages
      tags: apt
      block:
        - ansible.builtin.stat:
            path: "{{ item.creates }}"
          register: app

        - ansible.builtin.get_url:
            url: "{{ item.source }}"
            dest: /tmp/debfile.deb
            checksum: "{{ item.checksum }}"
          when: not app.exists

        - ansible.builtin.apt:
            deb: /tmp/debfile.deb
            state: present
          when: not app.exists
      loop: "{{ deb_packages[ansible_facts.distribution_release] }}"
```

(Except you can't use loops on blocks. We could use `include_tasks` though.)

Let's make that a new issue.

PR #2098, opened 2024-08-06
Set xfce4 defaults

jemrobinson:

n/a

- Use `xfce4-terminal` instead of the ugly `debian-x-terminal-emulator`
- Set the default colour scheme

n/a

### :microscope: Tests

Tested with a new VM in an existing SRE with all `/home` directories deleted.

Screenshot 2024-08-06 at 19 24 35


jemrobinson:
@JimMadge @craddm : Do we want to disable the screenlock (as was done previously) or leave it enabled?

jemrobinson:
Merging after personal testing in order to be ready for pentest.

JimMadge:
> Use xfce4-terminal instead of the ugly debian-x-terminal-emulator

I noticed that too :smile: meant to open an issue.

PR #2111, opened 2024-08-08
Improve VM SKU documentation and errors

jemrobinson:
## :white_check_mark: Checklist

- [x] You have given your pull request a meaningful title (_e.g._ `Enable foobar integration` rather than `515 foobar`).
- [x] You are targeting the appropriate branch. If you're not certain which one this is, it should be **`develop`**.
- [x] Your branch is up-to-date with the **target branch** (it probably was when you started, but it may have changed since then).

### :vertical_traffic_light: Depends on

n/a

### :arrow_heading_up: Summary

**IMPORTANT: I am confident that merging this will not affect the findings of the ongoing penetration test** 

- Add note to documentation explaining that premium-SSD-compatible VMs must be used
- Improve logging of Pulumi errors so that this also appears in the logs:

```console
 stderr:
Pulumi error: +  azure-native:compute:VirtualMachine sre_workspaces_vm_workspace_02 creating (3s)  Code="InvalidParameter" Message="Requested operation cannot be performed because the VM size Standard_D8_v5 does not support the storage account
type Premium_LRS of disk 'shm-pink-sre-coral-vm-workspace-02-osdisk'. Consider updating the VM to a size that supports Premium storage." Target="osDisk.managedDisk.storageAccountType"
Pulumi error: +  azure-native:compute:VirtualMachine sre_workspaces_vm_workspace_02 **creating failed**  Code="InvalidParameter" Message="Requested operation cannot be performed because the VM size Standard_D8_v5 does not support the storage
account type Premium_LRS of disk 'shm-pink-sre-coral-vm-workspace-02-osdisk'. Consider updating the VM to a size that supports Premium storage." Target="osDisk.managedDisk.storageAccountType"
Pulumi error:    pulumi:pulumi:Stack data-safe-haven-shm-pink-sre-coral running  update failed
Pulumi error:  Code="InvalidParameter" Message="Requested operation cannot be performed because the VM size Standard_D8_v5 does not support the storage account type Premium_LRS of disk 'shm-pink-sre-coral-vm-workspace-02-osdisk'. Consider
updating the VM to a size that supports Premium storage." Target="osDisk.managedDisk.storageAccountType"
Pulumi error:  update failed
Pulumi update failed.
Pulumi deployment failed.
Could not deploy Secure Research Environment 'coral'.
```

### :closed_umbrella: Related issues

Closes #2067

### :microscope: Tests

Tested by trying to add an incompatible VM to an existing SRE

Issue #1952, opened 2024-06-20
Fix deployment logging output


jemrobinson:
## :white_check_mark: Checklist

- [x] I have searched open and closed issues for duplicates.
- [x] This is a request for a new feature in the Data Safe Haven or an upgrade to an existing feature.
- [x] The feature is still missing in the [latest version](https://github.com/alan-turing-institute/data-safe-haven/releases).
- [x] I have read through the [documentation](https://alan-turing-institute.github.io/data-safe-haven/).
- [x] This isn't an open-ended question (open a [discussion](https://github.com/alan-turing-institute/data-safe-haven/discussions) if it is).

## :strawberry: Suggested change

Since the recent logging changes in #1936, the output from Pulumi is (a) extremely verbose and (b) full of unescaped ANSI codes. This should be fixed.

<details
<summary>Abridged log output</summary>

```
Refreshing stack shm-green.
Previewing changes for stack shm-green.
[38;5;13m[1mPreviewing update (shm-green):[0m
[0mdebug: [0m[0mregistering resource: ty=pulumi:pulumi:Stack, name=data-safe-haven-shm-green, custom=False, remote=False[0m
[0mdebug: [0m[0mregistering resource: ty=dsh:shm:NetworkingComponent, name=shm_networking, custom=False, remote=False[0m
[0mdebug: [0m[0mregistering resource: ty=azure-native:resources:ResourceGroup, name=shm_networking_resource_group, custom=True, remote=False[0m
[0mdebug: [0m[0mregistering resource: ty=azure-native:network:Zone, name=shm_networking_dns_zone, custom=True, remote=False[0m
[0mdebug: [0m[0mregistering resource: ty=azure-native:network:RecordSet, name=shm_networking_caa_record, custom=True, remote=False[0m
[0mdebug: [0m[0mregistering resource: ty=azure-native:network:RecordSet, name=shm_networking_domain_verification_record, custom=True, remote=False[0m
[0mdebug: [0m[0mWaiting for outstanding RPCs to complete[0m
[0mdebug: [0m[0mRPCs successfully completed[0m
[0mdebug: [0m[0mbeginning rpc register resource[0m
[0mdebug: [0m[0mbeginning rpc register resource[0m
[0mdebug: [0m[0mbeginning rpc register resource[0m
[0mdebug: [0m[0mbeginning rpc register resource[0m
[0mdebug: [0m[0mbeginning rpc register resource[0m
[...]

PR #2113, opened 2024-08-08
Fix pulumi colours

### LINKED ISSUE #2112: from_ansi strips colours

- **Operating System:** macOS
- **Data Safe Haven version:** develop @ bd6b270

## :package: Packages

<details>

```none
Paste list of packages here
```

</details>

The `from_ansi` function actually strips ANSI information from text. It converts `str -> rich.Text` and then passes this to `logger.info`. However, at some point in the `logger` handling chain, the `rich.Text` object is converted to a string, stripping the markup information before it reaches the `RichHandler`.

Either passing the ANSI-formatted string directly, or converting this to Rich markup would avoid this problem.

Screenshot 2024-08-08 at 16 06 05

<details>

```none
Your log details here
```

</details>

## :recycle: To reproduce

Deploy an SRE and observe the lack of colour in the Pulumi output.

jemrobinson:

- #2111

**IMPORTANT: I am confident that merging this will not affect the findings of the ongoing penetration test** 

Fix Pulumi colours

#### Before
Screenshot 2024-08-08 at 16 06 05

#### After
Screenshot 2024-08-08 at 16 17 30

Closes #2112

### :microscope: Tests

Tested by rerunning against an existing SRE.

### DIFF
--- a/data_safe_haven/functions/strings.py
+++ b/data_safe_haven/functions/strings.py
@@ -8,7 +8,6 @@ import uuid
 from collections.abc import Sequence
 
 import pytz
-from rich.text import Text
 
 from data_safe_haven.exceptions import DataSafeHavenValueError
 
@@ -113,11 +112,6 @@ def seeded_uuid(seed: str) -> uuid.UUID:
 def sha256hash(input_string: str) -> str:
     """Return the SHA256 hash of a string as a string."""
     return hashlib.sha256(input_string.encode("utf-8")).hexdigest()
-
-
-def strip_ansi(text: str) -> str:
-    """Strip all ANSI formatting from a string"""
-    return str(Text.from_ansi(text))
 
 
 def truncate_tokens(tokens: Sequence[str], max_length: int) -> list[str]:
--- a/data_safe_haven/functions/strings.py
+++ b/data_safe_haven/functions/strings.py
@@ -8,6 +8,7 @@ import uuid
 from collections.abc import Sequence
 
 import pytz
+from rich.text import Text
 
 from data_safe_haven.exceptions import DataSafeHavenValueError
 
@@ -112,6 +113,11 @@ def seeded_uuid(seed: str) -> uuid.UUID:
 def sha256hash(input_string: str) -> str:
     """Return the SHA256 hash of a string as a string."""
     return hashlib.sha256(input_string.encode("utf-8")).hexdigest()
+
+
+def strip_ansi(text: str) -> str:
+    """Strip all ANSI formatting from a string"""
+    return str(Text.from_ansi(text))
 
 
 def truncate_tokens(tokens: Sequence[str], max_length: int) -> list[str]:
--- a/data_safe_haven/functions/__init__.py
+++ b/data_safe_haven/functions/__init__.py
@@ -9,7 +9,6 @@ from .strings import (
     replace_separators,
     seeded_uuid,
     sha256hash,
-    strip_ansi,
     truncate_tokens,
 )
 
@@ -25,6 +24,5 @@ __all__ = [
     "replace_separators",
     "seeded_uuid",
     "sha256hash",
-    "strip_ansi",
     "truncate_tokens",
 ]
--- a/data_safe_haven/functions/__init__.py
+++ b/data_safe_haven/functions/__init__.py
@@ -9,6 +9,7 @@ from .strings import (
     replace_separators,
     seeded_uuid,
     sha256hash,
+    strip_ansi,
     truncate_tokens,
 )
 
@@ -24,5 +25,6 @@ __all__ = [
     "replace_separators",
     "seeded_uuid",
     "sha256hash",
+    "strip_ansi",
     "truncate_tokens",
 ]
--- a/data_safe_haven/infrastructure/project_manager.py
+++ b/data_safe_haven/infrastructure/project_manager.py
@@ -21,7 +21,7 @@ from data_safe_haven.exceptions import (
     DataSafeHavenPulumiError,
 )
 from data_safe_haven.external import AzureSdk, PulumiAccount
-from data_safe_haven.functions import get_key_vault_name, replace_separators
+from data_safe_haven.functions import get_key_vault_name, replace_separators, strip_ansi
 from data_safe_haven.logging import from_ansi, get_console_handler, get_logger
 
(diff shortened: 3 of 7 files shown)


JimMadge (changes requested):
I'm not sure I follow how this affects the Pulumi output.

Didn't we add `logging.from_ansi` because there was a problem with sending the output straight to the logger?

I think there might be an easier way to do this with a [rich style](https://rich.readthedocs.io/en/stable/reference/text.html?highlight=from_ansi#rich.text.Text.from_ansi).
The default style argument for `from_ansi` is `""` so that might explain why the rich `Text` objects have no colours.

jemrobinson:
The current process looks like the following:

1. The raw output from Pulumi logging has ANSI colour codes embedded in it (_e.g._ `"\x1b[38;5;13m\x1b[1mDuration\x1b[0m 45s"`)
2. We then pass this string to `ProjectManager::log_message()`
3. This passes the string on to `logging::from_ansi()`
4. Inside this function we convert the coloured string to a `rich.Text` object (_e.g._ `<text 'Duration 45s' [Span(0, 8, Style(color=Color('color(13)', ColorType.STANDARD, number=13), bold=True))]>`)
5. This is then passed to `logging.logger.info`
6. At some point before this object is passed to the `rich.Handler` it's converted from a `rich.Text` object to a string, dropping the formatting along the way
7. The `rich.Handler` then prints the non-coloured string to the console

The change in this PR does the following:

1. The raw output from Pulumi logging has ANSI colour codes embedded in it (_e.g._ `"\x1b[38;5;13m\x1b[1mDuration\x1b[0m 45s"`)
2. We pass this directly to `logging.logger.info`
3. No string conversion needs to take place since this is already just a string
4. The `rich.Handler` then prints the coloured string to the console


#### TL;DR

There's no problem with how `rich` is handling the string conversion - it's just that `logging` doesn't want to pass `rich.Text` objects around internally.

The other option I considered was to convert the ANSI markers to `rich` markup that would pass through the `logging` internals equally well, but there doesn't seem to be much point in converting from ANSI just to convert straight back again.

jemrobinson:
Note on the `strip_ansi()` function:

It's because you quite often get Pulumi output that looks like this:

`"\033[31;merror:\033[0m Code='AnErrorCode'"`

Screenshot 2024-08-09 at 10 48 39

which we'd like to log as `"Pulumi output: Code='AnErrorCode'"`.

If we try to do `error_line.replace("error:", "").strip()` then the first whitespace character is masked by the ANSI colour codes so you get:

`"Pulumi output: \033[31;m\033[0m Code='AnErrorCode'"`.

which looks like

Screenshot 2024-08-09 at 10 48 56

and effectively a double-space in the output.

If we can strip the ANSI codes first, then the `strip()` call will work. There are other ways to strip ANSI codes but they mainly involve complicated regexes. If you'd prefer to do this then we can.

I'll move this commit to the other PR though.

data_safe_haven/functions/strings.py
```diff
@@ -114,6 +115,11 @@ def sha256hash(input_string: str) -> str:
     return hashlib.sha256(input_string.encode("utf-8")).hexdigest()
 
 
+def strip_ansi(text: str) -> str:
+    """Strip all ANSI formatting from a string"""
+    return str(Text.from_ansi(text))
```
jemrobinson: From [this StackOverflow post](https://stackoverflow.com/questions/14693701/how-can-i-remove-the-ansi-escape-sequences-from-a-string-in-python) we could also use:

```suggestion
    ansi_escape = re.compile(r'\x1B(?:[@-Z\\-_]|\[[0-?]*[ -/]*[@-~])')
    return ansi_escape.sub("", text)
```

JimMadge:
@jemrobinson Could you check if https://github.com/alan-turing-institute/data-safe-haven/tree/pulumi_colour does the same?
It seems to work for me.

I can't quite remember the thinking behind removing the colour, I think it was more about removing the ANSI codes. The logging handler doesn't seem to work with rich Text (which is odd given it is a rich handler) but you can convert the Text object to a string with rich markup like we use for other log messages.

jemrobinson:
Yes it works, I've already tested this. I don't really see the benefit of converting ANSI-text to markup if it's not causing problems though: see my note above: `The other option I considered was to convert the ANSI markers to rich markup that would pass through the logging internals equally well, but there doesn't seem to be much point in converting from ANSI just to convert straight back again.`

I don't think we need the `log_message` or `from_ansi` functions at all.

NB. The logging handler **does** work with `rich.Text` objects, but the Python `logging` library converts these to `str` before they make it as far as `RichHandler` (I tested this by adding debugging print statements throughout the logging and Rich libraries).

JimMadge:
Before the ANSI codes were [printed](https://github.com/alan-turing-institute/data-safe-haven/issues/1952).

If that's not true now, I'm not sure what has changed but we wouldn't need the extra handling.
Otherwise, I think converting from ANSI markup to rich markup is most sensible because that is what we use elsewhere.

jemrobinson:
I don't see printed ANSI codes with this change, and I'm not sure why this was happening in #1952. Note that in `5.0.0-rc1` we also passed Pulumi output straight to the logger (see below) and this was printed correctly.

https://github.com/alan-turing-institute/data-safe-haven/blob/afb29b42510af9fa261b891348fb08d8baddca2b/data_safe_haven/infrastructure/stack_manager.py#L310-L318

If you think there are other benefits to converting to rich markup, I'm happy to restore a helper function that does so.

JimMadge:
Do we need the changes here then? The `strip_ansi` stuff feels like a lot of complexity to remove a space.

I feel like that is more likely to cause problems than just pre-pending the full Pulumi error with "Pulumi error:", replacing anything.

jemrobinson:
> Do we need the changes here then? The `strip_ansi` stuff feels like a lot of complexity to remove a space.

1. We **do** need the changes that replace the `from_ansi` function (or something equivalent)
2. We **don't** need the `strip_ansi` changes - the change you suggest would also remove the need for us to try to guess the structure of a Pulumi error message. It might look a bit like this:

Screenshot 2024-08-09 at 12 38 01

jemrobinson:
`strip_ansi()` removed in 51a0ff935. Are you happy to merge @JimMadge ?

JimMadge (approved):

PR #2103, opened 2024-08-06
Add application icons

jemrobinson:
## :white_check_mark: Checklist

- [x] You have given your pull request a meaningful title (_e.g._ `Enable foobar integration` rather than `515 foobar`).
- [x] You are targeting the appropriate branch. If you're not certain which one this is, it should be **`develop`**.
- [x] Your branch is up-to-date with the **target branch** (it probably was when you started, but it may have changed since then).

### :vertical_traffic_light: Depends on

n/a

### :arrow_heading_up: Summary

**IMPORTANT: I am confident that merging this will not affect the findings of the ongoing penetration test** 

- Add application icons for
  - Gitea
  - Hedgedoc
  - /data folder
  - /output folder
  - /shared folder

### :closed_umbrella: Related issues

n/a

### :microscope: Tests

Tested on a fresh SRE deployment

Screenshot 2024-08-06 at 21 17 07

PR #2092, opened 2024-08-06
Update mount points

JimMadge:
## :white_check_mark: Checklist

- [x] You have given your pull request a meaningful title (_e.g._ `Enable foobar integration` rather than `515 foobar`).
- [x] You are targeting the appropriate branch. If you're not certain which one this is, it should be **`develop`**.
- [x] Your branch is up-to-date with the **target branch** (it probably was when you started, but it may have changed since then).

### :vertical_traffic_light: Depends on

- #2103 

### :closed_umbrella: Related issues

Closes #2027 

### :microscope: Tests

- Tested in a clean deployment.

PR #2115, opened 2024-08-09
Add ansible vars file

### LINKED ISSUE #2075: Cloud-init variables

We currently expand our cloud-init file with a local of instance-specific variables. It would be better to pass these to Ansible for expansion in templates.

Not sure on the best way to pass variables to the Ansible playbook. @JimMadge might have thoughts.

JimMadge:

#2092 

Adds an ansible vars file, which is constructed in the infrastructure code and uploaded to the desired state share by Pulumi.
Includes changes from https://github.com/alan-turing-institute/data-safe-haven/pull/2103 for release v5.0.0

Closes #2075 

### :microscope: Tests

Tested on a fresh deployment.
All vars and templates as expected and system is working.


jemrobinson:
It might be easier to wait until release 5.0.0 is tagged (next week?) and merged back into develop before looking further into this. What do you think @JimMadge ? We could migrate other variables into this file too and close #2075 in the process.

JimMadge:
I don't think this would have to wait for the release, as those changes will be merged back into develop anyway.

In the first instance, I'd want to check that this works before moving all variables from cloud init across.
If it does, we could move all of them in this PR though :+1:.

jemrobinson:
I think I'd like to see:

- test that this works on a fresh deployment
- if it does, see how many other variables can be moved across

I agree there's no explicit *problem* in merging into develop before we tag v5.0.0 but note that this branch contains some changes that are already in `release-v5.0.0` and aren't in `develop` so it would be cleaner if we could:

- merge `release-v5.0.0` into `latest`
- tag release 5.0.0
- merge `latest` into `develop`
- THEN merge this PR into `develop`

JimMadge:
> I think I'd like to see:
> 
>     * test that this works on a fresh deployment
> 
>     * if it does, see how many other variables can be moved across
> 
> 

:+1:

> I agree there's no explicit _problem_ in merging into develop before we tag v5.0.0 but note that this branch contains some changes that are already in `release-v5.0.0` and aren't in `develop` so it would be cleaner if we could:
> 
>     * merge `release-v5.0.0` into `latest`
>     * tag release 5.0.0
>     * merge `latest` into `develop`
>     * THEN merge this PR into `develop`

I can't see that. They are the same changes end up in `develop` either way and it won't cause any conflicts.

jemrobinson:
Sorry, I meant cleaner in terms of the logic not whether there will be git conflicts (I agree - there won't be). If we merge release-v5.0.0 first then the logic is:

- this change was made in anticipation of release 5.0.0
- it was merged back into develop when the release was made
- then another PR that makes changes on top of that was added to develop

rather than

- this change was made in anticipation of release 5.0.0
- it was also separately made in another PR meged into develop

JimMadge:
Screenshot 2024-08-13 at 09 53 39

The vars file is being templated.

JimMadge:
Templating works

```console
root@shm-daimyo-sre-hojo-vm-workspace-01:~# ls /etc/skel/Desktop/
gitea.desktop  hedgedoc.desktop  input.desktop  output.desktop  shared.desktop
root@shm-daimyo-sre-hojo-vm-workspace-01:~# cat /etc/skel/Desktop/{gitea,hedgedoc}.desktop
[Desktop Entry]
Version=1.0
Type=Link
Name=Gitea
Comment=
Icon=/usr/local/share/icons/gitea.png
URL=http://gitea.hojo.daimyo.develop.turingsafehaven.ac.uk
[Desktop Entry]
Version=1.0
Type=Link
Name=HedgeDoc
Comment=
Icon=/usr/local/share/icons/hedgedoc.png
URL=http://hedgedoc.hojo.daimyo.develop.turingsafehaven.ac.uk
```

data_safe_haven/infrastructure/common/ip_ranges.py
```diff
@@ -14,7 +14,7 @@ class SREIpRanges:
     apt_proxy_server = vnet.next_subnet(8)
     clamav_mirror = vnet.next_subnet(8)
     data_configuration = vnet.next_subnet(8)
-    data_desired_state = vnet.next_subnet(8)
+    desired_state = vnet.next_subnet(8)
     data_private = vnet.next_subnet(8)
```
jemrobinson:
```suggestion
    data_private = vnet.next_subnet(8)
    desired_state = vnet.next_subnet(8)
```

data_safe_haven/infrastructure/programs/sre/desired_state.py
```diff
@@ -0,0 +1,248 @@
...
+        tags: Input[Mapping[str, Input[str]]] | None = None,
+    ) -> None:
+        super().__init__("dsh:sre:DesiredStateComponent", name, {}, opts)
+        child_opts = ResourceOptions.merge(opts, ResourceOptions(parent=self))
+        child_tags = {"component": "data"} | (tags if tags else {})
+
+        # Deploy desired state storage account
+        # - This holds the /desired_state container that is mounted by workspaces
+        # - Azure blobs have worse NFS support but can be accessed with Azure Storage Explorer
+        storage_account = storage.StorageAccount(
```
jemrobinson: Can we add a wrapped component (e.g. WrappedStorageAccount) that would handle all the common storage account features so that we can easily see which options are varying between different accounts? I think we are creating at least 3, and it would be good to ensure that changes are kept in-sync.

jemrobinson:
@JimMadge: worth picking up after RSECon?

data_safe_haven/infrastructure/programs/sre/desired_state.py
JimMadge: Maybe not for _all_ storage accounts.

This is one is a common archetype though, something like "NFSv3EnabledStorageAccount".
Maybe "V2StorageAccount" is another one.

jemrobinson: No, I mean we should create a new class that looks like:

```
class WrappedStorageAccount(StorageAccount):
  def __init__(
    self,
    a_parameter,
    a_second_parameter,
  ):
    super().__init__(
      common_parameter="value",
      common_parameter2="value2",
      a_parameter=a_parameter,
      a_second_parameter=a_second_parameter,
```

so the parameters that genuinely are common for all our storage accounts (e.g. TLS=1.2) are defined in one place.

We currently have a blob account (mounted as NFS), a files account (mounted by containers) and another files account (mounted as NFS) no two of which are identical, but all of which share some common parameters.

JimMadge: Yes, but isn't it sensible to have a hierarchy. NFSv3EnabledStorageAccount and V2StorageAccount inherit from GenericStorageAccount.

Otherwise it is the same argument, we have multiple WrappedStorageAccounts which share common parameters.

jemrobinson: Sure, that's fine, but we'll only have one `NFSv3EnabledStorageAccount` and one `V2StorageAccount` right?

JimMadge: There are two NFSv3 storage accounts. I think that is why I was more keen on abstracting that rather than creating a wrapper for all storage accounts (and as that is the kind of account this PR directly touches).

data_safe_haven/infrastructure/programs/sre/networking.py
```diff
@@ -483,10 +483,10 @@ def __init__(
             opts=child_opts,
             tags=child_tags,
         )
-        nsg_data_desired_state = network.NetworkSecurityGroup(
-            f"{self._name}_nsg_data_desired_state",
+        nsg_desired_state = network.NetworkSecurityGroup(
```
jemrobinson, line 486: Is it worth leaving the `_data` prefix in these names? This NSG/storage account etc. still contains data (desired state data) even if it's not part of the `SREData` component.

data_safe_haven/infrastructure/components/wrapped/nfsv3_storage_account.py
```diff
@@ -0,0 +1,68 @@
+from collections.abc import Mapping, Sequence
+
+from pulumi import Input, Output, ResourceOptions
+from pulumi_azure_native import storage
+
+from data_safe_haven.external import AzureIPv4Range
+
+
+class NFSV3StorageAccount(storage.StorageAccount):
```
jemrobinson: Prefer the name to be of the form `WrappedX` for clarity.

data_safe_haven/infrastructure/components/wrapped/nfsv3_storage_account.py
```diff
@@ -0,0 +1,68 @@
...
+    encryption_args = storage.EncryptionArgs(
```
jemrobinson, line 10: Is there any reason for these to be a class-level attribute rather than living in the `super().__init__` call?

jemrobinson (approved):
Overall LGTM. A couple of comments though.

data_safe_haven/infrastructure/components/wrapped/nfsv3_storage_account.py
JimMadge: What would you call it?

I didn't want to call it `WrappedStorageAccount` because it isn't really that general. It feels to me more like a specific archetype of a storage account that we use.

`WrappedNFSV3StorageAccount`?

data_safe_haven/infrastructure/programs/sre/networking.py
JimMadge: I preferred making the distinction for clarity as the data and desired state parts are not independent `ComponentResource`s.

data_safe_haven/infrastructure/components/wrapped/nfsv3_storage_account.py
JimMadge: It felt appropriate as it will be shared by all instances. If there is a mistake it should be processed before an instance is constructed (I think 🤔).

There can be some side-effects though. Do you think it is a problem?

data_safe_haven/infrastructure/components/wrapped/nfsv3_storage_account.py
jemrobinson: I think `WrappedNFSV3StorageAccount` is OK for now - we can add a more generic `WrappedStorageAccount` as an intermediate layer if/when we write a similar abstraction for the other storage accounts that we're using.

data_safe_haven/infrastructure/components/wrapped/nfsv3_storage_account.py
jemrobinson: I don't think it's a problem. Since these are all static properties anyway, it shouldn't matter if they're shared between instances or not.

Issue #2107, opened 2024-08-07
Smoke tests failing on Tier 3 equivalent SRE


craddm:
## :white_check_mark: Checklist

- [x] I have searched open and closed issues for duplicates.
- [x] This is a problem observed when **deploying** a Data Safe Haven.
- [x] I can reproduce this with the [latest version](https://github.com/alan-turing-institute/data-safe-haven/releases).
- [x] I have read through the [documentation](https://alan-turing-institute.github.io/data-safe-haven/).
- [x] This isn't an open-ended question (open a [discussion](https://github.com/alan-turing-institute/data-safe-haven/discussions) if it is).

## :computer: System information

- **Operating System:** macOs/Linux
- **Data Safe Haven version:** release-v5.0.0rc3

## :package: Packages

<details>
<summary>List of packages</summary>

```none
Paste list of packages here
```

</details>

## :no_entry_sign: Describe the problem

Smoke tests fail on Tier 3 SREs because necessary packages are not on the current allowlists. The [allowlists](https://github.com/alan-turing-institute/data-safe-haven/tree/7b4e6f464e0fec4f7a1461a475d9032b0c9101cc/data_safe_haven/resources/software_repositories/allowlists) need to be brought up-to-date.

## :deciduous_tree: Log messages

<details>
<summary>Relevant log messages</summary>

```none
root@shm-turtles-sre-leonardo-vm-workspace-01:/usr/local/smoke_tests# bats run_all_tests.bats 
 ✓ Mounted drives (/data) 
 ✓ Mounted drives (/home) 
 ✓ Mounted drives (/output) 
 ✓ Mounted drives (/shared) 
 ✗ Python package repository
   (from function `initialise_python_environment' in file run_all_tests.bats, line 11,
    in test file run_all_tests.bats, line 66)
     `initialise_python_environment' failed
     ERROR: HTTP error 403 while getting http://nexus.leonardo.blue.develop.turingsafehaven.ac.uk/repository/pypi-proxy/packages/pip/24.2/pip-24.2-py3-none-any.whl#sha256=2cd581cf58ab7fcfca4ce8efa6dcacd0de5bf8d0a3eb9ec927e07405f4d9e2a2 (from http://nexus.leonardo.blue.develop.turingsafehaven.ac.uk/repository/pypi-proxy/simple/pip/) (requires-python:>=3.8)
[...]

PR #2114, opened 2024-08-08
Fix Python and R repository smoke tests

craddm:
## :white_check_mark: Checklist

- [x] You have given your pull request a meaningful title (_e.g._ `Enable foobar integration` rather than `515 foobar`).
- [x] You are targeting the appropriate branch. If you're not certain which one this is, it should be **`develop`**.
- [x] Your branch is up-to-date with the **target branch** (it probably was when you started, but it may have changed since then).

### :vertical_traffic_light: Depends on

This is best done in combination with an update to the `nexus-allowlist` container to allow installation of older versions of R packages. Current smoke tests will still fail as they require installation of older R package versions, which is not currently allowed by the Nexus server.

### :arrow_heading_up: Summary

Minimally updates the PyPi allowlist, ensuring that the smoke tests can run with all required dependencies now on the list.

Updates the R smoke tests to now refer to packages that are on allowlist.

Fixes the following tests:
- Python package repository
- R package repository
- Python functionality
- MS SQL database (Python & R)
- Postgres database (Python & R)

Does not fix the R functionality test.

### :closed_umbrella: Related issues

Closes #2107 

### :microscope: Tests

Tested on a Tier 3 equivalent SRE deployment - `noteasybeing.green.develop.turingsafehaven.ac.uk`

PR #2116, opened 2024-08-09
Update Nexus Allowlist container images

JimMadge:

Update Nexus Allowlist to a release which allows installing archived CRAN packages.

Related to #2107

### DIFF
--- a/data_safe_haven/infrastructure/programs/sre/software_repositories.py
+++ b/data_safe_haven/infrastructure/programs/sre/software_repositories.py
@@ -161,7 +161,7 @@ class SRESoftwareRepositoriesComponent(ComponentResource):
                 container_group_name=f"{stack_name}-container-group-software-repositories",
                 containers=[
                     containerinstance.ContainerArgs(
-                        image="caddy:2.8.4",
+                        image="caddy:2.8",
                         name="caddy"[:63],
                         ports=[
                             containerinstance.ContainerPortArgs(
@@ -184,7 +184,7 @@ class SRESoftwareRepositoriesComponent(ComponentResource):
                         ],
                     ),
                     containerinstance.ContainerArgs(
-                        image="sonatype/nexus3:3.70.1",
+                        image="sonatype/nexus3:3.71.0",
                         name="nexus"[:63],
                         environment_variables=[],
                         ports=[],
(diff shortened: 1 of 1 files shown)


data_safe_haven/infrastructure/programs/sre/software_repositories.py
```diff
@@ -161,7 +161,7 @@ def __init__(
                 container_group_name=f"{stack_name}-container-group-software-repositories",
                 containers=[
                     containerinstance.ContainerArgs(
-                        image="caddy:2.8.4",
+                        image="caddy:2.8",
```
jemrobinson: Why has this changed?

jemrobinson (changes requested):
LGTM but are we sure we don't want this in the `release-v5.0.0` branch?

data_safe_haven/infrastructure/programs/sre/software_repositories.py
JimMadge: Feels like an over specification to me.

JimMadge:
> LGTM but are we sure we don't want this in the `release-v5.0.0` branch?

_I'm_ not very worried about it but it does change (quite significantly) what you can access through the CRAN proxy.

jemrobinson:
OK, but we need what goes into v5.0.0 to be deployable. Either we need this fix or we need changes to the smoke tests so they actually work. I'm not sure which is better - what do you think?

JimMadge:
> OK, but we need what goes into v5.0.0 to be deployable. Either we need this fix or we need changes to the smoke tests so they actually work. I'm not sure which is better - what do you think?

I think #2114 should fix the smoke tests without changing the access to packages.
I think it might just be that v5.0.0 has a known limitation that you can only install the latest version of each allowed CRAN package in T3. (I _think_ that must have been true in v4 too).

jemrobinson:
Are you able to easily test whether #2114 actually *does* fix the smoke tests? If not, let me know and I'll try to deploy something.

data_safe_haven/infrastructure/programs/sre/software_repositories.py
jemrobinson: No, we need it. The automated Docker image updater looks for the latest release, not the latest minor release, so this change is made automatically.

jemrobinson:
Sounds like we might want this to go into release 5.0.0. Am I OK to change the target of this PR @JimMadge ?

JimMadge:
@craddm @jemrobinson are we happy to put this in for the release?

jemrobinson (approved):
LGTM

PR #2117, opened 2024-08-09
Add pip to workspace software

jemrobinson:

n/a

Add `python3-pip` to the workspaces

n/a

### :microscope: Tests

Not tested

### DIFF
--- a/data_safe_haven/resources/workspace/ansible/host_vars/localhost.yaml
+++ b/data_safe_haven/resources/workspace/ansible/host_vars/localhost.yaml
@@ -51,6 +51,7 @@ package_categories:
       - octave  # Open source Matlab implementation
       - python-is-python3  # symlinks python to python3
       - python3  # System Python3 distribution
+      - python3-pip  # Python3 package management tool
       - python3-venv  # Python3 venv module
       - r-base  # R programming language
       - racket  # Racket functional programming language implementation


JimMadge (changes requested):
I think we shouldn't do this. We should be encouraging people to use virtual environments rather than installing packages to their home directories.

jemrobinson:
Sorry, I'm confused here. Isn't this the package that gives you the `pip` executable? What's that got to do with virtual environments vs. home directories?

JimMadge:
`python-pip` provides pip globally and will by default install packages to `~/.local` or `/usr/`.

It shouldn't be needed because we have `python-venv` which will put `pip` and `setup-tools` in a virtual environment when you create it.
I think what we want to do is encourage people to always use `python -m venv` (or `hatch`, `poetry`, _etc._) then pip inside that venv, never the global `pip`.

jemrobinson:
Ah OK. So `python-venv` also provides `pip` then? I guess symlinked into the virtualenvs when you create them?

JimMadge:
Yes, I'm not 100% certain how that works, however the venv has pip and when you activate it pip will be added to your path.
So there will be pip in the venv but not globally.

(I think this is why pip always complains about being out of date now, because you have pip pinned to your python distribution so there is likely a newer version on PyPI).

jemrobinson:
The entire content of the `pip` executable is just:

```python
#!/home/james.robinson/testvenv/bin/python
# -*- coding: utf-8 -*-
import re
import sys
from pip._internal.cli.main import main
if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])
    sys.exit(main())
```

so (after testing) it looks like `python-venv` just writes that to `VENV_PATH/bin`. It must be installing the `pip` package to `site-packages` from **somewhere**, but I'm not sure where.

jemrobinson:
Working as intended.

Issue #2096, opened 2024-08-06
Tidy Ansible


JimMadge:
Related to #2095 

There is a lot that could be done to tidy up the desired state play.

- Use task files
- Reusable elements (roles, include_tasks)
- Blocks for related tasks
- Tags

PR #2121, opened 2024-08-12
Ansible: create icons directory

jemrobinson:

n/a

**IMPORTANT: I am confident that merging this will not affect the findings of the ongoing penetration test**

Create missing `/usr/local/share/icons` directory in Ansible

n/a

### :microscope: Tests

Tested on running SRE

### DIFF
--- a/data_safe_haven/resources/workspace/ansible/desired_state.yaml
+++ b/data_safe_haven/resources/workspace/ansible/desired_state.yaml
@@ -172,7 +172,7 @@
         regexp: '^TerminalEmulator='
         line: 'TerminalEmulator=xfce4-terminal'
 
-    - name: Set default terminal colourscheme
+    - name: Copy default terminal colourscheme
       ansible.builtin.copy:
         src: etc/xdg/xfce4/terminal/terminalrc
         dest: /etc/xdg/xfce4/terminal/terminalrc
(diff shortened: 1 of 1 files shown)


JimMadge (approved):
Did you check if 

```yaml
    - name: Copy icons
      ansible.builtin.copy:
        src: "usr/local/share/icons/{{ item }}"
        dest: /usr/local/share/icons/
        mode: '0444'
      loop:
        - gitea.png
        - hedgedoc.png
```

worked?

Either is way fine I think.

craddm:
> Did you check if
> 
> ```yaml
>     - name: Copy icons
>       ansible.builtin.copy:
>         src: "usr/local/share/icons/{{ item }}"
>         dest: /usr/local/share/icons/
>         mode: '0444'
>       loop:
>         - gitea.png
>         - hedgedoc.png
> ```
> 
> worked?
> 
> Either is way fine I think.

If you mean whether that alone worked with no further changes, that was failing on my VM until I manually created `/urs/local/share/icons/`

jemrobinson:
This seems to work

> If `dest` is a non-existent path and if either `dest` ends with “/” or `src` is a directory, `dest` is created.

I'll update the YAML and try again.

jemrobinson:
@JimMadge : I've simplified a few `ansible.builtin.copy` tasks to copy everything in `src` to `dest`. Unless there's any reason **not** to do this, I think it's cleaner and minimises the possibility of forgetting to copy a file.

jemrobinson:
@craddm : I think the trailing `/` is the thing that triggers whether or not the directory is created.

data_safe_haven/resources/workspace/ansible/desired_state.yaml
```diff
@@ -307,12 +297,12 @@
...
-    - name: Install ClamAV services and timer
+    # This will create directories if src or dest ends in '/'
+    - name: Copy ClamAV services directory
       ansible.builtin.copy:
-        src: "{{ item }}"
+        src: "etc/systemd/system/"
         dest: /etc/systemd/system/
         mode: '0644'
-      with_fileglob: 'etc/systemd/system/clamav*'
       notify: Systemd daemon reload
```
JimMadge, line 306: This one is fine for now, as there are no other systemd units but we should revert it if we add more.

This would copy all units and not just the clamav units.

jemrobinson: Yes - I guess if we add more we should change the name of the task.

JimMadge: Could be part of https://github.com/alan-turing-institute/data-safe-haven/issues/2096

PR #2129, opened 2024-08-14
List of SRE configs available in the current context

craddm:
## :white_check_mark: Checklist

- [x] You have given your pull request a meaningful title (_e.g._ `Enable foobar integration` rather than `515 foobar`).
- [x] You are targeting the appropriate branch. If you're not certain which one this is, it should be **`develop`**.
- [x] Your branch is up-to-date with the **target branch** (it probably was when you started, but it may have changed since then).

### :arrow_heading_up: Summary

Adds a command to list the available and deployed SRE configs within the currently selected context.

```
(data-safe-haven) deploydsh@0b11d89ae8c5:/workspaces/data-safe-haven$ dsh config available
You are logged into the Azure CLI as:                                                                                                      
        user: Matthew Craddock (1b2a59f2-f3ea-42e7-b8b8-aeea30572764)                                                                      
        tenant: turing.ac.uk (4395f4a7-e455-4f95-8a9f-1fbaef6384f9)                                                                        
Are these details correct? [y/n] (y): y
Loaded stack shm-turtles-sre-donatello.                                                                                                    
Loaded stack shm-turtles-sre-leonardo.                                                                                                     
Loaded stack shm-turtles-sre-shimazu.                                                                                                      
Available SRE configurations for context 'turtles':
┏━━━━━━━━━━━┳━━━━━━━━━━┓
┃ SRE Name  ┃ Deployed ┃
┡━━━━━━━━━━━╇━━━━━━━━━━┩
│ donatello │          │
│ leonardo  │ x        │
│ shimazu   │          │
└───────────┴──────────┘
```

### :closed_umbrella: Related issues

Closes #1812 

### :microscope: Tests

Tested locally on existing contexts/SHMs

PR #2132, opened 2024-08-15
Fixes for AzureSdk blob handling

jemrobinson:

n/a

- Ensure the `blob_client()` method is used everywhere we need a `BlobServiceClient`
- Ensure that exceptions raised by `blob_client()` are caught and (if appropriate) reraised by the calling function
- Make error messages more informative about what went wrong

n/a

### :microscope: Tests

Tested on an existing SRE

### DIFF
--- a/data_safe_haven/external/api/azure_sdk.py
+++ b/data_safe_haven/external/api/azure_sdk.py
@@ -124,7 +124,7 @@ class AzureSdk:
             f"DefaultEndpointsProtocol=https;AccountName={storage_account_name};AccountKey={storage_account_keys[0].value};EndpointSuffix=core.windows.net"
         )
         if not isinstance(blob_service_client, BlobServiceClient):
-            msg = f"Could not connect to storage account '{storage_account_name}'."
+            msg = f"Could not load blob client for storage account '{storage_account_name}'."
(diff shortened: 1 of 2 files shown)


craddm (commented):
In #2129 I need to use a blob client, but can't use this function as it requires a blob name. Is it possible to make it an optional parameter?

jemrobinson:
I think that would be fine. If you're happy with these changes otherwise, maybe the easiest solution is to merge this as-is, merge release-v5.0.0 into your branch for #2129 and then make the blob name optional there. What do you think?

jemrobinson:
Or actually, separate this function into a `blob_service_client` and `blob_client` function. I can do that here if you'd prefer?

jemrobinson:
@craddm : Does f14ae787e fix your issue?

craddm:
> @craddm : Does https://github.com/alan-turing-institute/data-safe-haven/commit/f14ae787e75ed81aa4218ef857899b2c18f4d67c fix your issue?

Tested this with some edits to #2129 and it works!

jemrobinson:
Great - are we OK to merge this one then?

craddm:
LGTM

PR #2133, opened 2024-08-15
Mandate minimum TLS version of 1.2 for all storage accounts

craddm:

As per pentest report, storage accounts in SREs currently allow a minimum TLS version of 1.0, which is deprecated.

Note that this also applies to the SHM storage account, and this PR also modifies the minimum version there.

### :microscope: Tests

Deployed SRE using these changes.

- [x] mounting the accounts inside the SRE
- [x] reading/writing from inside the SRE
- [x] accessing them from Azure Store Explorer


jemrobinson:
Is this a from-scratch SRE? I'd like to know whether this causes any problems for any of:

- mounting the accounts inside the SRE
- reading/writing (as appropriate) from inside the SRE
- accessing them from Azure Storage Explorer

Could you give details on this issue when you have them?

craddm:
> Is this a from-scratch SRE? I'd like to know whether this causes any problems for any of:
> 
> * mounting the accounts inside the SRE
> * reading/writing (as appropriate) from inside the SRE
> * accessing them from Azure Storage Explorer
> 
> Could you give details on this issue when you have them?

It isn't a from scratch SRE - I deleted and redeployed the storage accounts. However, currently having issues restarting the `gitea` container group. Going to try a fresh deployment instead.

craddm:
> Is this a from-scratch SRE? I'd like to know whether this causes any problems for any of:
> 
> * mounting the accounts inside the SRE
> * reading/writing (as appropriate) from inside the SRE
> * accessing them from Azure Storage Explorer
> 
> Could you give details on this issue when you have them?

All worked fine as far as I can tell. Successfully uploaded and downloaded from ingress and egress via storage explore. Successfully copied from input and to output from the SRE desktop

jemrobinson:
Cool - that's /ingress and /egress. Can you check /home (presumably working if you logged in) and /shared as well?

data_safe_haven/external/api/azure_sdk.py
```diff
@@ -581,6 +581,7 @@ def ensure_storage_account(
                     kind=StorageAccountKind.STORAGE_V2,
                     location=location,
                     tags=tags,
+                    minimum_tls_version="TLS1_2",
```
JimMadge: Is there an enum from the Azure SDK we can use here instead of a raw string?

JimMadge (commented):
LGTM. We should be careful to test this thoroughly though as we are merging to the release branch.

data_safe_haven/external/api/azure_sdk.py
jemrobinson: `storage.MinimumTlsVersion`

JimMadge: Is that `pulumi.azure_native`? I think this is the Python Azure SDK.

jemrobinson: Sorry - you're right!

jemrobinson: It's from `azure.mgmt.storage.v2021_08_01.models import MinimumTlsVersion`

craddm: done

jemrobinson:
@JimMadge : I'm comfortable that Matt has tested /data and /output (also /home and maybe /shared). Do you think we need/want an independent test of these?

craddm:
> @JimMadge : I'm comfortable that Matt has tested /data and /output (also /home and maybe /shared). Do you think we need/want an independent test of these?

Yes, home and shared working fine too 

Just now deployed an SHM with the updated TLS too

jemrobinson:
Happy to merge this @JimMadge ?

JimMadge (approved):

PR #2143, opened 2024-08-19
Add --tier option to provide default settings

jemrobinson:

n/a

Add --tier option to provide default settings

n/a

### :microscope: Tests

Tested on new config generation

### DIFF
--- a/data_safe_haven/config/sre_config.py
+++ b/data_safe_haven/config/sre_config.py
@@ -47,11 +47,11 @@ class SREConfig(AzureSerialisableModel):
     def template(cls: type[Self], tier: int | None = None) -> SREConfig:
         """Create SREConfig without validation to allow "replace me" prompts."""
         # Set tier-dependent defaults
-        if tier == 0:  # noqa: PLR2004
+        if tier == 0:
             remote_desktop_allow_copy = True
             remote_desktop_allow_paste = True
             software_packages = SoftwarePackageCategory.ANY
-        elif tier == 1:  # noqa: PLR2004
+        elif tier == 1:
             remote_desktop_allow_copy = True
             remote_desktop_allow_paste = True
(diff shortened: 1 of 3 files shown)


JimMadge (commented):
We should have some tests for this behaviour.

Is this critical for v5.0.0? If not, it might be best to put this in a patch release if it isn't ready to go.

jemrobinson:
@JimMadge : I don't think it's essential and there are a few things needed here anyway (tests, use an enum). Mainly I think we should have *something* in the docs that explains how to deploy a tier 0/1/2/3/4 environment. Doesn't have to be this, but we should have some documentation instead in that case.

data_safe_haven/config/sre_config.py
```diff
@@ -44,8 +44,34 @@ def from_remote_by_name(
         return cls.from_remote(context, filename=sre_config_name(sre_name))
 
     @classmethod
-    def template(cls: type[Self]) -> SREConfig:
+    def template(cls: type[Self], tier: int | None = None) -> SREConfig:
         """Create SREConfig without validation to allow "replace me" prompts."""
+        # Set tier-dependent defaults
+        if tier == 0:  # noqa: PLR2004
```
jemrobinson:
```suggestion
        if tier == 0:
```

data_safe_haven/config/sre_config.py
```diff
@@ -44,8 +44,34 @@ def from_remote_by_name(
...
+            remote_desktop_allow_copy = True
+            remote_desktop_allow_paste = True
+            software_packages = SoftwarePackageCategory.ANY
+        elif tier == 1:  # noqa: PLR2004
```
jemrobinson:
```suggestion
        elif tier == 1:
```

JimMadge (approved):

PR #2149, opened 2024-08-20
Fix Guacamole copy/paste configuration

JimMadge:

### :microscope: Tests

Tested on a fresh deployment. Unable to copy/paste to/from local machine and workspace.


craddm (approved):
Confirmed that the connection settings in guacamole now show appropriate configuration:

image

Testing shows that the user can be no longer copy or paste into or out of the workspace, but can copy and paste within it (e.g. from one terminal window to another)

Issue #2084, opened 2024-08-05
Suggest users install with pip install of using hatch


jemrobinson:
## :white_check_mark: Checklist

- [x] I have searched open and closed issues for duplicates
- [x] This is a request for a change to documentation
- [x] This isn't a feature request or bug report
- [x] This isn't an open-ended question (open a [discussion](https://github.com/alan-turing-institute/data-safe-haven/discussions) if it is).

## :ledger: Where is the problem?

Using `hatch` is overly complicated for end users. We should suggest they install with pip instead.

## :memo: Required documentation changes

Update hatch references

JimMadge:
A few things we need to be careful about,

- (How) do we enforce a strict set of pinned dependencies like in the requirements.txt.
  That is currently don't with a hatch pre-install hook
- I think we have files that wouldn't be included in a wheel build (non-python files in `data_safe_haven`)

PR #2154, opened 2024-08-22
Pin pyproject dependencies

jemrobinson:

n/a

- Pin Python package dependencies in `pyproject.toml`
- Drop custom pre-installation of pinned versions
- Drop custom updating scripts in favour of `dependabot`
- Use recommended structure for environment requirements: group them into optional feature groups which means they can (if desired) be installed with e.g. `pip install ".[test]"`

Justification:
- "Should I pin my dependencies?" discussion [here](https://stackoverflow.com/questions/28509481/should-i-pin-my-python-dependencies-versions) and [here](https://discuss.python.org/t/should-i-be-pinning-my-dependencies/13159)
  - summary: yes, if you're writing an application that isn't expecting other packages to depend on it

Potential issues:
- ~This only pins direct dependencies, not the things that those packages depend on~
  - I think this is OK, as this shouldn't affect the functions we're calling. If it does, then we probably have a (hidden) direct dependency and should probably add that to `pyproject.toml`
  - Solved by using `hatch-pip-compile` which is a built-in version of what we were previously doing manually
- Dependabot might try to update the pinned `requirements` files
  - Not sure whether this is a problem or not

Part of #2084 

### :microscope: Tests

Tested on a fork of this repository - dependabot is updating `pyproject.toml` as expected.


JimMadge (commented):
I don't think this is the way I want to go.

We loose the pinned dependencies here as we only have direct dependencies. That would make debugging more difficult and docs builds unreproducible.

I think there is an advantage to having the minimal dependencies for development (for example, when we had to set a minimum patch version for `pulumi-azure-native`) and the full pinned dependencies. `pip-tools` feels like a good, hosting agnostic way to do that.

A better solution might be to figure out how to support a `requirements.in` with the full dependencies going to `pyproject.toml` rather than `requirement.txt`

jemrobinson:
In principle pinning direct dependencies *should* be OK. If any of the functions from our directly-called libraries change their behaviour because of the specific version of *their* dependencies then we actually depend on that library too and we should be explicit about that.

I disagree that having minimal dependencies listed in `pyproject.toml` is helpful. We actually have no idea whether the set of constraints in this file were correct, because we never used them in development - we were always installing from `requirements.txt`.

jemrobinson:
@JimMadge : Transitive dependencies now pinned using `hatch-pip-compile`

pyproject.toml
```diff
@@ -12,42 +12,68 @@ authors = [
...
-  "pydantic>=2.4",
-  "pyjwt>=2.8",
-  "pytz>=2023.3",
-  "PyYAML>=6.0",
-  "rich>=13.4",
-  "simple-acme-dns>=3.0",
-  "typer>=0.9",
-  "websocket-client>=1.5",
+  "appdirs==1.4.4",
+  "azure-core==1.30.2",
```
JimMadge, line 16: I still don't see the advantage of pinning dependencies both here and in the `requirements.txt` files. Won't this just make it more difficult to resolve a full set of dependencies?

You can't tell which are these are strict `==` and which at actually `>=` some version.

pyproject.toml
```diff
@@ -12,42 +12,68 @@ authors = [
...
+]
+
+[project.optional-dependencies]
+docs = [
+  "emoji==2.12.1",
+  "myst-parser==4.0.0",
+  "pydata-sphinx-theme==0.15.4",
+  "sphinx-togglebutton==0.3.2",
+  "sphinx==8.0.2",
+]
```
JimMadge, line 56: I intentionally avoided this before as the features are in addition to the package dependencies.

You don't need any of those to build the documentation. Does this mean that our instructions for building the docs now involves installing the full set of dependencies for `dsh`?

Installing the Pulumi packages takes quite a while so I think it would be best to avoid that if we can.

pyproject.toml
```diff
@@ -12,42 +12,68 @@ authors = [
...
+  "mypy==1.10.1",
+  "pandas-stubs==2.2.2.240807",
+  "pydantic==2.8.2",
+  "ruff==0.5.0",
+  "types-appdirs==1.4.3.5",
+  "types-chevron==0.14.2.20240310",
+  "types-pytz==2024.1.0.20240417",
+  "types-pyyaml==6.0.12.20240311",
+  "types-requests==2.32.0.20240622",
+]
```
JimMadge, line 70: The same here, I don't _think_ the core dependencies are needed for the linting scripts.

JimMadge (changes requested):

pyproject.toml
jemrobinson: - the point of pinning requirements here is to have a single place where the direct requirements of this project are pinned
- the requirements files are just a way to ensure that the indirect requirements are also pinned (using `pip-compile`)

Ideally there would be a way to pin the indirect dependencies in `pyproject.toml` too, but this method is the one recommended by the `hatch` developers (see [here](https://github.com/pypa/hatch/issues/749))

> You can't tell which are these are strict == and which at actually >= some version.

Everything is strict `==` since we want installing this package to require the installation of a strict set of dependencies.

pyproject.toml
jemrobinson: The pattern of using `features` for this comes from a suggestion from the `hatch` developers [here](https://github.com/pypa/hatch/issues/1165).

If you do `pip install project_name[docs]` then you'll get both sets of packages, but adding it as a `feature` just installs this set:

```
$ hatch -e docs run pip freeze
accessible-pygments==0.0.5
alabaster==1.0.0
babel==2.16.0
beautifulsoup4==4.12.3
build==1.2.1
certifi==2024.7.4
charset-normalizer==3.3.2
click==8.1.7
docutils==0.21.2
emoji==2.12.1
idna==3.7
imagesize==1.4.1
Jinja2==3.1.4
markdown-it-py==3.0.0
MarkupSafe==2.1.5
mdit-py-plugins==0.4.1
mdurl==0.1.2
myst-parser==4.0.0
packaging==24.1
pip-tools==7.4.1
pydata-sphinx-theme==0.15.4
Pygments==2.18.0
pyproject_hooks==1.1.0
PyYAML==6.0.2
requests==2.32.3
setuptools==73.0.1
snowballstemmer==2.2.0
soupsieve==2.6
Sphinx==8.0.2
sphinx-togglebutton==0.3.2
sphinxcontrib-applehelp==2.0.0
sphinxcontrib-devhelp==2.0.0
sphinxcontrib-htmlhelp==2.1.0
sphinxcontrib-jsmath==1.0.1
sphinxcontrib-qthelp==2.0.0
sphinxcontrib-serializinghtml==2.0.0
typing_extensions==4.12.2
urllib3==2.2.2
wheel==0.44.0
```

pyproject.toml
jemrobinson: As above - this won't install the core dependencies.

<details>
<summary>Package list</summary

```
$ hatch -e lint run pip freeze
annotated-types==0.7.0
ansible==10.2.0
ansible-builder==3.1.0
ansible-compat==24.8.0
ansible-core==2.17.3
ansible-creator==24.7.1
ansible-dev-environment==24.7.0
ansible-dev-tools==24.7.2
ansible-lint==24.7.0
ansible-navigator==24.8.0
ansible-runner==2.4.0
ansible-sign==0.1.1
attrs==24.2.0
bindep==2.11.0
black==24.4.2
bracex==2.5
build==1.2.1
cachetools==5.5.0
cffi==1.17.0
chardet==5.2.0
click==8.1.7
click-help-colors==0.9.4
colorama==0.4.6
cryptography==43.0.0
distlib==0.3.8
distro==1.9.0
docutils==0.21.2
enrich==1.2.7
execnet==2.1.1
filelock==3.15.4
importlib_metadata==8.4.0
iniconfig==2.0.0
Jinja2==3.1.4
jsonschema==4.23.0
jsonschema-specifications==2023.12.1
lockfile==0.12.2
markdown-it-py==3.0.0
MarkupSafe==2.1.5
mdurl==0.1.2
molecule==24.8.0
mypy==1.10.1
mypy-extensions==1.0.0
numpy==2.1.0
onigurumacffi==1.3.0
packaging==24.1
pandas-stubs==2.2.2.240807
Parsley==1.3
pathspec==0.12.1
pbr==6.0.0
pexpect==4.9.0
pip-tools==7.4.1
platformdirs==4.2.2
pluggy==1.5.0
ptyprocess==0.7.0
pycparser==2.22
pydantic==2.8.2
pydantic_core==2.20.1
Pygments==2.18.0
pyproject-api==1.7.1
pyproject_hooks==1.1.0
pytest==8.3.2
pytest-ansible==24.8.0
pytest-xdist==3.6.1
python-daemon==3.0.1
python-gnupg==0.5.2
PyYAML==6.0.2
referencing==0.35.1
resolvelib==1.0.1
rich==13.7.1
rpds-py==0.20.0
ruamel.yaml==0.18.6
ruamel.yaml.clib==0.2.8
ruff==0.5.0
setuptools==73.0.1
subprocess-tee==0.4.2
tox==4.18.0
tox-ansible==24.8.0
types-appdirs==1.4.3.5
types-chevron==0.14.2.20240310
types-pytz==2024.1.0.20240417
types-PyYAML==6.0.12.20240311
types-requests==2.32.0.20240622
typing_extensions==4.12.2
tzdata==2024.1
urllib3==2.2.2
virtualenv==20.26.3
wcmatch==9.0
wheel==0.44.0
yamllint==1.35.1
zipp==3.20.0
```
</details>

pyproject.toml
jemrobinson: It's a bit of a conceptual shift away from "which versions of our dependencies will work" to "which versions of our dependencies to we want users to install". It wouldn't make sense if we were expecting people to run `from package_name import whatever` but it does for a package that we want to be installed in its own, dedicated venv.

JimMadge: If there were a way to get `pip install .` and `hatch env create` to have _exactly_ that same dependencies (without putting a full pinned list in `pyproject.toml`), that would be good. It feels like that might not be an option though.

JimMadge: We sort of want the opposite of what we are doing here (like the pip-tools docs recommends) a `requirements.in` which we use to generate the dependencies in `pyproject.toml`.

We _could_ presumably write that ourselves. Using `pip-compile` as a library to generate the pinned dependencies, then a toml library to write that list to `pyproject.toml`.

JimMadge (approved):
Happy to merge this as it doesn't look like there is a sensible way to do what we would ideally want.

We can review that again in the future.

PR #2196, opened 2024-09-24
Raise exception when admin group name is not found

### LINKED ISSUE #2187: No clear error message when admin group name is incorrect in context

- **Operating System:** macOs
- **Data Safe Haven version:** develop

## :package: Packages

<details>

```none
Paste list of packages here
```

</details>

Deploying an SHM fails if the `admin-group-name` does not correspond to a known security group. This situation is already captured in the codebase, but does not return an error. It'd be much easier for a user to diagnose this problem if an appropriate error message were returned.

<details>

```none
        user: Matthew Craddock (1b2a59f2-f3ea-42e7-b8b8-aeea30572764)                                                                      
        tenant: turing.ac.uk (4395f4a7-e455-4f95-8a9f-1fbaef6384f9)                                                                        
Preparing to deploy testadmingroup SHM.                                                                                                    
Ensured that resource group shm-checkadmin-rg exists in uksouth.                                                                           
Ensured that managed identity shm-checkadmin-identity-reader exists.                                                                       
Ensured that storage account shmcheckadmin exists.                                                                                         
Ensured that storage container config exists.                                                                                              
Ensured that storage container pulumi exists.                                                                                              
Failed to create key vault shm-checkadmin-kv.                                                                                              
Failed to deploy resources needed by Pulumi.                                                                                               
Failed to deploy Data Safe Haven infrastructure. 
```

</details>

## :recycle: To reproduce

Deploy an SHM after setting up a context with an incorrect `admin-group-name`

craddm:

Raises an exception when the admin group name is specified incorrectly in the context. 

Closes #2187 

### :microscope: Tests

Tested locally


data_safe_haven/config/shm_config.py
```diff
@@ -27,10 +28,14 @@ def from_args(
     ) -> SHMConfig:
         """Construct an SHMConfig from arguments."""
         azure_sdk = AzureSdk(subscription_name=context.subscription_name)
-        admin_group_id = (
-            azure_sdk.entra_directory.get_id_from_groupname(context.admin_group_name)
-            or "admin-group-id-not-found"
-        )
```
JimMadge, line 33: Do we want to continue here if the id cannot be found?

JimMadge: Looks like this change might have been added to satisfy mypy as the function can return `None` https://github.com/alan-turing-institute/data-safe-haven/commit/b5e08768fc2cca8d52ff84b728c36c21a54b5792

JimMadge:
Returning `None` is important [here](https://github.com/alan-turing-institute/data-safe-haven/blob/cdd76a3b51a8278ab1562f90ab8637ad8ab5276b/data_safe_haven/external/api/graph_api.py#L317) where if there is no group then one is created.

JimMadge:
Also used in `remove_user_from_group` and `add_user_to_group` where `None` is not handled.

A good solution could be adding a method like

```python
def ensure_id_from_groupname(name):
   if (id := get_id_from_groupname(name)):
       return id
   else:
       raise Exception
```

data_safe_haven/config/shm_config.py
```diff
@@ -27,10 +28,14 @@ def from_args(
...
+        try:
+            admin_group_id = azure_sdk.entra_directory.get_id_from_groupname(
+                context.admin_group_name
+            )
+        except DataSafeHavenMicrosoftGraphError as exc:
+            msg = f"Failed to get ID for group '{context.admin_group_name}'"
+            raise DataSafeHavenMicrosoftGraphError(msg) from exc
+
```
jemrobinson: How about this instead of adding a new method?

```suggestion
        admin_group_id := azure_sdk.entra_directory.get_id_from_groupname(
            context.admin_group_name
        )
        if not admin_group_id:
            msg = f"Failed to get ID for group '{context.admin_group_name}'"
            raise DataSafeHavenMicrosoftGraphError(msg) from exc

```

Then we can leave the function as returning `str | None` and avoid needing to add exception handling. Note that this will also throw the exception if the ID of the group is `""` but I think that's reasonable, since a valid ID will be a GUID.

data_safe_haven/external/api/graph_api.py
```diff
@@ -515,7 +515,7 @@ def get_service_principal_by_name(
         except (DataSafeHavenMicrosoftGraphError, StopIteration):
             return None
 
-    def get_id_from_groupname(self, group_name: str) -> str | None:
+    def get_id_from_groupname(self, group_name: str) -> str:
```
jemrobinson: If taking my suggestion above, we should revert this change.

data_safe_haven/config/shm_config.py
JimMadge: Won't this raise the exception if `admin_group_id` is `None`? `not None is True`

craddm:
> Also used in `remove_user_from_group` and `add_user_to_group` where `None` is not handled.
> 
> A good solution could be adding a method like
> 
> ```python
> def ensure_id_from_groupname(name):
>    if (id := get_id_from_groupname(name)):
>        return id
>    else:
>        raise Exception
> ```

This feels weird to me.

We have a function called `get_id_from_groupname` which gets the id if the group exists or returns `None`, and now a function called `ensure_id_from_groupname` which also gets the id if the group exists, but throws an exception if it doesn't exist. I feel like that is a bit confusing, especially given that (almost) all the other `ensure_*` functions check if something exists and create it if it doesn't.

It feels like we should have a function that checks if the group exists, and then one that returns the group id (but is only used when we know the group exists). The slight problem with that is the way to check if the group exists is to try to retrieve the whole list of groups and search through it, which you then would have to do multiple times instead of once.

JimMadge:
This is the internal API so I wouldn't worry _too_ much about names.

You could change how the current `create_group` works to return the name if it exists but that might be too much for one function.

Making two API calls when you only need one feels wasteful.
I still think just adding a method for the common pattern of ensuring the groups exists makes most sense. It is just taking the code that would be duplicated in a number of places and instead writing it once.

craddm:
> This is the internal API so I wouldn't worry _too_ much about names.
> 
> You could change how the current `create_group` works to return the name if it exists but that might be too much for one function.
> 
> Making two API calls when you only need one feels wasteful. I still think just adding a method for the common pattern of ensuring the groups exists makes most sense. It is just taking the code that would be duplicated in a number of places and instead writing it once.

Agree about the multiple API calls, which would be easy enough to stop by instead passing the list of groups to the functions from a separate call to `read_groups()`. But anyway, gone back to your suggested style

data_safe_haven/config/shm_config.py
jemrobinson: Yes, if it's `None` (the lookup failed) or `""` (the lookup returned empty string, which shouldn't be possible). That's what we want, right?

JimMadge (approved):
LGTM :+1:

PR #2206, opened 2024-09-25
Update devcontainer

craddm:

Since Sphinx and document building are now handled by `hatch`, it is no longer necessary to install sphinx separately using `apt`. In addition, since `requirements.txt` is no longer in the docs folder, the devcontainer could not successfully build.

Also updates to the most recent version of Azure CLI.

### :microscope: Tests

Tested locally


JimMadge (approved):
LGTM

PR #2226, opened 2024-10-07
Remove desktop files for gitea/hedgedoc

JimMadge:
These are already provided by templates.

Removes desktop files for Gitea and Hedgedoc, which are already provided by [templates](https://github.com/alan-turing-institute/data-safe-haven/blob/c5ce29f431524c02b63973dac527e07358c3d5b5/data_safe_haven/resources/workspace/ansible/tasks/xfce.yaml#L60).

These files would break the idempotence of the desired state play.

### DIFF
--- a/data_safe_haven/resources/workspace/ansible/files/etc/skel/Desktop/gitea.desktop
+++ b/data_safe_haven/resources/workspace/ansible/files/etc/skel/Desktop/gitea.desktop
@@ -1,6 +1,0 @@
-[Desktop Entry]
-Version=1.0
-Type=Link
-Name=Gitea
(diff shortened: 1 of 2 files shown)


jemrobinson:
What about the ingress/egress/shared desktop files? Should they live here?

JimMadge:
> What about the ingress/egress/shared desktop files? Should they live here?

Yes, those ones are just files that get copied, there are no templates.

(We needed templates for the services as their URLs are not known before run time. The files this PR removes wouldn't work as they are missing that field.)

craddm (approved):
LGTM

PR #2234, opened 2024-10-11
Simplify code for checking config availability and SRE deployment status

craddm:

Simplifies the code for checking which configs are available on the SHM and which SREs are deployed. No longer needs to load each stack, and thus is much faster.

### :microscope: Tests

Tested locally


JimMadge (approved):
Very nice, good use of the Pulumi SDK to avoid reinventing the wheel.

PR #2237, opened 2024-10-16
Unchangable Pulumi workspace configuration

### LINKED ISSUE #2199: Changing SRE subscription ID
The subscription ID for an SRE is set in the configuration file.
When and SRE is deployed, this is written to the `projects` hash map in `pulumi.yaml` in Azure storage (I believe this is the equivalent on `pulumi.stack.yaml` for local storage).

If you change the subscription ID in the configuration, the remote value is _not_ updated.

We could,

- Update this value and allow changing an SRE's subscription
  (This would likely mean deleting and recreating most or all of the resources)
- Prevent moving subscriptions
  In which case, we shouldn't give users the option to change this configuration value

JimMadge:

#2241 

Protects against changing values which we have declared shouldn't be updated in the Pulumi workspace.

closes #2199 

### :microscope: Tests

```console
❯ hatch run dsh sre deploy hojo
        user: Jim Madge (…)
        tenant: turing.ac.uk (…)
You are logged into the Microsoft Graph API as:
        user: aad.admin.jim.madge@green.develop.turingsafehaven.ac.uk
(…)
        tenant: green.develop.turingsafehaven.ac.uk (…)
SRE will be deployed to subscription 'Data Safe Haven Development 2'
('…')
SRE will be registered in SHM 'daimyo.develop.turingsafehaven.ac.uk'
SHM subscription '…'
Loaded stack shm-daimyo-sre-hojo.
Unchangeable configuration variable 'azure-native:subscriptionId' not consistent, your configuration: '…', Pulumi workspace: '…'.
Applying Pulumi configuration options failed..
Pulumi deployment failed.
Could not deploy Secure Research Environment 'hojo'.
```

### DIFF
--- a/data_safe_haven/infrastructure/project_manager.py
+++ b/data_safe_haven/infrastructure/project_manager.py
@@ -286,12 +286,22 @@ class ProjectManager:
             raise DataSafeHavenPulumiError(msg) from exc
 
     def ensure_config(self, name: str, value: str, *, secret: bool) -> None:
-        """Ensure that config values have been set, setting them if they do not exist"""
+        """
+        Ensure that config values have been set.
+
+        Values will be set if they do not exist.
+
+        If the value is already set and does not match the `value` argument,
+        `DataSafeHavenPulumiError` will be raised.
+        """
         try:
             existing_value = self.stack.get_config(name).value
         except automation.CommandError:
+            # Set value if it does not already exist
             self.set_config(name, value, secret=secret)
 
+        # If the value does already exist, ensure it is consistent with the declared
+        # value
         if existing_value != value:
             msg = (
                 f"Unchangeable configuration option '{name}' not consistent, "
--- a/data_safe_haven/infrastructure/project_manager.py
+++ b/data_safe_haven/infrastructure/project_manager.py
@@ -294,7 +294,7 @@ class ProjectManager:
 
         if existing_value != value:
             msg = (
-                f"Unchangeable configuration variable '{name}' not consistent, "
+                f"Unchangeable configuration option '{name}' not consistent, "
                 f"your configuration: '{value}', Pulumi workspace: '{existing_value}'."
             )
             raise DataSafeHavenPulumiError(msg)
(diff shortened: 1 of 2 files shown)


data_safe_haven/infrastructure/project_manager.py
```diff
@@ -288,10 +288,17 @@ def destroy(self) -> None:
...
         except automation.CommandError:
             self.set_config(name, value, secret=secret)
 
+        if existing_value != value:
+            msg = (
+                f"Unchangeable configuration variable '{name}' not consistent, "
+                f"your configuration: '{value}', Pulumi workspace: '{existing_value}'."
+            )
+            raise DataSafeHavenPulumiError(msg)
+
```
jemrobinson: Is this the case for every current use of `ensure_config` or do we want a separate command for just those options which are immutable?

jemrobinson (changes requested):

data_safe_haven/infrastructure/project_manager.py
JimMadge: Not certain yet. However, none of those values would get updated if they were changed so if this change doesn't work I think it would mean there is a problem elsewhere.

JimMadge: The options with `replace==False` are

- location
- subscription id
- tenant id

To change any of those would require many resources to be destroyed then newly created in the new sub/location/tenant.
I think it is safe to raise an exception in those cases as the most sensible thing to do would be to teardown and deploy a new SRE.

jemrobinson: OK. Is it worth renaming the function and/or updating the docstring to better indicate that these are immutable values?

JimMadge: I'm happy with the name, the docstring/comments could do with updating though.

JimMadge: Done in b16256f

jemrobinson (approved):
LGTM

data_safe_haven/infrastructure/project_manager.py
```diff
@@ -286,12 +286,29 @@ def destroy(self) -> None:
             raise DataSafeHavenPulumiError(msg) from exc
 
     def ensure_config(self, name: str, value: str, *, secret: bool) -> None:
```
jemrobinson, line 288: Just wondering whether `ensure` is still the best verb for this. Can you think of a better one @JimMadge ?

JimMadge: Ensure is what I would choose, so I might not be the best person to ask :smile:

PR #2238, opened 2024-10-16
Add command reference to documentation

craddm:

Adds a new top-level section to the documentation. This section shows the documentation for all of the available `dsh` commands using Typer rich console formatting.

### :microscope: Tests

Documentation only


JimMadge (commented):
These look really nice. On my work computer they seem to render with MacOS close, minimise, the-other-one buttons. Does that chang e based on your OS?

It would be good to add some cross references to these pages from the deployment guide and management guide.

docs/source/index.md
```diff
@@ -9,6 +9,7 @@ design/index.md
 deployment/index.md
 management/index.md
 roles/index.md
+reference/index.md
```
JimMadge, line 12: Maybe we should call the section CLI reference? I'm not super opinionated though.

craddm:
> These look really nice. On my work computer they seem to render with MacOS close, minimise, the-other-one buttons. Does that chang e based on your OS?

It doesn't change, no. I'm not super-satisfied with that styling but it's not completely clear to me how to get rid of it. It looks like it may be possible by telling it to render directly to html rather than the default `svg`. I'll give it a go.

> It would be good to add some cross references to these pages from the deployment guide and management guide.

👍

docs/source/conf.py
```diff
@@ -5,7 +5,9 @@
 # https://www.sphinx-doc.org/en/master/usage/configuration.html
 import datetime
 import emoji
-
+import sys
+import os
+sys.path.insert(0, os.path.abspath(os.path.join("..", "..")))
```
jemrobinson: I really don't like this. Is there a better way to do it?

pyproject.toml
```diff
@@ -120,7 +121,6 @@ lock-filename = ".hatch/requirements.txt"
 [tool.hatch.envs.docs]
 type = "pip-compile"
 lock-filename = ".hatch/requirements-docs.txt"
```
jemrobinson, line 123: Do we definitely need all the production dependencies installed for doc-generation to work?

JimMadge: Ah, I forgot to mention that.

It is a shame to add more dependency syncing. But I think it is worth it to get CLI reference docs from the code. That way we only need to write an explanation of the commands/options once.

docs/source/conf.py
JimMadge:
```suggestion
from pathlib import Path
import sys

sys.path.insert(0, Path(__file__).parent.parent.absolute())
```

Gets rid of the old-style `os` module.
I presume we need to add the src directory to the path though so the Sphinx addon can parse the commands.
Could we not add `../../data-safe-haven` instead then?

jemrobinson: Since `docs` is in a separate directory, can we not `pip install .` as part of setting up this `hatch` environment and then just `import data_safe_haven`?

JimMadge: Possibly. I'm guessing the important thing is that `data_safe_haven` is in the Python path.

Not sure how you would represent that in the dependencies.

Speaking of which, how has the lockfile been updated when the docs env in `pyproject.toml` doesn't have all of those dependencies?

jemrobinson: Looks like `hatch` automatically installs the package into its environment unless you specify `detached = True`.

```
> hatch -e default shell
> pip freeze
...
cryptography==43.0.3
-e git+ssh://git@github.com/jemrobinson/data-safe-haven.git@6586af9affce845574a799a07e6ea03691b9c298#egg=data_safe_haven
debugpy==1.8.7
...
```

So we should be able to drop the `sys.path.insert` stuff in favour of `import data_safe_haven` and get the local checked-out version.

craddm: Yes, I removed `detached = True` from `[tool.hatch.envs.docs]` in `pyproject.toml` previously. `import data_safe_haven` doesn't work for me, locally - it still complains the module is not found; perhaps I'm missing something.

jemrobinson: @craddm, so if you do the following it doesn't work?

```
> hatch -e docs shell
> python
>>> import data_safe_haven
```

craddm: no, that works, but the docs won't build - to be clear, where should import data_safe_haven be going? In the `conf.py` or somewhere else?

JimMadge: What about things like `import data_safe_haven.commands.sre` like in the directives?

craddm: > What about things like `import data_safe_haven.commands.sre` like in the directives?

Which directives exactly? the Typer ones don't permit any content

jemrobinson: OK, let's take a step back.

You're saying that there's an error that can only be solved by manually adding the current directory to the path (which lets you do `from data_safe_haven import X`)

However, it also looks like you're already able to run `from data_safe_haven import X` in the `docs` environment.

What's the actual error message that you get if you remove the `sys.path.insert` line?

craddm:
```
/workspaces/data-safe-haven/docs/source/reference/config.md:5: CRITICAL: Failed to import "data_safe_haven.commands.config:config_command_group"The following exception was raised:
Traceback (most recent call last):
  File "/home/deploydsh/.local/share/hatch/env/pip-compile/data-safe-haven/Ce8xTVjM/docs/lib/python3.12/site-packages/sphinxcontrib/typer/__init__.py", line 309, in import_object
    obj = import_module(try_path)
          ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/importlib/__init__.py", line 90, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1324, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'data_safe_haven' [docutils]
```

JimMadge: I mean these directives,

```markdown
:::{typer} data_safe_haven.commands.sre:sre_command_group
:width: 65
:prog: dsh sre
:show-nested:
:make-sections:
:::
```

It looks a bit like a Python import so I'm wondering whether `import data_safe_haven` is fine but `import data_safe_haven.commands.sre` isn't for some reason.
If not, it might be doing something more low level that relies on looking for files in the Python path.

craddm: > I mean these directives,
> 
> ```
> :::{typer} data_safe_haven.commands.sre:sre_command_group
> :width: 65
> :prog: dsh sre
> :show-nested:
> :make-sections:
> :::
> ```
> 
> It looks a bit like a Python import so I'm wondering whether `import data_safe_haven` is fine but `import data_safe_haven.commands.sre` isn't for some reason. If not, it might be doing something more low level that relies on looking for files in the Python path.

~~Yes, those directives don't allow any content. I'm not sure where I would put `import data_safe_haven`.~~ ah i get it now

jemrobinson: Does `/home/deploydsh/.local/share/hatch/env/pip-compile/data-safe-haven/Ce8xTVjM/docs/lib/python3.12/site-packages/` have a package called `data_safe_haven` in it?

craddm:
```
/workspaces/data-safe-haven/docs/source/reference/config.md:5: CRITICAL: Failed to import "data_safe_haven:config_command_group"The following exception was raised:
Traceback (most recent call last):
  File "/home/deploydsh/.local/share/hatch/env/pip-compile/data-safe-haven/Ce8xTVjM/docs/lib/python3.12/site-packages/sphinxcontrib/typer/__init__.py", line 309, in import_object
    obj = import_module(try_path)
          ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/importlib/__init__.py", line 90, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1324, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'data_safe_haven' [docutils]
```

So, no.

craddm: > Does `/home/deploydsh/.local/share/hatch/env/pip-compile/data-safe-haven/Ce8xTVjM/docs/lib/python3.12/site-packages/` have a package called `data_safe_haven` in it?

No

jemrobinson: OK, I just checked out this branch and did the following:

1. Removed the `sys.path.insert` line

2. Ran `hatch run docs:build`

This worked as expected.

If I now look at the `docs` environment I see the `data_safe_haven` package is installed there.

```
> ls -1 /Users/jrobinson/Library/Application\ Support/hatch/env/pip-compile/data-safe-haven/sPksTL1q/docs/lib/python3.12/site-packages/
Arpeggio-2.0.2.dist-info
CHANGELOG.rst
CONTRIBUTING.md
...
_data_safe_haven.pth
...
data_safe_haven-5.1.0.dist-info
...
```

Can you try completely deleting the folder where your `docs` hatch environment is stored and retry without the `sys.path.insert` line? This should rebuild the environment.

JimMadge: > > Does `/home/deploydsh/.local/share/hatch/env/pip-compile/data-safe-haven/Ce8xTVjM/docs/lib/python3.12/site-packages/` have a package called `data_safe_haven` in it?
> 
> No

That is strange, would suggest the `detached=False` isn't working or the env is out of date (although hatch should check for that each time you run a command or script :shrug:).

With what @jemrobinson said it might just mean recreating the env `hatch env remove docs && hatch run docs:build`.

craddm: Looks like it doesn't update installed versions of the module after creating the environment (and apparently doesn't install it if you switch from `detached=True` to `detached=False`) https://github.com/pypa/hatch/issues/771

craddm: So yes, it works now I've cleared and recreated the environment!

craddm:
> > These look really nice. On my work computer they seem to render with MacOS close, minimise, the-other-one buttons. Does that chang e based on your OS?
> 
> It doesn't change, no. I'm not super-satisfied with that styling but it's not completely clear to me how to get rid of it. It looks like it may be possible by telling it to render directly to html rather than the default `svg`. I'll give it a go.

Not been able to get it to successfully render any other way.

> > It would be good to add some cross references to these pages from the deployment guide and management guide.
> 
> 👍

Done!

JimMadge (changes requested):
The command blocks don't show in the build preview now.

Does it work locally?

craddm:
> The command blocks don't show in the build preview now.
> 
> Does it work locally?

strange -  yes, it does!

JimMadge:
> > The command blocks don't show in the build preview now.
> > Does it work locally?
> 
> strange - yes, it does!

Any clues in the ReadTheDocs output?

Might be worth trying in a completely fresh venv, it might be that things lingering from before make it work?

craddm:
> > > The command blocks don't show in the build preview now.
> > > Does it work locally?
> > 
> > 
> > strange - yes, it does!
> 
> Any clues in the ReadTheDocs output?
> 
> Might be worth trying in a completely fresh venv, it might be that things lingering from before make it work?

ReadTheDocs is not finding the module:

```
/home/docs/checkouts/readthedocs.org/user_builds/data-safe-haven/checkouts/2238/docs/source/reference/config.md:5: CRITICAL: Failed to import "data_safe_haven.commands.config:config_command_group"The following exception was raised:
Traceback (most recent call last):
  File "/home/docs/checkouts/readthedocs.org/user_builds/data-safe-haven/envs/2238/lib/python3.12/site-packages/sphinxcontrib/typer/__init__.py", line 309, in import_object
    obj = import_module(try_path)
```

Will try to tweak it, but might need to put the path back in `conf.py`.

Docs now not building locally or via actions due to package conflicts

JimMadge:
@craddm Maybe you need to run `pip-compile` again? Merging develop may have made the two envs incompatible.

docs/source/deployment/deploy_shm.md
```diff
@@ -54,7 +54,7 @@ Before deploying the Safe Haven Management (SHM) infrastructure you need to deci
   Use the short name without spaces, _e.g._ **uksouth** not **UK South**
   :::
 
-Once you've decided on these, run the following command: [approx 5 minutes]:
+Once you've decided on these, run the {typer}`dsh-shm-deploy` command: [approx 5 minutes]:
```
jemrobinson, line 57: If this is supposed to produce a link configured by `sphinxcontrib.typer` it isn't currently working for me (in the online preview).

craddm: works locally, that's probably because the docs aren't generating correctly on ReadTheDocs. The module itself isn't installed on the ReadTheDocs runner so it can't just `import data_safe_haven`

docs/source/conf.py
```diff
@@ -5,7 +5,10 @@
 # https://www.sphinx-doc.org/en/master/usage/configuration.html
 import datetime
 import emoji
-
+from pathlib import Path
```
jemrobinson: Instead of trying to hack `conf.py` you can change the `build` command that RTD uses in `.readthedocs.yaml` (https://docs.readthedocs.io/en/stable/config-file/v2.html#build-commands)

docs/source/deployment/deploy_shm.md
JimMadge: I suppose that is because it isn't part of the `requirements.txt` file.

[extra requirements](https://docs.readthedocs.io/en/stable/config-file/v2.html#python) should fix this I think.

docs/source/conf.py
jemrobinson: Actually, there's an example of how to do exactly the thing we want to do in the `python` section:

```python
python:
  install:
    - requirements: docs/requirements.txt
    # Install our python package before building the docs
    - method: pip
      path: .
```

jemrobinson: N.B. I did previously try using

```
build:
  command: hatch build
```

but this skips some of the extras that RTD do for you (e.g. building PDF versions of the docs).

craddm: > Actually, there's an example of how to do exactly the thing we want to do in the `python` section:
> 
> ```python
> python:
>   install:
>     - requirements: docs/requirements.txt
>     # Install our python package before building the docs
>     - method: pip
>       path: .
> ```

That did the job, working now!

JimMadge (approved):
Great, looks good

PR #2240, opened 2024-10-17
Fix Pulumi/dsh Python mismatch

### LINKED ISSUE #2239: Pulumi using wrong Python/Python path in pipx install

- **Operating System:** 
- **Data Safe Haven version:** 

## :package: Packages

<details>

```none
Paste list of packages here
```

</details>

After pipx install, when calling `dsh` Pulumi will look for a Python interpreter and find system Python, not the Python venv with all of the dependencies. It will fail because of missing packages.

On option to fix,
[can declare venv path](https://www.pulumi.com/docs/iac/languages-sdks/python/#pypi-packages) in Pulumi package settings.
We set these [here](https://github.com/alan-turing-institute/data-safe-haven/blob/53058974451dc0febb535d594e4849b74067f4b2/data_safe_haven/infrastructure/project_manager.py#L86)

<details>

```none
Your log details here
```

</details>

jemrobinson:

n/a

Set PULUMI_PYTHON_CMD to use the same Python executable that is being used by dsh. 

Thanks for the pointer @craddm !

Closes #2239 

### :microscope: Tests

Tested locally, but we should test on the rc1 branch once this is merged into it

### DIFF
--- a/data_safe_haven/external/interface/pulumi_account.py
+++ b/data_safe_haven/external/interface/pulumi_account.py
@@ -1,5 +1,6 @@
 """Manage Pulumi accounts"""
 
+import sys
 from shutil import which
 from typing import Any
 
@@ -39,5 +40,6 @@ class PulumiAccount:
                 "AZURE_STORAGE_ACCOUNT": self.storage_account_name,
                 "AZURE_STORAGE_KEY": str(storage_account_keys[0].value),
                 "AZURE_KEYVAULT_AUTH_VIA_CLI": "true",
+                "PULUMI_PYTHON_CMD": sys.executable,
             }
         return self._env


JimMadge (commented):
I'm not sure this will fix #2239 on its own. I don't think it is just the Python interpreter but also the `PYTHONPATH`.

craddm:
Seems to work on my Docker container

jemrobinson:
Yeah, it also worked for me. I'm not sure why, but I assume it sets the path correctly behind the scenes?

JimMadge:
Interesting, maybe the interpreter "knows" where it's modules are, or the default path is relative to the executable file or something. If this works on its own though I don't see much benefit in also finding and setting the venv path :+1:.

JimMadge (approved):

jemrobinson:
I tried the venv method and couldn't get it to work, so if this fix works for @helendduncan then I'm happy with it :)

JimMadge:
> Seems to work on my Docker container

Reading this a second time, does that mean this works in an environment where it _didn't_ work before (a `pipx` install), or it still works in your container where it worked previously?

craddm:
> > Seems to work on my Docker container
> 
> Reading this a second time, does that mean this works in an environment where it _didn't_ work before (a `pipx` install), or it still works in your container where it worked previously?

`pipx` didn't work previously - same issue with not finding the modules. Note this is standalone Docker container, not the `.devcontainer` that I usually use.

JimMadge:
Great, that sounds like a positive result for this change working then :tada:.

jemrobinson:
Confirmed working by @helendduncan

PR #2241, opened 2024-10-18
Merge development updates into 5.0.1rc1 release branch

JimMadge:


JimMadge:
I would guess it is the psycopg downgrade that is causing a conflict here.

jemrobinson:
Are you sure this is a good idea? We need to make sure that rc1 is tested and produces a deployment that actually works. The current branch I think meets these requirements.

JimMadge:
> Are you sure this is a good idea? We need to make sure that rc1 is tested and produces a deployment that actually works. The current branch I think meets these requirements.

Probably 😅. I don't think it is really a rc, it is a pointer to where a team are currently testing. The milestone isn't finished yet and work towards that is still going on on `develop`.

craddm:
I agree with Jim; this rc1 felt more like laying down a marker for where we were at when TRESA were starting testing rather than a genuine release candidate

JimMadge:
Closing due to merge conflicts on develop, I'll do the merge on a new branch and open a PR from that.

PR #2246, opened 2024-10-21
Merge develop changes in 5.0.1rc1

JimMadge:


.hatch/requirements-test.txt
```diff
@@ -295,18 +295,18 @@ psycopg-binary==3.1.19
...
     #   pulumi-azure-native
     #   pulumi-azuread
     #   pulumi-random
-pulumi-azure-native==2.64.3
+pulumi-azure-native==2.66.0
     # via
     #   -c .hatch/requirements.txt
     #   hatch.envs.test
-pulumi-azuread==5.53.4
+pulumi-azuread==6.0.0
```
jemrobinson, line 309: Do we know whether this is a breaking change?

data_safe_haven/config/config_sections.py
```diff
@@ -67,7 +68,7 @@ class ConfigSectionSRE(BaseModel, validate_assignment=True):
     @field_validator(
         "admin_ip_addresses",
         "data_provider_ip_addresses",
-        "research_user_ip_addresses",
+        # "research_user_ip_addresses",
```
jemrobinson, line 71: Why is this validated separately? If it's because it's allowed to include "Internet", then I think this is actually something we should allow for all three.

data_safe_haven/config/sre_config.py
```diff
@@ -98,7 +98,10 @@ def template(cls: type[Self], tier: int | None = None) -> SREConfig:
                     allow_copy=remote_desktop_allow_copy,
                     allow_paste=remote_desktop_allow_paste,
                 ),
-                research_user_ip_addresses=["List of IP addresses belonging to users"],
+                research_user_ip_addresses=[
+                    "List of IP addresses belonging to users",
+                    "You can also use the tag 'Internet' instead of a list",
```
jemrobinson, line 103:
```suggestion
                    "You can also use an Azure ServiceTag (such as 'Internet') instead of a list",
```

jemrobinson: Also - should this be a `list[str | AzureServiceTag]` instead of `list[str] | AzureServiceTag`?

data_safe_haven/validators/validators.py
```diff
@@ -124,7 +124,7 @@ def ip_address(ip_address: str) -> str:
     try:
         return str(ipaddress.ip_network(ip_address))
     except Exception as exc:
-        msg = "Expected valid IPv4 address, for example '1.1.1.1'."
+        msg = "Expected valid IPv4 address, for example '1.1.1.1', or 'Internet'."
```
jemrobinson, line 127:
```suggestion
        msg = "Expected valid IPv4 address, for example '1.1.1.1', or an Azure ServiceTag, for example 'Internet'."
```

docs/source/deployment/deploy_sre.md
```diff
@@ -69,6 +69,48 @@ sre:
...
 
+:::{admonition} Supported Azure regions
+:class: dropdown important
+
+Some of the SRE resources are not available in all Azure regions.
+
+- Workspace virtual machines use zone redundant storage managed disks which have [limited regional availability](https://learn.microsoft.com/en-us/azure/virtual-machines/disks-redundancy).
+- Some shares mounted on workspace virtual machines require premium file shares which have [limited regional availability](https://learn.microsoft.com/en-us/azure/storage/files/redundancy-premium-file-shares).
+
+The regions which satisfy all requirements are,
```
jemrobinson, line 80: If we're prepared to be this definitive here, are we also prepared to enforce this in pydantic validation?

jemrobinson (changes requested):
I realise that all of these PRs have been individually reviewed, but some things struck me as possibly worth revisiting before we merge into the RC branch.

.hatch/requirements-test.txt
JimMadge: Yep, I checked https://github.com/alan-turing-institute/data-safe-haven/pull/2235

data_safe_haven/config/config_sections.py
JimMadge: Yes, it was deliberately only the research users as that seemed to be the use case.

Exposing the admin filter to the Internet feels much riskier.
I would be inclined to wait and see if it is a feature users want.

data_safe_haven/config/sre_config.py
JimMadge: That isn't right, only the Internet tag can be used.

In that case, it seemed sensible for the type to be tag or list of IPs rather than list of IPs and tags.

data_safe_haven/validators/validators.py
JimMadge: As above, only Internet.

data_safe_haven/config/config_sections.py
jemrobinson: Agreed that admins might be worth holding off. Data providers is definitely needed - it's really common that they don't have a fixed IP address (or don't know what their assigned range is if they do).

docs/source/deployment/deploy_sre.md
JimMadge: Could do.

That list could change, so you could also argue it is better not to be too prescriptive in the code so you don't get a "X is not valid" message when it really is.

data_safe_haven/config/config_sections.py
JimMadge: `curl https://ipinfo.io` 🙄.

OK, let's make that change. Shall we do that as a new PR into the rc branch? I think it will be clearer where it comes from rather than hiding it in this merge branch that way.

jemrobinson: Yes, I think that's a good idea

data_safe_haven/validators/validators.py
jemrobinson: OK, but the current wording implies that 'Internet' is a valid IPv4 address. Can you find a better phrasing?

JimMadge: Hmm, good point. I _think_ it is correct with the commas, but it is complicated.

Maybe two sentences, "Expected either a valid IPv4 address or 'Internet'. IPv4 addresses should be formatted like '1.1.1.1'.".

JimMadge: Actually, I think this is a mistake! This validator _only_ accepts IP addresses.

data_safe_haven/config/config_sections.py
JimMadge: It is in #2247

data_safe_haven/validators/validators.py
JimMadge: Also fixed in #2247

jemrobinson (approved):
LGTM

PR #2247, opened 2024-10-21
Allow 'Internet' for data providers IP

JimMadge:

#2246 

Allows data providers (as well as research users) IP addresses to be defined as the Azure service tag 'Internet'.

### :microscope: Tests

Tested on a new deployment. Ingress and Egress blobs anonymous access and with SAS token. Checked download from egress is possible with appropriate SAS token. Checked upload to ingress is possible with appropriate SAS token.

With the storage account set to allow all network connections and blob public access disabled (which is default, but also explicitly set here),

With the storage account URL, you cannot anonymously access the containers

Screenshot 2024-10-31 at 15 35 07

With a valid SAS token you can access the containers and manipulate the data according to the permissions the SAS allows

Screenshot 2024-10-31 at 15 36 09


JimMadge:
This might be a bit more tricky as storage accounts use [IP address ranges](https://learn.microsoft.com/en-gb/azure/storage/common/storage-network-security?tabs=azure-portal#grant-access-from-an-internet-ip-range) and not service tags for restricting external access.

It is possible to allow connection from any network.

craddm (approved):
LGTM

jemrobinson (changes requested):
Should this be targeting `develop` instead of `latest`?

JimMadge:
> Should this be targeting `develop` instead of `latest`?

I did have it as a hotfix. However, happy to move it to the next minor release as that should be soon, and this does change networking somewhat.

jemrobinson (approved):

PR #2248, opened 2024-10-22
Use Pulumi to create Entra applications

### LINKED ISSUE #2215: Create Entra applications with Pulumi

Create Entra applications with Pulumi instead of SREProvisioningManager.

Use the `pulumi-azuread` Provider.

jemrobinson:

n/a

Replace creation of SRE Entra applications through the SDK with creation through `pulumi-azuread`.

**Note** we still need one SDK-created Entra application in the SHM which is used to authenticate `pulumi-azuread`.

The `msgraph_permissions` map looks up the GUIDs for all possible permissions. We could consider replacing this with a static lookup table that only covers the ones we need.

Closes #2215 

### :microscope: Tests

Tested on a fresh SRE deployment


data_safe_haven/infrastructure/components/composite/entra_application.py
```diff
@@ -0,0 +1,158 @@
...
+                            )
+                            for permission_type, permission in props.application_permissions
+                        ],
+                        resource_app_id=props.msgraph_client_id,
+                    )
+                ]
+                if props.application_permissions
+                else []
+            ),
+            sign_in_audience="AzureADMyOrg",
```
JimMadge: Is there a object in `pulumi-azuread` for this value? Is it worth making one ourselves (I think we only use it once so not a huge deal)?

data_safe_haven/infrastructure/components/composite/entra_application.py
```diff
@@ -0,0 +1,158 @@
...
+            if permission_type == EntraAppPermissionType.DELEGATED:
+                entra.ServicePrincipalDelegatedPermissionGrant(
+                    replace_separators(
+                        f"{self._name}_application_delegated_grant_{permission_type.value}_{permission}",
+                        "_",
+                    ).lower(),
+                    claim_values=[permission],
+                    resource_service_principal_object_id=props.msgraph_object_id,
+                    service_principal_object_id=self.application_service_principal.object_id,
+                )
```
JimMadge: Might be clearer if we can do this as a pair of list comprehensions. Something like

```Python
[entra.AppRoleAssignment(...permission...) for permission_type, permission in ... if permission_type is EntraAppPermissionType.APPLICATION]
[entra.ServicePrincipalDelegatedPermissionGrant(...permission...) for permission_type, permission in ... if permission_type is EntraAppPermissionType.DELEGATED]
```

data_safe_haven/infrastructure/programs/imperative_shm.py
```diff
@@ -147,7 +147,12 @@ def deploy(self) -> None:
         try:
             graph_api.create_application(
                 self.context.entra_application_name,
-                application_scopes=["Group.ReadWrite.All"],
+                application_scopes=[
+                    "Application.ReadWrite.All",
+                    "AppRoleAssignment.ReadWrite.All",
+                    "Directory.ReadWrite.All",
+                    "Group.ReadWrite.All",
+                ],
```
JimMadge: These new scopes are required for the application to do the new azuread stuff?

JimMadge (commented):
Looks good I think moving more of this stuff to community supported Pulumi modules will help a lot.

Just a few style questions.

data_safe_haven/infrastructure/components/composite/entra_application.py
jemrobinson: There's not, but I've added an Enum for this in b0dbcf685.

data_safe_haven/infrastructure/components/composite/entra_application.py
jemrobinson: Maybe? It means you need to iterate through the list twice, and I'm not sure whether it's really clearer or not. Happy to do it if you think it's worth it though.

data_safe_haven/infrastructure/programs/imperative_shm.py
jemrobinson: In-line explanatory comments added in bc3db5a52.

jemrobinson:
@JimMadge : My biggest worry here is that we add an entry for the Microsoft Graph Service Principal to our state file. This contains a huge amount of data (which more than doubles the size of the state file). Since we're only using it to get the GUIDs for various named permissions, we *could* replace the bits we need with an Enum or lookup-table. However, this would mean that we would be out-of-sync if any of these GUIDs changed (unlikely to happen as this would break a *lot* of third-party code).

What do you think?

data_safe_haven/infrastructure/components/composite/entra_application.py
JimMadge: Yeah, I doubt this is the rate limiting part of the code though.

_I_ think it is clearer. I've definitely been using list comprehensions long enough that `[something(item) for item in items if something_else(item)]` looks a lot more sensible than a for loop with conditional blocks.

JimMadge:
@jemrobinson is that the long version of the note about `msgraph_permissions` in the PR?

How large is it exactly, does it cause problems?
I'd be happy to leave it. Might actually provide some protection against Entra configuration changing.

jemrobinson:
> @jemrobinson is that the long version of the note about `msgraph_permissions` in the PR?

Yes
 
> How large is it exactly, does it cause problems? I'd be happy to leave it. Might actually provide some protection against Entra configuration changing.

About 12k lines (of which we use maybe 4-5 lines) of a 30k JSON file for a full SRE deployment. Also, we should note that the GraphApi class already has a partial copy of this (see below), so abstracting it into a common class could actually simplify our current structure.

https://github.com/alan-turing-institute/data-safe-haven/blob/1bf9d0c9afae78f67db5673c630be5ba3b780645/data_safe_haven/external/api/graph_api.py#L27-L49

JimMadge:
@jemrobinson That sounds like a good idea then, if the stack doesn't really need the full set :+1:.

I'm not super worried though because it is just a few kb in a file we don't expect people to read or edit by hand.

jemrobinson:
@JimMadge : this is actually a bigger refactor than I'd realised. Let's merge this as-is (if you're happy).

data_safe_haven/infrastructure/components/composite/entra_application.py
JimMadge: Would still like this change unless there is a problem with it.

jemrobinson: I think replacement would look like:

```python
[
    entra.AppRoleAssignment(
        replace_separators(
            f"{self._name}_application_role_grant_{permission_type.value}_{permission}",
            "_",
        ).lower(),
        app_role_id=props.msgraph_permissions[permission_type][permission],
        principal_object_id=self.application_service_principal.object_id,
        resource_object_id=props.msgraph_object_id,
    )
    for permission_type, permission in props.application_permissions
    if permission_type == EntraAppPermissionType.APPLICATION
]
[
    entra.ServicePrincipalDelegatedPermissionGrant(
        replace_separators(
            f"{self._name}_application_delegated_grant_{permission_type.value}_{permission}",
            "_",
        ).lower(),
        claim_values=[permission],
        resource_service_principal_object_id=props.msgraph_object_id,
        service_principal_object_id=self.application_service_principal.object_id,
    )
    for permission_type, permission in props.application_permissions
    if permission_type == EntraAppPermissionType.DELEGATED
]
```

Is this what you were thinking of? I'm not convinced that it's better than what we currently have.

JimMadge: Yep, I prefer that.

JimMadge (approved):

PR #2253, opened 2024-10-23
Improve DNS delegation feedback

### LINKED ISSUE #2228: Better explanation of domain setup

- [x] I have searched open and closed issues for duplicates
- [x] This is a request for a change to documentation
- [x] This isn't a feature request or bug report

The instructions about how to register a domain are very opaque. They also conflate the idea of "I own this domain" with "Entra is allowed to register users with this in their user name".

We should change these to something more user-friendly, e.g. 

```
- "Domain 'mytre.example.org' is not registered
- You need to register 'mytre' with 'example.org'
- Create a new NS record for 'example.org'
- Add the following records ...
```

JimMadge:

Closes #2228

### :microscope: Tests

Tested locally

```console
Ensured that DNS TXT record @ exists in zone daimyo.develop.turingsafehaven.ac.uk.
daimyo.develop.turingsafehaven.ac.uk DNS is not delegated correctly.
To proceed you will need to delegate daimyo.develop.turingsafehaven.ac.uk to specific
Azure nameservers
Create 4 NS records for daimyo.develop.turingsafehaven.ac.uk (for example in the zone of
develop.turingsafehaven.ac.uk)
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ domain                               ┃ record type ┃ value                  ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━┩
│ daimyo.develop.turingsafehaven.ac.uk │ NS          │ ns1-36.azure-dns.com.  │
│ daimyo.develop.turingsafehaven.ac.uk │ NS          │ ns2-36.azure-dns.net.  │
│ daimyo.develop.turingsafehaven.ac.uk │ NS          │ ns3-36.azure-dns.org.  │
│ daimyo.develop.turingsafehaven.ac.uk │ NS          │ ns4-36.azure-dns.info. │
└──────────────────────────────────────┴─────────────┴────────────────────────┘
You can learn more about NS records here:
https://www.cloudflare.com/learning/dns/dns-records/dns-ns-record/
Are you ready to check whether daimyo.develop.turingsafehaven.ac.uk has been delegated to
the correct Azure nameservers? [y/n] (y):
```

### DIFF
--- a/data_safe_haven/external/api/graph_api.py
+++ b/data_safe_haven/external/api/graph_api.py
@@ -1035,7 +1035,7 @@ class GraphApi:
                 # Check whether all expected nameservers are active
                 with suppress(resolver.NXDOMAIN):
                     self.logger.debug(
-                        f"Checking [green]{domain_name}[/] domain registration status ..."
+                        f"Checking [green]{domain_name}[/] DNS delegation."
                     )
                     active_nameservers = [
                         str(ns) for ns in iter(resolver.resolve(domain_name, "NS"))
@@ -1045,23 +1045,28 @@ class GraphApi:
                         for nameserver in expected_nameservers
                     ):
                         self.logger.info(
-                            f"Verified that [green]{domain_name}[/] is registered as a custom Entra ID domain."
+                            f"[green]{domain_name}[/] DNS has been delegated correctly."
                         )
                         break
                 self.logger.warning(
-                    f"Domain [green]{domain_name}[/] is not currently registered as a custom Entra ID domain."
+                    f"[green]{domain_name}[/] DNS is not delegated correctly."
                 )
                 # Prompt user to set domain delegation manually
-                docs_link = "https://learn.microsoft.com/en-us/azure/dns/dns-delegate-domain-azure-dns#delegate-the-domain"
                 self.logger.info(
-                    f"To proceed you will need to delegate [green]{domain_name}[/] to Azure ({docs_link})"
+                    f"To proceed you will need to delegate [green]{domain_name}[/] to specific Azure nameservers"
                 )
                 ns_list = ", ".join([f"[green]{n}[/]" for n in expected_nameservers])
                 self.logger.info(
-                    f"You will need to create NS records pointing to: {ns_list}"
+                    f"Create {len(ns_list)} [green]NS[/] records for [green]{domain_name}[/] with values: {ns_list}"
+                )
+                docs_link = (
+                    "https://www.cloudflare.com/learning/dns/dns-records/dns-ns-record/"
+                )
+                self.logger.info(
+                    f"You can learn more about NS records here: {docs_link}"
                 )
                 if not console.confirm(
-                    f"Are you ready to check whether [green]{domain_name}[/] has been delegated to Azure?",
+                    f"Are you ready to check whether [green]{domain_name}[/] has been delegated to the correct Azure nameservers?",
                     default_to_yes=True,
                 ):
                     self.logger.error("User terminated check for domain delegation.")
--- a/data_safe_haven/external/api/graph_api.py
+++ b/data_safe_haven/external/api/graph_api.py
@@ -1056,8 +1056,13 @@ class GraphApi:
                     f"To proceed you will need to delegate [green]{domain_name}[/] to specific Azure nameservers"
                 )
                 ns_list = ", ".join([f"[green]{n}[/]" for n in expected_nameservers])
+                domain_parent = ".".join(domain_name.split(".")[1:])
                 self.logger.info(
-                    f"Create {len(ns_list)} [green]NS[/] records for [green]{domain_name}[/] with values: {ns_list}"
+                    f"Create {len(ns_list)} [green]NS[/] records for [green]{domain_name}[/] (for example in the zone of {domain_parent})"
+                )
+                console.tabulate(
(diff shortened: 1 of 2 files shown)


jemrobinson (approved):
LGTM. Could @dsj976 @helendduncan @edwardchalstrey1 @edchapman88 confirm whether they find this wording clearer?

JimMadge:
Is this what you were imagining @jemrobinson?

I think it is clearer. There was a mix of delegation (which is a DNS concept) with "registration", which I think is a concept we created?

I also think the code is the wrong place to be explaining DNS, so if we need more I think it would be better to link to useful resources in the docs. Or, we could add a section to the deployment docs giving an overview of why DNS delegation is necessary and links to resources.

data_safe_haven/external/api/graph_api.py
```diff
@@ -1045,23 +1045,28 @@ def verify_custom_domain(
...
                 # Prompt user to set domain delegation manually
-                docs_link = "https://learn.microsoft.com/en-us/azure/dns/dns-delegate-domain-azure-dns#delegate-the-domain"
                 self.logger.info(
-                    f"To proceed you will need to delegate [green]{domain_name}[/] to Azure ({docs_link})"
+                    f"To proceed you will need to delegate [green]{domain_name}[/] to specific Azure nameservers"
                 )
                 ns_list = ", ".join([f"[green]{n}[/]" for n in expected_nameservers])
                 self.logger.info(
-                    f"You will need to create NS records pointing to: {ns_list}"
+                    f"Create {len(ns_list)} [green]NS[/] records for [green]{domain_name}[/] with values: {ns_list}"
```
jemrobinson: Should we be even more clear here?

`Create an NS record for {subdomain} in {parent_domain} with these nameservers: {ns_list}`

We'd have to construct `subdomain` and `parent_domain` but I think this will (always?) be `domain_name.split(".")[0]` and `".".join(domain_name.split(".")[1:])`.

JimMadge: _Maybe_.

I did think about that, but I'm not sure if it assumed more about our users DNS set up than we can.

I think what we actually require is that all requests to `sre.tre.com` resolve to the Azure nameservers. Which isn't the same as saying that the nameservers for `tre.com` hold those NS records.

jemrobinson: Sure, but the registrar for `tre.com` has to have an NS record for `sre` which is the place we need users to add these records. I think we can safely decompose `sre.tre.com` into `sre` (subdomain) and `tre.com` (parent domain) right?

JimMadge: What I'm saying is, `tre.com` doesn't need to have NS records for `sre.tre.com` _to_ Azure nameservers. It could point to some other nameservers, as long as those all eventually resolve to the Azure nameservers.

I'm not really sure _why_ you would want to do that, but it isn't really my area of expertise.

I suppose what I'm trying to figure out is,

Is the benefit of being a bit more specific ("create this record, and create it here" rather than "create this record, and I assume you know where to do that") worth the cost of saying something that might not be quite true 🤔.

Maybe this is one of those cases where if you have some more complex DNS setup, you probably know what you are doing anyway?

jemrobinson: Yeah, maybe we could have something like "you'll need to ensure {subdomain} is delegated to the nameservers: {ns_list}. If you're not sure what this means then do X"

JimMadge: Hmm, I'm less keen on that because I don't feel confident I know what they _should_ do.

Maybe,

`Create a NS records for {subdomain} (for example in the {parent_domain} zone) ...`

JimMadge:
@jemrobinson how about the current output? I've added a note about the likely DNS zone and tabulated the NS records for clarity.

jemrobinson:
It's mostly right, but there's only one NS record which has all four servers under it (e.g. like this)

Screenshot 2024-10-23 at 15 02 16

I'm not a DNS expert but I think whatever you're using for SHM FQDN, you'll still need to have an NS record in `parent-domain`. In other words, if you want to use `tre.com` as your SHM FQDN and have this hosted on Azure, you'll need to (get someone to) set an NS record for `tre` in `com`.

JimMadge:
> It's mostly right, but there's only one NS record which has all four servers under it (e.g. like this)

That's just how the portal (incorrectly 😠) displays it. There are four records, one for each nameserver

```console
❯ nslookup -type=ns daimyo.develop.turingsafehaven.ac.uk
Server:         192.168.178.202
Address:        192.168.178.202#53

Non-authoritative answer:
daimyo.develop.turingsafehaven.ac.uk    nameserver = ns1-36.azure-dns.com.
daimyo.develop.turingsafehaven.ac.uk    nameserver = ns2-36.azure-dns.net.
daimyo.develop.turingsafehaven.ac.uk    nameserver = ns3-36.azure-dns.org.
daimyo.develop.turingsafehaven.ac.uk    nameserver = ns4-36.azure-dns.info.

Authoritative answers can be found from:

```

PR #2266, opened 2024-10-28
Add confirmation checks and check for deployed SREs before teardown operations

### LINKED ISSUE #2252: Enhance the teardown documentation for clarity

- [x] I have searched open and closed issues for duplicates
- [x] This is a request for a change to documentation
- [x] This isn't a feature request or bug report

https://data-safe-haven.readthedocs.io/en/develop/management/index.html#removing-a-deployed-data-safe-haven

## :memo: Required documentation changes

- Explain a bit more exactly what will happen when running `dsh sre teardown YOUR_SRE_NAME` or `dsh shm teardown` so System Managers new to DSH don't delete things without meaning
- Add some big scary warnings, especially to the SHM teardown part saying "deleting this will affect all SREs" etc
- I'm not sure if in DSH v5 the storage account is now falling under the SRE resources as oppose to the SHM subscription (as was the case in v4)? If so, also add big scary warning saying data will be deleted by SRE teardown (unless this is not the case, in which case explain that)

craddm:

Prevents SHM teardown if an SRE linked to the current SHM is deployed, and adds a confirmation step before teardown
image

Adds a confirmation step before SRE teardown
image

Closes #2252 

### :microscope: Tests

Tested locally


data_safe_haven/commands/shm.py
```diff
@@ -142,6 +142,14 @@ def teardown() -> None:
     try:
         config = SHMConfig.from_remote(context)
         shm_infra = ImperativeSHM(context, config)
+        console.print(
+            "Tearing down the Safe Haven Management environment will permanently delete all associated resources, including remotely stored configurations."
+        )
+        if not console.confirm(
+            "Do you wish to continue tearing down the SHM?", default_to_yes=True
```
JimMadge:
```suggestion
            "Do you wish to continue tearing down the SHM?", default_to_yes=False
```
I think it would make sense to default to no when we are adding a protection like this.

data_safe_haven/commands/shm.py
```diff
@@ -142,6 +142,14 @@ def teardown() -> None:
...
+        ):
+            logger.info("SHM teardown cancelled by user.")
```
JimMadge:
```suggestion
```
I think we shouldn't need this because the `console.confirm()` method adds to the logs.

data_safe_haven/commands/sre.py
```diff
@@ -197,6 +199,17 @@ def teardown(
         pulumi_config = DSHPulumiConfig.from_remote(context)
         sre_config = SREConfig.from_remote_by_name(context, name)
 
+        console.print(
+            "Tearing down the Secure Research Environment will permanently delete all associated resources, "
+            "including all data stored in the environment.\n"
+            "Ensure that any desired outputs have been extracted before continuing."
+        )
+        if not console.confirm(
+            "Do you wish to continue tearing down the SRE?", default_to_yes=True
```
JimMadge:
```suggestion
            "Do you wish to continue tearing down the SRE?", default_to_yes=False
```

data_safe_haven/commands/sre.py
```diff
@@ -197,6 +199,17 @@ def teardown(
...
+        ):
+            logger.info("SRE teardown cancelled by user.")
```
JimMadge:
```suggestion
```

data_safe_haven/infrastructure/programs/imperative_shm.py
```diff
@@ -172,6 +172,12 @@ def teardown(self) -> None:
             DataSafeHavenAzureError if any resources cannot be destroyed
         """
         logger = get_logger()
+        pulumi_config = DSHPulumiConfig.from_remote(self.context)
```
JimMadge: Do we need to handle the case where the pulumi config doesn't exist here?

tests/commands/test_shm.py
```diff
@@ -41,10 +43,12 @@ def test_infrastructure_auth_failure(
 class TestTeardownSHM:
     def test_teardown(
         self,
+        mocker,
         runner,
         mock_imperative_shm_teardown_then_exit,  # noqa: ARG002
         mock_shm_config_from_remote,  # noqa: ARG002
     ):
+        mocker.patch.object(Confirm, "ask", return_value="yes")
```
JimMadge: We can use the runner to [test input](https://typer.tiangolo.com/tutorial/testing/#call-pytest).

tests/commands/test_sre.py
```diff
@@ -111,6 +113,7 @@ def test_teardown(
         mock_sre_config_from_remote,  # noqa: ARG002
         mock_sre_project_manager_teardown_then_exit,  # noqa: ARG002
     ) -> None:
+        mocker.patch.object(Confirm, "ask", return_value="yes")
```
JimMadge: We can use the runner to [test input](https://typer.tiangolo.com/tutorial/testing/#call-pytest).

JimMadge (changes requested):

data_safe_haven/commands/shm.py
craddm: I think it's still useful to confirm it for the user, so changed these to console.print

data_safe_haven/infrastructure/programs/imperative_shm.py
craddm: ~~Yes, although maybe in a different PR? At the moment you can't teardown an SHM if no SREs have ever been deployed, since there is no pulumi config~~ I was convinced that was already a problem on develop but doesn't look like it, my mistake. I'll catch it here.

JimMadge: But that means this change introduces a bug right? I think we should fix it here.

JimMadge (changes requested):
I think we need to be careful to not knowingly introduce a bug here. Other than that it looks great.

data_safe_haven/infrastructure/programs/imperative_shm.py
craddm: Yes, see updated comment!

JimMadge (approved):
Great :+1:

JimMadge (approved):
Even better! 🚀 😆

PR #2284, opened 2024-11-08
Change method of sanitising SRE names

### LINKED ISSUE #2278: `dsh config available` is not showing deployed SREs

- **Operating System:** macOS
- **Data Safe Haven version:** `Data Safe Haven 5.0.1`

## :package: Packages

<details>

```none
Paste list of packages here
```

</details>

The SRE with ID `testt2hdl` is live and accessible over `https://testt2hdl.prod5.turingsafehaven.ac.uk/`. However, when listing deployed SREs using `dsh config available` we see the following:

```
┏━━━━━━━━━━━━┳━━━━━━━━━━┓
┃ SRE Name   ┃ Deployed ┃
┡━━━━━━━━━━━━╇━━━━━━━━━━┩
│ cgcdemosre │          │
│ edtest2p5  │          │
│ edtestp5   │          │
│ sbox123    │          │
│ testsre    │          │
│ testt2hdl  │          │
└────────────┴──────────┘
```

Notice no SRE is marked as deployed.

<details>

```none
Your log details here
```

</details>

## :recycle: To reproduce

Just run `dsh config available` from any account with production access. @dsj976 and @helendduncan were able to reproduce.

craddm:

Drops the `json_safe` function, which is only used to generate SRE filenames without non-alphanumeric characters. The SRE names from the config files are now used, with those names already being sanitised as `SafeStrings` containing only alphanumeric characters, hyphens, and underscores, all of which can be included in filenames.

This prevents issues with mismatches between the config filename and the SRE name in the config.

Closes #2278 

### :microscope: Tests

Tested locally


jemrobinson (approved):
LGTM

JimMadge:
Can this be a patch release, or will it break existing deployments?

craddm:
I guess it might break them, since the expected config names will now be different (although this should be fixable by reuploading the config).

jemrobinson:
If you can write a workaround (download config, upgrade, upload config) then I think this can still be a point release.

data_safe_haven/config/sre_config.py
```diff
@@ -18,8 +17,8 @@
 
 
 def sre_config_name(sre_name: str) -> str:
-    """Construct a safe YAML filename given an input SRE name."""
-    return f"sre-{json_safe(sre_name)}.yaml"
+    """Construct a YAML filename given an input SRE name."""
+    return f"sre-{sre_name.lower()}.yaml"
```
JimMadge: Are we worried this will introduce the same kind of problem?

We are handling `sre_name` differently across the code?

JimMadge (changes requested):
Looks good, just want to be careful we aren't reintroducing the same or a similar problem here.

data_safe_haven/config/sre_config.py
JimMadge: _i.e._, should the validator insist lowercase, or should we allow uppercase in the file names?

jemrobinson: +1 - let the validator do this work.

craddm: Added a new validator, as SafeString is used for docker PATs (among other things) and thus needs to accept upper case

JimMadge (approved):
Looks good.

I pushed a commit to swap the `*` in regexes with `+` so that blank strings are not allowed.
Not sure if it would cause errors but I don't think we expect to allow that.
We could also modify the annotated types to give them a minimum length.

PR #2292, opened 2024-11-14
Check that a user belongs to the correct SHM domain when registering with an SRE

### LINKED ISSUE #2275: Existing users domains not matching when added

- [x] This is a problem observed when **managing** a Data Safe Haven.

- **Operating System:** macOS
- **Data Safe Haven version:** 5.0.1

## :package: Packages

<details>

```none
Paste list of packages here
```

</details>

A test SRE was sucsessfully deployed and two users were added to it via the [CLI](https://data-safe-haven.readthedocs.io/en/latest/management/#add-users-to-the-data-safe-haven). 
One user (Bob Smith) had a bob.smith@prod**4**... Entra user which was *updated*. The second user (James Allen) did not, and their Entra user was created.

Both Bob and James were added to the test SRE users group - however Bob, with the prod4 account, couldn't see any connections the guacamole interface.

This is because the domain doesn't match the expected domain.

Manually editing Bob's User principal name Domain via [Entra](entra.microsoft.com) allowed access, but would need to be manually edited for all users who had accounts from previous iterations of DSH

craddm:
…with SRE

Adds a check of the user's principal name when registering with an SRE. If the domain in the principal name does not match the FQDN of the SHM that the SRE is linked to, the user will not be added to the SRE's security group and an error will be printed explaining why.

Example output:
image

Relates to #2275

Does not close #2275, as that also encompasses the `add` command, whereas this PR only handles registering users

### :microscope: Tests

Tested locally


data_safe_haven/commands/users.py
```diff
@@ -146,15 +146,27 @@ def register(
 
         # List users
         users = UserHandler(context, graph_api)
-        available_usernames = users.get_usernames_entra_id()
+        available_users = users.entra_users.list()
```
JimMadge, line 150: I think the name `available_users` is a bit confusing here. Especially when we are checking if the username is already taken.

data_safe_haven/commands/users.py
```diff
@@ -146,15 +146,27 @@ def register(
...
+        user_dict = {
+            user.preferred_username.split("@")[0]: user.preferred_username.split("@")[1]
+            for user in available_users
+        }
         usernames_to_register = []
         for username in usernames:
-            if username in available_usernames:
-                usernames_to_register.append(username)
+            if username in user_dict.keys():
```
JimMadge, line 156:
```suggestion
            if username in user_dict:
```

This should be equivalent.

data_safe_haven/commands/users.py
JimMadge: Actually, no, this is the register command, right?

data_safe_haven/commands/users.py
```diff
@@ -146,15 +146,27 @@ def register(
...
+                user_domain = user_dict[username]
```
JimMadge:
```suggestion
            if user_domain:=user_dict.get(username)
```

data_safe_haven/commands/users.py
```diff
@@ -146,15 +146,27 @@ def register(
...
+                if shm_config.shm.fqdn not in user_domain:
+                    logger.error(
+                        f"User [green]'{username}'[/green]'s principal domain name is [blue]'{user_domain}'[/blue].\n"
+                        f"SRE [yellow]'{sre}'[/yellow] belongs to SHM domain [blue]'{shm_config.shm.fqdn}'[/blue].\n"
+                        "The user's principal domain name must match the domain of the SRE to be registered."
+                    )
```
JimMadge: I think we've had problems with multi-line log messages before (although we might strip those characters). Could you check the output?

JimMadge (changes requested):
Looks good, a suggestion and a couple of questions :+1:.

data_safe_haven/commands/users.py
craddm: Looks like this in the log file:
image

Seems fine to me

data_safe_haven/commands/users.py
craddm: This is the register command, yes. I chose `available_users` because the output is a list of `ResearchUsers` - not just usernames, but the full records. 
We need the user principal name here, as we need to see what domain they're attached to.

`get_usernames_entra_id()` just gets the first part of the user principal name.

It's possible we may want a `get_user_principal_names` function, as we may want to do the same thing for the `add` function - but I'll look at doing that when it comes to working on that.

data_safe_haven/commands/users.py
JimMadge: The problem there is we have lines in the log file which are not in the log format which will make it difficult to parse.

We could,

- Join the lines in the log file
- Make each of these a different message
- Put this output to stdout but a shorter summary to the log

data_safe_haven/commands/users.py
```diff
@@ -146,15 +146,26 @@ def register(
...
+                else:
+                    usernames_to_register.append(username)
             else:
                 logger.error(
-                    f"Username '{username}' does not belong to this Data Safe Haven deployment."
-                    " Please use 'dsh users add' to create it."
+                    f"Username '{username}' does not belong to this Data Safe Haven deployment.\n"
+                    "Please use 'dsh users add' to create this user."
```
JimMadge: Also here

JimMadge (approved):
LGTM :+1:

Issue #1922, opened 2024-05-30
Better logging


jemrobinson:
## :white_check_mark: Checklist

- [x] I have searched open and closed issues for duplicates.
- [x] This is a request for a new feature in the Data Safe Haven or an upgrade to an existing feature.
- [x] The feature is still missing in the [latest version](https://github.com/alan-turing-institute/data-safe-haven/releases).
- [x] I have read through the [documentation](https://alan-turing-institute.github.io/data-safe-haven/).
- [x] This isn't an open-ended question (open a [discussion](https://github.com/alan-turing-institute/data-safe-haven/discussions) if it is).

## :strawberry: Suggested change

We should ensure that our logging is improved by:

- logging more things
- alerting when things are not as expected

## :steam_locomotive: How could this be done?

Could consolidate in LogAnalyticsWorkspace and send to e.g. GreyLog? Not sure on the best way to alert

## Task list

- [x] #1931 
- [x] #2295
- [x] #2307
- [x] #2309
- [x] #2318

JimMadge:
Related to #1931

JimMadge:
[Azure monitor data source](https://learn.microsoft.com/en-us/azure/azure-monitor/data-sources)

Log sources are:

- Azure resources (most azure resources, I think this is basic data like SKU, state, ...)
- Entra ID (audit data, authentication events _etc._)
- App insights (data from app services like webapps)
- Virtual machines (_e.g._ syslog)
- Kubernetes Clusters
- Custom sources, using the REST API

I think our working assumption was that using the Azure logging resources would mean we are easily able to ingest logs from Azure resources, whereas using an alternative log manager would involve lots of work to ingest logs.

The number of things which Azure monitor can ingest natively is fairly limited though so there might not be such a big advantage to using it.

craddm:
Additional logs to investigate:

- Network events
  - Firewall logs
  - NSG logs
- Service logs
  - Containers
    - apricot
    - guac user sync
    - Nexus allowlist
    - Gitea
    - Hedgedoc
- Event logs
  - Data ingress
  - Data egress
- Authentication logs
  - From VM (PAM, ldap)
  - Entra (would need to work cross tenant)

craddm:
Logging from container instances https://learn.microsoft.com/en-us/azure/container-instances/monitor-azure-container-instances

[...]

PR #2295, opened 2024-11-15
Add logging for container instances

JimMadge:

Contributes to #1922 

### :microscope: Tests

Tested in a new deployment, screenshots below.


JimMadge:
Logs are ingested into the log analytics workspace.
There are event logs from the container instance (container started, image pulled, bootloop occurring) and not the output of the container process.

Screenshot 2024-11-19 at 10 38 33

And container process logs

Screenshot 2024-11-19 at 10 43 15

.mdlstyle.rb
```diff
@@ -6,7 +6,7 @@
 exclude_rule 'MD013'
 exclude_rule 'MD024'
 rule 'MD026', :punctuation => ".,;"
-rule 'MD029', :style => :ordered
+rule 'MD029', :style => :one
```
jemrobinson, line 9: Using automatic numbering like this can cause lists with subitems to break. For example

```
1. something

:::{image} path
:alt: Caption
:align: center
:::

1. Something else
```

might show as

```
1. something
1. something else
```

Using `style: ordered` allows you to explicitly set the second bullet to `2.` and avoid this.

JimMadge: Tried this locally and the numbering is corrected in the rendered output.
May have been a `myst-parser` bug which is fixed now?

I did see some places where there were separate lists though,

```markdown
1. first list

2. second list
```

which is incorrect.

JimMadge: So, I think one is better because the numbers are always correct in the output, we don't have to manually index ordered lists.

jemrobinson (approved):
LGTM and this is definitely worth merging on its own. However, is it possible to get the container logs into the workspace too? They're more interesting than logging when containers turn on and off (which is unlikely to be very useful to many people).

JimMadge:
> LGTM and this is definitely worth merging on its own. However, is it possible to get the container logs into the workspace too? They're more interesting than logging when containers turn on and off (which is unlikely to be very useful to many people).

Yes, they are there too :rocket:

jemrobinson:
@JimMadge: does this include logs from terminated containers (which are otherwise inaccessible)?

JimMadge:
Yes, as long as the workspace gets the log it is kept there.

PR #2305, opened 2024-11-22
Update release checklist

JimMadge:


.github/ISSUE_TEMPLATE/release_checklist.md
```diff
@@ -44,16 +44,16 @@ Refer to the [Deployment](https://data-safe-haven.readthedocs.io/en/latest/deplo
...
 
-## :computer: Release information
+### Making the release
 
-- **Version number:** _
-- **SHM ID:** _
-- **T2 SRE ID:** _
-- **T3 SRE ID:** _
+- [ ] Merge release branch into `latest`
+- [ ] Push tag in the format `v0.0.1` to the merge commit into `latest`
```
craddm: Don't quite follow this phrasing. 

Maybe something like

```suggestion
- [ ] Publish the release with `latest` as the target branch, tagging it with the format `v0.0.1`
```

craddm (changes requested):

.github/ISSUE_TEMPLATE/release_checklist.md
JimMadge: Actually, I think we need to do both of these things.

What I meant was create a tag (on the merge commit) `git tag -sm 'v0.0.1'` and push it to the remote `git push origin v0.0.1`.
Releases are a GitHub thing so you also need to make that an associate it with the tag you have pushed.

jemrobinson: You can create the tag via the GitHub interface that you use to create the release though, you don't need to use the command line for this unless you want to. I think the instructions that cover both scenarios are something like:

```
- Merge the release branch into `latest`
- Create a tag of the form `v0.0.1` pointing to the most recent commit on `latest`
- Create a new GitHub release using this tag
```

craddm: Not sure how I ended up highlighting both lines there - it was only line 50 I had a problem with, as I didn't really understand it.

I use the GitHub interface to do releases, which is probably why I found it hard to parse, and which is also what my suggestion reflects.

James's more agnostic suggestion is better.

JimMadge: Agreed. What I wrote expected that you have a tag on your local repo which you need to push.

Didn't know you could create them in the browser :tada:.

JimMadge: Done in 56081d8

> - Create a new GitHub release using this tag

I changed the wording of that because we should have already created a draft release, so to say publish your draft to be clear that you don't need to create a brand new release at this point.

craddm (approved):
LGTM

PR #2327, opened 2024-12-03
Modifying Firewall rules to provide Internet Access to T0/T1

### LINKED ISSUE #2283: Active support for T0/T1

The current SREs cannot support T0/T1 RE internet access

cptanalatriste:

The proposed approach adds an extra field to the SRE config file (`allow_workspace_internet` ) and based on its value does the following: 1) If `false` , business as usual, 2) if `true` , we remove all the firewall's application rules, we add a network rule allowing connections to the internet, and remove  `user_rules`  DNS server configuration. 

(@JimMadge After my changes)

- `allow_workspace_internet` configuration value in the `sre` section (default false)
- When allowed, compared to the default rules,
  - A network rule is added to allow outbound internet access from workspaces
  - The application rules to allow _specific_ outbound access (snap, apt) are removed
- Add new variable to templates and documentation
- Add warnings about possible data leakage when `allow_workspace_internet` is true.

Closes #2283 

### :microscope: Tests

(@JimMadge After my changes)

Outbound internet access from workspace when `allow_workspace_internet` is true

Screenshot 2025-01-10 at 14 40 17

No outbound internet access from workspace when `allow_workspace_internet` is false


data_safe_haven/infrastructure/programs/sre/firewall.py
```diff
@@ -109,205 +111,9 @@ def __init__(
...
-                                    protocol_type=network.AzureFirewallApplicationRuleProtocolType.HTTPS,
-                                ),
-                            ],
-                            source_addresses=props.subnet_workspaces_prefixes,
-                            target_fqdns=ForbiddenDomains.UBUNTU_SNAPCRAFT,
-                        ),
-                    ],
-                ),
-            ],
+            application_rule_collections=self._get_application_rule_collections(props),
```
jemrobinson: I'm not sure we gain much by moving these definitions into a function.

tests/infrastructure/programs/sre/test_firewall.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def call(self, _: pulumi.runtime.MockCallArgs):
+        return {}
+
+
+# TODO: These breaks many other tests!
+pulumi.runtime.set_mocks(
+    MyMocks(),
+    preview=False,
+)
```
jemrobinson: Luckily, you don't need these :). See the `test_application_gateway.py` for an example of Pulumi runtime tests without `set_mocks`.

tests/infrastructure/programs/sre/test_firewall.py
```diff
@@ -0,0 +1,175 @@
...
+            props=firewall_props,
+            tags=tags,
+        )
+
+    return set_allow_workspace_internet
+
+
+class InternetAccess(Enum):
+    ENABLED = True
+    DISABLED = False
```
jemrobinson: If we need this enum, it's probably useful outside just testing. Should it live in `types/enums.py`?

data_safe_haven/infrastructure/programs/sre/firewall.py
```diff
@@ -108,206 +111,236 @@ def __init__(
...
+                ],
+            ),
+        ]
+
+        network_rule_collections: list[
+            network.AzureFirewallNetworkRuleCollectionArgs
+        ] = []
+
+        if props.allow_workspace_internet:
+            application_rule_collections = []
```
craddm: Here you remove all the application rules. I think that the problem is, the way the network rules are subsequently defined, only traffic directly from workspaces is allowed out to the internet. This means the various containers (identity, remote desktop gateway etc) can't communicate with the internet, so you can't get to guacamole to be able to connect to the workspaces.

Not all of the containers actually need internet access, but some do.

data_safe_haven/infrastructure/programs/sre/firewall.py
```diff
@@ -108,206 +111,236 @@ def __init__(
...
-                                    protocol_type=network.AzureFirewallApplicationRuleProtocolType.HTTPS,
-                                ),
-                            ],
+                        network.AzureFirewallNetworkRuleArgs(
+                            description="Enables internet access to workspaces.",
+                            destination_addresses=["*"],
+                            destination_ports=["*"],
+                            name="allow-internet-access",
+                            protocols=[network.AzureFirewallNetworkRuleProtocol.ANY],
                             source_addresses=props.subnet_workspaces_prefixes,
```
craddm, line 263: This line defines which source addresses are allowed. At the moment it's just the subnet for workspaces (`10.0.2.0/29`), whereas the containers are all in `10.0.1.0/29`. 

I guess there are two options. 
1. Keep most of the application rules (it's just the snapcraft ones we wanted to get rid of)
2. Modify this line to allow all source addresses

craddm (dismissed):
I can deploy an SRE, but can't actually connect to anything. At the moment, literally *only* traffic from workspaces is allowed. None of the container services can connect to the internet, ~~so a user can't get to the remote desktop gateway, for example~~ I misdescribed that a bit. Guacamole can't talk to the microsoft Auth servers over the internet, so it can't properly log you in.

data_safe_haven/infrastructure/programs/sre/firewall.py
craddm: I've also tried fixing this locally (with option 2 from above) but still can't browse to anything from within the SRE.

Internet access is possible: I can `ftp` to `ftp.ubuntu.com` from a workspace.

So I think we need to also look at the DNS.
We have a local container that we use for DNS that (among other things) blocks most FQDNs from resolving.

jemrobinson: Allowed DNS domains are defined in `types.enums.PermittedDomains`.

cptanalatriste: Hi @craddm , thanks for checking into this. Just to check my understanding, when you tested Option 2 did you have all the application rules enabled + the new network rule? Also, changing the DNS container means modifying `data_safe_haven/resources/dns_server/AdGuardHome.mustache.yaml`?

jemrobinson: You might need to slightly change `AdGuardHome.mustache.yaml`. Currently it forbids everything then allows everything from a specified allowlist. We might need to make the first part configurable as otherwise the rules will be "deny everything" and "allow everything" and I have no idea what that will do!

craddm:
Have tested a fresh deployment with internet access enabled, and can confirm it works!

.github/workflows/lint_code.yaml
```diff
@@ -102,9 +102,20 @@ jobs:
...
+            test_config="tests/mustache/$filename.config.json"
+
+            if [ -e "$test_config" ]; then
+              mustache $test_config $yamlfile > $yamlfile
+            else
+              sed "s|{{\([/#]\)[^}]*}}|{{\1array}}|g" $yamlfile > expanded.tmp  # replace mustache arrays
+              sed -i "s|{{[^#/].\{1,\}}}|{{variable}}|g" expanded.tmp           # replace mustache variables
+              mustache .mustache_config.json expanded.tmp > $yamlfile           # perform mustache expansion overwriting original file
+            fi
+
```
JimMadge: Is there a better way to do this? Can we configure yamllint to accept template statements or is there a mustache validations tool?

tests/mustache/AdGuardHome.mustache.config.json
JimMadge, line 1: If we keep this I would prefer to put it somewhere else as it isn't part of the Python test suite.

data_safe_haven/infrastructure/programs/sre/dns_server.py
```diff
@@ -69,6 +72,9 @@ def __init__(
         )
 
         # Expand AdGuardHome YAML configuration
+        mustache_values: dict[str, object] = {
+            "allow_workspace_internet": props.allow_workspace_internet
+        }
```
JimMadge: This feels a bit redundant, can we put this in the `Output.all` statement below?

data_safe_haven/infrastructure/programs/sre/firewall.py
```diff
@@ -111,206 +114,238 @@ def __init__(
...
+            application_rule_collections = []
+            network_rule_collections.append(
+                network.AzureFirewallNetworkRuleCollectionArgs(
                     action=network.AzureFirewallRCActionArgs(
-                        type=network.AzureFirewallRCActionType.DENY
+                        type=network.AzureFirewallRCActionType.ALLOW
                     ),
-                    name="workspaces-deny",
-                    priority=FirewallPriorities.SRE_WORKSPACES_DENY,
+                    name="workspaces-all-allow",  # TODO: Fix other names.
```
JimMadge: Let's solve this here.

data_safe_haven/infrastructure/programs/sre/firewall.py
JimMadge: Is this a problem? You said the changes worked later @craddm.

data_safe_haven/infrastructure/programs/sre/firewall.py
```diff
@@ -111,206 +114,238 @@ def __init__(
...
+                ],
+            ),
+        ]
+
+        network_rule_collections: list[
+            network.AzureFirewallNetworkRuleCollectionArgs
+        ] = []
+
+        if props.allow_workspace_internet:
+            application_rule_collections = []
```
JimMadge: It feels strange to be to do it this way,

```python
list_a = [...]
list_b = [...]

if condition:
    list_a = []
    list_b.append(...)
```

I think it might be clearer like

```python
list_b_common = [...]

if condition:
   list_a = []
   list_b = list_b_common.append(...)
else:
   list_a = [...]
   list_b = list_b_common
```

tests/infrastructure/programs/sre/test_firewall.py
```diff
@@ -0,0 +1,179 @@
...
+                    replace_separators(
+                        f"test_sre_dns_server_private_zone_{dns_zone_name}", "_"
+                    ),
+                    location="Global",
+                    private_zone_name=f"privatelink.{dns_zone_name}",
+                    resource_group_name=resource_group_name,
+                    tags=tags,
+                )
+                for dns_zone_name in AzureDnsZoneNames.ALL
+            },  # TODO: Check if this works
```
JimMadge, line 41: Is this done?

tests/infrastructure/programs/sre/test_firewall.py
```diff
@@ -0,0 +1,179 @@
...
+        ):
+            application_rule_collections: list[dict] = args[0]
+            network_rule_collections: list[dict] = args[1]
+
+            # TODO: Be more precise in rule filtering.
+            allow_internet_collection: list[dict] = [
+                rule_collection
+                for rule_collection in network_rule_collections
+                if rule_collection["name"] == "workspaces-all-allow"
+            ]
```
JimMadge: Is this done?

data_safe_haven/infrastructure/programs/sre/firewall.py
```diff
@@ -111,206 +114,238 @@ def __init__(
...
-                        ),
+                        network.AzureFirewallNetworkRuleArgs(
+                            description="Enables internet access to workspaces.",
+                            destination_addresses=["*"],
+                            destination_ports=["*"],
+                            name="allow-internet-access",
+                            protocols=[network.AzureFirewallNetworkRuleProtocol.ANY],
+                            source_addresses=[
+                                "*"
+                            ],  # TODO: Check if we can make this more restrictive.
```
JimMadge:
```suggestion
                            source_addresses=["*"],
```

JimMadge:
I don't think I like adding `TODO` comments to code in a PR. I think they are too easy to miss and sidestep the issue/PR process. I think the two options are,

1. fix it now
2. open and issue

What does everyone else think?

.github/workflows/lint_code.yaml
jemrobinson: It feels like there are two steps here: 1) expand templates with mustache and 2) validate YAML templates. The previous approach was to ignore (1) and try to strip the mustache syntax out but it looks like we've reached the limit of what's possible there.

Can we try adding a mustache config file with dummy values to the repo which will let us run something like:

```
for yamlfile in $(find . -name "*.yml" -o -name "*.yaml"); do
  mustache-expand $yamlfile | yamllint
done
```

data_safe_haven/infrastructure/programs/sre/dns_server.py
```diff
@@ -28,12 +28,15 @@ class SREDnsServerProps:
...
         self,
+        *,
+        allow_workspace_internet: bool,
         dockerhub_credentials: DockerHubCredentials,
         location: Input[str],
         resource_group_name: Input[str],
         shm_fqdn: Input[str],
     ) -> None:
         self.admin_username = "dshadmin"
+        self.allow_workspace_internet: bool = allow_workspace_internet
```
jemrobinson: Not sure we need the `: bool` type hint since `allow_workspace_internet` is already hinted as a bool.

data_safe_haven/infrastructure/programs/sre/firewall.py
```diff
@@ -111,206 +114,238 @@ def __init__(
...
         )
 
-        # Deploy firewall
-        firewall = network.AzureFirewall(
-            f"{self._name}_firewall",
-            application_rule_collections=[
-                network.AzureFirewallApplicationRuleCollectionArgs(
-                    action=network.AzureFirewallRCActionArgs(
-                        type=network.AzureFirewallRCActionType.ALLOW
+        application_rule_collections: list[
```
jemrobinson: Can we make a new PR that just has the change that moves the application rules to the variable `application_rule_collections = [...]` without adding/removing/changing any rules? This would make it much easier to review what's actually changing here.

data_safe_haven/infrastructure/programs/sre/firewall.py
jemrobinson: `list_b_common + [...]` as I think `.append()` is in-place.

data_safe_haven/infrastructure/programs/sre/firewall.py
jemrobinson: Or in a separate pre-PR that just does this.

data_safe_haven/infrastructure/programs/sre/firewall.py
jemrobinson: Can we not use `props.subnet_workspaces_prefixes` here?

data_safe_haven/resources/dns_server/AdGuardHome.mustache.yaml
```diff
@@ -11,14 +11,19 @@ dns:
...
   {{#filter_allow}}
   - "@@||{{.}}"
   {{/filter_allow}}
+{{/allow_workspace_internet}}
 log:
   verbose: true
 # Note that because we are only providing a partial config file we need the
 # `schema_version` key or the full set of YAML migrations will get run.
-schema_version: 24
+schema_version: 24
```
jemrobinson: Minor - can you restore the empty line at the end of this file?

tests/infrastructure/programs/sre/test_firewall.py
```diff
@@ -0,0 +1,179 @@
+import pulumi
+import pulumi.runtime
+import pytest
```
jemrobinson: `from pytest import fixture`?

.github/workflows/lint_code.yaml
JimMadge: I'm wondering whether we need to expand them. We end up linting a YAML file we will never use because it is completed with the dummy values.

However, that relies on their being a tool which is aware of the mustache sections and can skip over them. If it doesn't, I think completing the YAML with some minimal valid values is probably the most sensible thing to do.

tests/mustache/AdGuardHome.mustache.config.json
jemrobinson: Where is this file actually used?

jemrobinson (changes requested):

data_safe_haven/infrastructure/programs/sre/firewall.py
craddm: Indeed, this isn't a problem any more after the changes to the network rules were made.

tests/mustache/AdGuardHome.mustache.config.json
JimMadge: In the yaml linting workflow prior to running yamllint.

.github/workflows/lint_code.yaml
JimMadge: Probably best to push this to a new issue.

data_safe_haven/infrastructure/programs/sre/firewall.py
JimMadge: I'm not actually sure what this means, I can't see any other instances of this class.

data_safe_haven/infrastructure/programs/sre/firewall.py
JimMadge: Doing that breaks some container services https://github.com/alan-turing-institute/data-safe-haven/pull/2327#pullrequestreview-2481906814

JimMadge: Although, that looks to be because the application rules are also all removed.

craddm: Yes, a more principled way might be to keep the application rules for the containers and and then use `subnet_workspaces_prefixes`

data_safe_haven/infrastructure/programs/sre/firewall.py
jemrobinson: I think the suggestion was to add "allow" or "deny" to all firewall rule names.

data_safe_haven/infrastructure/programs/sre/firewall.py
jemrobinson: Yes, we shouldn't remove the rules that are (correctly) allowing other things to work. Neither do we necessarily want a blanket "all resources can access the internet" just to let the workspaces do so. Might make the rule logic a bit more complicated but I think it's worth doing.

JimMadge: Agreed. Let's try that.

data_safe_haven/resources/dns_server/AdGuardHome.mustache.yaml
JimMadge: Yes, that bothers me too 😆.

LF > CR LF and newlines before EOF are correct 😄.

.github/scripts/expand_mustache.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+echo '{"array": ["dummy"], "variable": "dummy"}' > .mustache_config.json
+
+while read -r yamlfile; do
+
+    filename=$(basename -- "$yamlfile")
+    filename="${filename%.*}"
+    test_config=".github/resources/$filename.config.json"
+
+    if [ -e "$test_config" ]; then
```
jemrobinson, line 10: Would it be better to do this for all mustache files we want to expand? It's nicer than the hack of replacing all the mustache markers with `array` or `variable` that's used in the other `if` branch?

data_safe_haven/infrastructure/programs/sre/firewall.py
```diff
@@ -111,140 +114,163 @@ def __init__(
...
         )
 
-        # Deploy firewall
-        firewall = network.AzureFirewall(
-            f"{self._name}_firewall",
-            application_rule_collections=[
-                network.AzureFirewallApplicationRuleCollectionArgs(
-                    action=network.AzureFirewallRCActionArgs(
-                        type=network.AzureFirewallRCActionType.ALLOW
+        application_rule_collections_common = [
```
jemrobinson, line 117: Can we confirm that these rules are identical to the previous rules except where deliberately changed? This sounds difficult. One way to convince ourselves could be to make a separate PR from develop that *just* makes the change of putting the rules into the `application_rule_collections_common` list without changing the rules themselves. It would then be easier to see what changes are being made here.

data_safe_haven/infrastructure/programs/sre/firewall.py
```diff
@@ -111,140 +114,163 @@ def __init__(
...
+                        source_addresses=props.subnet_user_services_software_repositories_prefixes,
+                        target_fqdns=PermittedDomains.SOFTWARE_REPOSITORIES_PYTHON,
+                    ),
+                ],
+            ),
+        ]
+
+        if props.allow_workspace_internet:
+            application_rule_collections = application_rule_collections_common
+            network_rule_collections = [
```
jemrobinson, line 249: Can you add a comment explaining why this needs to be a network rule not an application rule? I think possibly because application rules only apply to HTTP(S) but maybe I'm wrong.

data_safe_haven/resources/dns_server/AdGuardHome.mustache.yaml
```diff
@@ -11,12 +11,17 @@ dns:
...
+user_rules: []
+{{/allow_workspace_internet}}
+{{^allow_workspace_internet}}
 user_rules:
   # https://github.com/AdguardTeam/AdGuardHome/wiki/Hosts-Blocklists#adblock-style-syntax
   - "*.*"
   {{#filter_allow}}
   - "@@||{{.}}"
   {{/filter_allow}}
+{{/allow_workspace_internet}}
```
jemrobinson: Would this logic be better as:

```
user_rules:
  # https://github.com/AdguardTeam/AdGuardHome/wiki/Hosts-Blocklists#adblock-style-syntax
  {{#filter_block}}
  - "{{.}}"
  {{/filter_block}}
  {{#filter_allow}}
  - "@@||{{.}}"
  {{/filter_allow}}
```

with the default being: `filter_block = ["*.*"], filter_allow = ["some", "urls", "here"]` and the allow-all being `filter_block = [], filter_allow = ["*.*"]`? I'm not totally sure whether this would work - thoughts?

jemrobinson: Alternatively, we generate the `@@||` prefixes in the calling script and just have

```
user_rules:
  # https://github.com/AdguardTeam/AdGuardHome/wiki/Hosts-Blocklists#adblock-style-syntax
  {{#filters}}
  - "{{.}}"
  {{/filters}}
```

which might be the simplest/cleanest of all.

tests/infrastructure/programs/sre/conftest.py
```diff
@@ -1,8 +1,27 @@
...
 from pulumi_azure_native import managedidentity, network, resources
 from pytest import fixture
 
 from data_safe_haven.infrastructure.common import SREIpRanges
 
+# Mock configuration.
+
+
+class DataSafeHavenMocks(pulumi.runtime.Mocks):
+    def new_resource(self, args: pulumi.runtime.MockResourceArgs):
```
jemrobinson:
```suggestion
class DataSafeHavenMocks(pulumi.runtime.Mocks):
    """Configuration for Pulumi mocks"""
    def new_resource(self, args: pulumi.runtime.MockResourceArgs):
```

jemrobinson (changes requested):
Looks broadly OK but I have some suggested changes.

.github/scripts/expand_mustache.sh
JimMadge: It would, out of the scope of this PR though. I'll open an issue.

data_safe_haven/infrastructure/programs/sre/firewall.py
JimMadge: I'll compare by hand.

data_safe_haven/resources/dns_server/AdGuardHome.mustache.yaml
JimMadge: I think the first should work.

I don't like the second option because it takes the AdGuard syntax outside of the template.

data_safe_haven/infrastructure/programs/sre/firewall.py
JimMadge: I wasn't very involved in those conversations, do you remember @craddm?

craddm: Been a while and I can't quite remember. But yes, that app rules are only for HTTP(s) does ring a bell - e.g. may want to use ftp for transfer or something like that.

data_safe_haven/infrastructure/programs/sre/firewall.py
JimMadge: Confirmed that `allow_workspace_internet=false` does not change any of the dns rules.

JimMadge:
```
user_rules:
```

and

```
user_rules:
  - @@||*.*
```

don't seem to do what we would like.

---

also

```
user_rules:
  - @@||*
```

JimMadge:
The configuration which I believe last worked was `user_rules: []`.

.github/scripts/expand_mustache.sh
JimMadge: Done https://github.com/alan-turing-institute/data-safe-haven/issues/2362

JimMadge:
Seems to work at [29bf408](https://github.com/alan-turing-institute/data-safe-haven/pull/2327/commits/29bf408817e06a6a38eec82be15ac3fe47c04fb4). `user_rules: []` (or at least some explicit list) seems to be necessary.

jemrobinson (approved):
LGTM

JimMadge (approved):

craddm (approved):

Issue #2316, opened 2024-11-28
Allowlist modification support and documentation


JimMadge:
## :white_check_mark: Checklist

- [x] I have searched open and closed issues for duplicates.
- [x] This is a request for a new feature in the Data Safe Haven or an upgrade to an existing feature.
- [x] The feature is still missing in the [latest version](https://github.com/alan-turing-institute/data-safe-haven/releases).
- [x] I have read through the [documentation](https://alan-turing-institute.github.io/data-safe-haven/).
- [x] This isn't an open-ended question (open a [discussion](https://github.com/alan-turing-institute/data-safe-haven/discussions) if it is).

## :strawberry: Suggested change

The CRAN and PyPI allowlists are currently defined "in code". There is no support to modify them, and, if you did modify them, running `dsh sre deploy` would overwrite any modifications you make.

## :steam_locomotive: How could this be done?

- [ ] Add commands to manipulate (upload, show, download) allowlist files
  Or, should this be configuration?
- [ ] Document modifying allowlists

PR #2346, opened 2025-01-06
Adding commands to manipulate allowlists

craddm:

Adds functions for manipulating allowlists for CRAN and PyPi

#2316


data_safe_haven/allowlist/allowlist.py
```diff
@@ -0,0 +1,162 @@
...
+
+T = TypeVar("T", bound="Allowlist")
+
+
+class Allowlist:
+    """Allowlist for packages"""
+
+    @classmethod
+    def from_remote(
+        cls: type[T],
```
jemrobinson: If you add `from typing import Self` above, you should be able to use `cls: Self` here instead of the TypeVar work-around. Added in Python 3.11.

data_safe_haven/infrastructure/programs/sre/software_repositories.py
```diff
@@ -124,31 +124,25 @@ def __init__(
...
-            resources_path / "software_repositories" / "allowlists" / "cran.allowlist"
-        )
         FileShareFile(
             f"{self._name}_file_share_cran_allowlist",
             FileShareFileProps(
-                destination_path=cran_reader.name,
+                destination_path="cran.allowlist",
                 share_name=file_share_nexus_allowlists.name,
-                file_contents=cran_reader.file_contents(),
+                file_contents="",
```
jemrobinson, line 132: Does this mean that we don't upload a default allowlist at all?

craddm: Correct, just a blank file. Could also just not upload a file at all. Not needed at Tiers <3. Think the idea here is to make people actively have to whitelist things (and in the long run, would be nice to integrate with something like https://github.com/uk-tre/packages)

JimMadge: I think the best approach is for Pulumi to provision a file, but to ignore changes.

That way,

- Pulumi tracks the file and can remove it with teardown
- Users are free to modify the contents without Pulumi interfering/overwriting

craddm: Pulumi only seems to detect changes when the local copy in the source changes, not when the remote copy is modified. May still be worth adding an `ignoreChanges` flag anyway

data_safe_haven/infrastructure/programs/sre/software_repositories.py
```diff
@@ -124,31 +124,25 @@ def __init__(
...
                 storage_account_key=props.storage_account_key,
                 storage_account_name=props.storage_account_name,
             ),
             opts=ResourceOptions.merge(
                 child_opts, ResourceOptions(parent=file_share_nexus)
             ),
```
JimMadge:
```suggestion
            opts=ResourceOptions.merge(
                child_opts, ResourceOptions(
                parent=file_share_nexus,
                ignore_changes=["file_contents"],
            ),
```

Just to be on the safe side :smile:

data_safe_haven/allowlist/allowlist.py
```diff
@@ -0,0 +1,120 @@
...
+        sre_resource_group: str,
+        storage_account_name: str,
+    ) -> str:
+        """Get the current package allowlist"""
+
+        # Get the Azure SDK
+        azure_sdk = AzureSdk(subscription_name=context.subscription_name)
+
+        # Get the file share name
+        file_share_name = "software-repositories-nexus-allowlists"
```
JimMadge: It would be nice to do this in a way that we ensure it is the same as what Pulumi will deploy.

data_safe_haven/allowlist/allowlist.py
```diff
@@ -0,0 +1,120 @@
...
+        if repository:
+            file_share_file = f"{repository.value}.allowlist"
```
JimMadge: Similarly, would be good to ensure this is the same as in Pulumi.

data_safe_haven/allowlist/allowlist.py
```diff
@@ -0,0 +1,120 @@
...
+        )
+        return share_file
+
+    @classmethod
+    def remote_exists(
+        cls: type[Self],
+        context: Context,
+        *,
+        # pulumi_config: DSHPulumiConfig,
+        # sre_config: SREConfig,
```
JimMadge:
```suggestion
```

data_safe_haven/infrastructure/programs/sre/software_repositories.py
jemrobinson: `pulumi refresh` should probably check whether the hash of the remote file has changed. If it's not doing that, I don't think that's behaviour we should rely on.

data_safe_haven/allowlist/allowlist.py
```diff
@@ -0,0 +1,120 @@
...
+        sre_resource_group: str,
+        repository: AllowlistRepository,
+        storage_account_name: str,
+    ) -> bool:
+        # Get the Azure SDK
+        azure_sdk = AzureSdk(subscription_name=context.subscription_name)
+
+        # Get the file share name
+        file_share_name = "software-repositories-nexus-allowlists"
+        file_name = f"{repository.value}.allowlist"
```
JimMadge: These bits are duplicated, it would be nice to put them in a class/module variable/function.

data_safe_haven/allowlist/allowlist.py
```diff
@@ -0,0 +1,120 @@
...
+        *,
+        storage_account_name: str,
+        sre_resource_group: str,
+        repository: AllowlistRepository,
+        allowlist: str,
+    ) -> None:
+        # Get the Azure SDK
+        azure_sdk = AzureSdk(subscription_name=context.subscription_name)
+        file_share_name = "software-repositories-nexus-allowlists"
+        file_name = f"{repository.value}.allowlist"
```
JimMadge: And here.

data_safe_haven/allowlist/allowlist.py
```diff
@@ -0,0 +1,120 @@
...
+
+    @classmethod
+    def remote_diff(
+        cls: type[Self],
+        context: Context,
+        *,
+        sre_resource_group: str,
+        storage_account_name: str,
+        # pulumi_config: DSHPulumiConfig,
+        # sre_config: SREConfig,
```
JimMadge:
```suggestion
```

data_safe_haven/allowlist/allowlist.py
```diff
@@ -0,0 +1,120 @@
...
+        repository: AllowlistRepository,
+        allowlist: str,
+    ) -> list[str]:
+        # Get the Azure SDK
+        azure_sdk = AzureSdk(subscription_name=context.subscription_name)
+
+        # Get the file share name
+        file_share_name = "software-repositories-nexus-allowlists"
+        file_name = f"{repository.value}.allowlist"
```
JimMadge: Here too :smile:

data_safe_haven/commands/allowlist.py
```diff
@@ -0,0 +1,199 @@
...
+        str,
+        typer.Argument(help="Name of SRE to upload the allowlist for."),
+    ],
+    file: Annotated[
+        Path,
+        typer.Argument(help="Path to the allowlist file to upload."),
+    ],
+    repository: Annotated[
+        AllowlistRepository,
+        typer.Argument(help="Name of the repository to upload the allowlist for."),
```
JimMadge:
```suggestion
        typer.Argument(help="Repository type of the allowlist."),
```

I think this might be clearer. When you run `dsh allowlist upload --help` does it show all of the possible options?

data_safe_haven/commands/allowlist.py
```diff
@@ -0,0 +1,199 @@
...
+        logger.error(msg)
+        raise typer.Exit(1)
+
+    sre_stack = SREProjectManager(
+        context=context,
+        config=sre_config,
+        pulumi_config=pulumi_config,
+    )
+
+    sre_resource_group = f"{sre_stack.stack_name}-rg"
```
JimMadge: Can we get this from the stack instead?

docs/source/management/allowlist.md
```diff
@@ -0,0 +1,55 @@
+# Managing allowlists
+
+For Tier 3 SREs, the Python and R software packages that users are allowed to download from the PyPi and CRAN repositories are restricted.
+Connection to PyPi and CRAN is achieved using [Sonatype Nexus Repository](https://www.sonatype.com/products/sonatype-nexus-repository).
```
JimMadge:
```suggestion
For Tier 3 SREs, the Python and R software packages that users are allowed to download from the PyPI and CRAN repositories are restricted.
Connection to PyPI and CRAN is achieved using [Sonatype Nexus Repository](https://www.sonatype.com/products/sonatype-nexus-repository).
```

docs/source/management/allowlist.md
```diff
@@ -0,0 +1,55 @@
...
+blob
+data.table
+DBI
+hms
+lubridate
+RPostgres
+withr
+:::
+::::
+```
```
JimMadge: I think this might look clearer if the blocks are not nested.

JimMadge (commented):
Looks good.

A few suggestions to reduce duplication but otherwise this is great :+1:.

data_safe_haven/commands/allowlist.py
craddm: It does show the options, yes:

```
Usage: dsh allowlist upload [OPTIONS] NAME FILE REPOSITORY:{cran|pypi} 

Upload a package allowlist                                                                                                                                  
                                                                                                                                                             
╭─ Arguments ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ *    name            TEXT                    Name of SRE to upload the allowlist for. [default: None] [required]                                          │
│ *    file            PATH                    Path to the allowlist file to upload. [default: None] [required]                                             │
│ *    repository      REPOSITORY:{cran|pypi}  Name of the repository to upload the allowlist for. [default: None] [required]    
```

docs/source/management/allowlist.md
```diff
@@ -0,0 +1,55 @@
...
+
+```{code} shell
+dsh allowlist upload YOUR_SRE_NAME PATH_TO_ALLOWLIST_FILE REPOSITORY_NAME
+```
+
+The local allowlist file does not need to have a specific name.
+
+## Example allowlists
+
+Example allowlists for PyPi and CRAN can be generated using {typer}`dsh allowlist template`
```
JimMadge:
```suggestion
Example allowlists for PyPI and CRAN can be generated using {typer}`dsh allowlist template`
```

docs/source/reference/allowlist.md
```diff
@@ -0,0 +1,10 @@
+# `allowlist`
+
+`dsh allowlist` commands are used to manage which packages are allowed to be downloaded from software repositories in Tier 2 and Tier 3 SREs.
```
JimMadge:
```suggestion
`dsh allowlist` commands are used to manage which packages are allowed to be downloaded from software repositories in Tier 3 SREs.
```

JimMadge: Would also be good to have a link to the explanation of tiers here.

docs/source/reference/index.md
```diff
@@ -25,6 +26,9 @@ All commands begin with `dsh`.
 The subcommands can be used to manage various aspects of a Data Safe Haven deployment.
 For further detail on each subcommand, navigate to the relevant page.
 
+[Allowlist](allowlist.md)
+: Management of the package allowlists for Tier 2 and Tier 3 SREs
```
JimMadge:
```suggestion
: Management of the package allowlists for Tier 3 SREs
```

data_safe_haven/allowlist/allowlist.py
```diff
@@ -0,0 +1,118 @@
...
+        azure_sdk.upload_file_share(
+            allowlist,
+            self.filename,
+            self.sre_resource_group,
+            self.storage_account_name,
+            self.share_name,
+        )
+
+    @classmethod
+    def remote_diff(
```
JimMadge: Is this the same as with the configs?

It might be cleaner to have a diff method for the `Allowlist` so you could do something like,

```Python
local = Allowlist(...)
remote = Allowlist.from_remote(...)
diff = local.diff(remote)
```

where

```Python
def diff(self, other: Allowlist) -> str:
   ...
```

docs/source/reference/allowlist.md
JimMadge: [This page](https://data-safe-haven.readthedocs.io/en/latest/design/security/objectives.html) I think.

JimMadge (commented):
Looks good, just a couple of style suggestions.

data_safe_haven/allowlist/allowlist.py
craddm: Yes, this was modelled on the `remote_yaml_diff` method for configs (well, for `AzureSerialisableModel`s). Did wonder about doing it this way, so will try it.

JimMadge: In that case, let's open an issue and use another PR for that so we can do both together.

JimMadge (approved):

PR #2367, opened 2025-01-20
Fix docker image CI script

jemrobinson:

n/a

I noticed that the Docker image updating script was breaking due to `chmey/clamav-mirror` not having any numbered versions. This fixes that issue.

n/a

### :microscope: Tests

Tested in a local run.

### DIFF
--- a/.github/scripts/update_docker_versions.py
+++ b/.github/scripts/update_docker_versions.py
@@ -68,7 +68,7 @@ def annotate(
             ):
                 continue
             annotated.append((version_str, version_))
-    return annotated
+    return sorted(annotated, key=lambda v: v[1], reverse=True)
 
 
 for filename in (pathlib.Path("data_safe_haven") / "infrastructure").glob("**/*.py"):
@@ -90,7 +90,7 @@ for filename in (pathlib.Path("data_safe_haven") / "infrastructure").glob("**/*.
                     print(f"No stable releases identified for {image}!")
                     v_latest = v_current
                 else:
-                    v_latest = sorted(candidate_versions, key=lambda v: v[1], reverse=True)[0][0]
+                    v_latest = candidate_versions[0][0]
                 if v_current != v_latest:
                     print(f"Updating {image} from {v_current} to {v_latest} in {filename}")  # noqa: T201
                     needs_replacement = True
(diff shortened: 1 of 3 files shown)


.github/scripts/update_docker_versions.py
```diff
@@ -81,8 +85,12 @@ def annotate(
                     image, v_current, available = get_quayio_versions(image_details)
                 else:
                     image, v_current, available = get_dockerhub_versions(image_details)
-                stable_versions = [v for v in annotate(available, stable_only=True)]
-                v_latest = sorted(stable_versions, key=lambda v: v[1], reverse=True)[0][0]
+                # Consider only stable versions unless there are none available
+                if not (candidate_versions := [v for v in annotate(available, stable_only=True)]):
```
JimMadge:
```suggestion
                if not (candidate_versions := annotate(available, stable_only=True)):
```

Isn't this the same?

.github/scripts/update_docker_versions.py
```diff
@@ -81,8 +85,12 @@ def annotate(
...
+                    print(f"No stable releases identified for {image}!")
+                    v_latest = v_current
+                else:
+                    v_latest = sorted(candidate_versions, key=lambda v: v[1], reverse=True)[0][0]
```
JimMadge: I think it would be clearer to put the logic for finding the latest release into a function, or do the sort in `annotate`.

I'm guessing from this that `version.Version`s are sortable?

JimMadge (commented):
Looks good.

A few suggestions for tidying the script.

Does the packaging library handle versions like `1.2.3` and `v1.2.3` both?

Also, the clamav mirror image looks to be inactive. Should we adopt it?

.github/scripts/update_docker_versions.py
jemrobinson: Yes - not sure why it was written this way before.

.github/scripts/update_docker_versions.py
jemrobinson: Done in ee67777

jemrobinson:
> Also, the clamav mirror image looks to be inactive. Should we adopt it?

Sounds reasonable. I think this is the source repository: https://github.com/chmey/docker-clamav-mirror/.

JimMadge (approved):

PR #2374, opened 2025-01-21
Create GitHub Action to upload to PyPI on release

### LINKED ISSUE #2373: Improve build instructions

- [ ] This is a request for a new feature in the Data Safe Haven or an upgrade to an existing feature.
- [ ] The feature is still missing in the [latest version](https://github.com/alan-turing-institute/data-safe-haven/releases).

We should probably update our build instructions to be explicit about what to do. For example

```
git clean -fxd # delete any untracked git files
git checkout -- . # unstage any changes
hatch clean # remove previous build artifacts
hatch build # build the tarball and wheel releases
hatch publish # push to PyPI
```

jemrobinson:

n/a

Add a GitHub action that should upload to PyPI when a release is made.

Structure comes from [this python.org tutorial](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/). It doesn't seem to require a user-managed shared secret between GitHub and PyPI

> This guide relies on PyPI’s [trusted publishing](https://docs.pypi.org/trusted-publishers/) implementation to connect to [GitHub Actions CI/CD](https://github.com/features/actions). This is recommended for security reasons, since the generated tokens are created for each of your projects individually and expire automatically.

Closes #2373

### :microscope: Tests

NOT TESTED!

### DIFF
--- a/.github/workflows/publish_pypi.yaml
+++ b/.github/workflows/publish_pypi.yaml
@@ -1,0 +1,30 @@
+---
+name: Upload to PyPI
+
+# Run workflow on pushes to matching branches
+on:  # yamllint disable-line rule:truthy
+  workflow_dispatch:
+  release:
+    types: [published]
+
+jobs:
+  build-and-publish-to-pypi:
+    name: Build distribution and publish to PyPI
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.12
+
+      - name: Install hatch
+        run: pip install hatch
+
+      - name: Build with hatch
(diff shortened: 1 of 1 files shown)


JimMadge (commented):
The [docs](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#defining-a-workflow-job-environment) show a job with env variables and non-default permissions. Do we need that here?

jemrobinson:
I don't think we need environment variables. We can add an [environment](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-deployments/managing-environments-for-deployment) which will also be verified on the PyPI side (but this isn't required). The non-default permissions do look like they're needed though - I'll add them both now.

JimMadge (approved):

PR #2389, opened 2025-01-28
Stop unneeded deployment of Nexus repositories for Tier0/1 SREs

### LINKED ISSUE #2382: Don't use Nexus for T0/T1

T0/T1 instance have unfiltered outbound internet access, so there is no need for the Nexus proxies for PyPI/CRAN. It _could_ provide some caching, but it probably isn't worth it.

We could not deploy the Nexus container group and not template the `/etc/pip.conf` and `/etc/RProfile` files for these tiers.

- [ ] Software repository deployment should be based on `allow_workspace_internet` - if it is allowed, there is no need for repositories
- [ ] Catch incompatible config parameters - if `allow_workspace_internet` is `true`, then all software packages should be allowed
- [ ] Desired state should either template the pip or R config files differently or don't template them at all
- [ ] Modify allowlist functions to catch situations where there are no allowlists
- [ ] Ensure software repository related outputs from user services are handled correctly when there are no software repositories.

craddm:

Makes deployment of software repositories and their respective configurations contingent on the `allow_workspace_internet` configuration option. They are not required when workspace internet is allowed.

In summary:

When, in the SRE config file, the `allow_workspace_internet` is set to `false`:

* We won't be deploying Nexus as proxy for PyPI/CRAN: this means that the container group `-container-group-remote-desktop` and the firewall rules it depends on won't be present.
* SRDs will be connecting directly to PyPi/CRAN for package installation, which means we do not need to edit their configuration in Ansible.
* We do not believe this will open an attack surface: when `allow_workspace_internet` is set to `false` users can already bypass Nexus by, for example, using the `--index-url` option of `pip`.  While we will loose the caching offered by Nexus,  it will be compensated by the monetary savings of not deploying container instances and the simplified administration.

Closes #2382 

### :microscope: Tests

We tested the following feature in 2 settings. Here's on a T0 SRE:

image

And here' on a T2 SRE, using Nexus for package installation:

image


data_safe_haven/commands/allowlist.py
```diff
@@ -45,6 +45,18 @@ def show(
...
+        logger.info(
+            "No package allowlist is required for this SRE. "
+            "All packages are allowed."
+        )
+        raise typer.Exit()
+    elif sre_config.sre.software_packages == SoftwarePackageCategory.NONE:
+        logger.info(
+            "No package allowlist is required for this SRE. No packages are allowed."
+        )
+        raise typer.Exit()
```
JimMadge: This block is repeated below, we should move it to a function/method.

data_safe_haven/resources/workspace/ansible/desired_state.yaml
```diff
@@ -45,6 +45,7 @@
     - name: Configure package proxies
       ansible.builtin.import_tasks: tasks/package_proxy.yaml
       tags: package_proxies
+      when: not allow_workspace_internet | bool
```
JimMadge: For clarity, maybe it would be better to call the variable something like `use_package_proxies`.

tests/commands/test_allowlist.py
```diff
@@ -64,6 +65,30 @@ def test_show(
...
+        mocker.patch.object(
+            SREConfig, "from_remote_by_name", return_value=sre_config_any_packages
+        )
+        mocker.patch.object(Allowlist, "from_remote", return_value=mock_allowlist)
+        result = runner.invoke(
+            allowlist_command_group,
+            ["show", sre_name, repository],
+        )
+        assert result.exit_code == 0
+        assert "required" in result.output
```
JimMadge: It would be good to be more specific on this test.

JimMadge (commented):
Looks good.

A few code style suggestions.

Have you tested this in a new deployment.

JimMadge:
@cptanalatriste do you have capacity to resolve the comments here?

cptanalatriste:
@JimMadge at the moment I do not have bandwidth to jump into this and test it properly. However, I expect to base the cost-saving features to CVD-Net in this logic, so I should come back to this PR in the near future.

cptanalatriste:
> Looks good.
> 
> A few code style suggestions.
> 
> Have you tested this in a new deployment.

I just test it on a new deployment. After re-running cloud-init (most likely due to this https://github.com/alan-turing-institute/data-safe-haven/issues/2458) `pip` seems to be working as expected: 
image

data_safe_haven/commands/allowlist.py
cptanalatriste: Will fix!

tests/commands/test_allowlist.py
cptanalatriste: I will assert over the whole error message, to make it more clear

data_safe_haven/commands/allowlist.py
```diff
@@ -11,11 +11,29 @@
 from data_safe_haven.exceptions import DataSafeHavenConfigError, DataSafeHavenError
 from data_safe_haven.infrastructure import SREProjectManager
 from data_safe_haven.logging import get_logger
-from data_safe_haven.types import AllowlistRepository
+from data_safe_haven.types import AllowlistRepository, SoftwarePackageCategory
 
 allowlist_command_group = typer.Typer()
 
 
+def is_allowlist_required(sre_config: SREConfig) -> tuple[bool, str | None]:
```
jemrobinson: I think this should return `tuple[bool, str]` with the default being `(True, "A package allowlist is required for this SRE.")

data_safe_haven/commands/allowlist.py
```diff
@@ -44,6 +62,10 @@ def show(
         raise typer.Exit(1) from exc
 
     sre_config = SREConfig.from_remote_by_name(context, name)
+    allowlist_required, error_message = is_allowlist_required(sre_config)
+    if not allowlist_required:
+        logger.info(error_message)
```
jemrobinson:
```suggestion
    allowlist_required, reason = is_allowlist_required(sre_config)
    if not allowlist_required:
        logger.info(reason)
```

data_safe_haven/commands/allowlist.py
```diff
@@ -136,6 +158,11 @@ def upload(
         raise typer.Exit(1)
     sre_config = SREConfig.from_remote_by_name(context, name)
 
+    allowlist_required, error_message = is_allowlist_required(sre_config)
+    if not allowlist_required:
+        logger.info(error_message)
+        raise typer.Exit()
```
jemrobinson:
```suggestion
    allowlist_required, reason = is_allowlist_required(sre_config)
    if not allowlist_required:
        logger.info(reason)
        raise typer.Exit()
```

data_safe_haven/resources/workspace/ansible/desired_state.yaml
cptanalatriste: Will add the `use_package_proxies` variable to `pulumi_vars.yaml`

data_safe_haven/commands/allowlist.py
cptanalatriste: I like that! I'll make the changes

jemrobinson:
> We do not believe this will open an attack surface

Do we still allow the SRDs to connect to the nexus proxy private IP range? Should we allow this? Could anything else be deployed there (deliberately or accidentally) that might represent a data egress risk?

_N.B. It's OK if the answer is "we don't think this is worth changing"_

jemrobinson:
@cptanalatriste can you confirm that a) a tier 0/1 and b) a tier 2/3 work as-expected when deployed following these changes? I'm particularly wondering whether `/etc/pip.conf` and `/etc/RProfile` are set correctly.

cptanalatriste:
> > We do not believe this will open an attack surface
> 
> Do we still allow the SRDs to connect to the nexus proxy private IP range? Should we allow this? Could anything else be deployed there (deliberately or accidentally) that might represent a data egress risk?
> 
> _N.B. It's OK if the answer is "we don't think this is worth changing"_

Can you clarify what do you mean by "nexus proxy private IP range". At the moment, we do allow outbound connections from the workspaces subnet to the software-repositories subnet, that contains the container group with `nexus`, `caddy`, and `nexus-allowlist`.

Also, I would love to hear from @craddm and @JimMadge , since they were involved from the beginning in the inception of this feature.

cptanalatriste:
> @cptanalatriste can you confirm that a) a tier 0/1 and b) a tier 2/3 work as-expected when deployed following these changes? I'm particularly wondering whether `/etc/pip.conf` and `/etc/RProfile` are set correctly.

For evidence for it working properly in a T0 SRE, please refer to https://github.com/alan-turing-institute/data-safe-haven/pull/2389#issuecomment-3113332077 . Here's evidence of it working properly on T2: 
image

jemrobinson:
@cptanalatriste : can you put these screenshots in the top-level PR comment? Makes it easier for anyone who wants to look back on this PR in future.

craddm:
> > We do not believe this will open an attack surface
> 
> Do we still allow the SRDs to connect to the nexus proxy private IP range? Should we allow this? Could anything else be deployed there (deliberately or accidentally) that might represent a data egress risk?
> 
> _N.B. It's OK if the answer is "we don't think this is worth changing"_

It does still allow them to connect to the nexus proxy private IP range but 1) nothing else is deployed there 2) this is Tier 0/1 so we're fine with data egress, no?

I think at the point I moved off this, I'd been looking for a simple way to simply excise the software repositories subnet and rules completely; I think the simplest would be to create the lists (e.g. the list of subnets that get created during the networking stage of the deployment) outside the function that does the creation of the network/NSGs, append the extra items if the `allow_workspace_internet` flag is set, and then pass the complete list to the creating function.

data_safe_haven/infrastructure/programs/sre/desired_state.py
```diff
@@ -181,6 +183,7 @@ def __init__(
                 ldap_user_filter=props.ldap_user_filter,
                 ldap_user_search_base=props.ldap_user_search_base,
                 software_repository_hostname=props.software_repository_hostname,
+                use_package_proxies=not props.allow_workspace_internet,
```
jemrobinson: Should this be `use_software_repositories` for consistency with how we name the Nexus repos elsewhere?

data_safe_haven/infrastructure/programs/sre/networking.py
```diff
@@ -2070,103 +1548,734 @@ def __init__(
...
+                priority=NetworkingPriorities.ALL_OTHER,
+                protocol=network.SecurityRuleProtocol.ASTERISK,
+                source_address_prefix="*",
+                source_port_range="*",
+            ),
+        ]
+
+        return [
+            inbound_rule for inbound_rule in inbound_rules if inbound_rule is not None
+        ]
```
jemrobinson: This feels messy. If we're doing this in a function, why not only add the rule if it's needed, rather than adding a rule all the time, but sometimes it's None?

data_safe_haven/infrastructure/programs/sre/networking.py
```diff
@@ -2070,103 +1548,734 @@ def __init__(
...
+
+    def get_nsg_data_nexus_inbound_rule_args(self) -> network.SecurityRuleArgs | None:
```
jemrobinson: When is this ever None?

data_safe_haven/infrastructure/programs/sre/networking.py
```diff
@@ -2070,103 +1548,734 @@ def __init__(
...
+                direction=network.SecurityRuleDirection.OUTBOUND,
+                name="DenyAllOtherOutbound",
+                priority=NetworkingPriorities.ALL_OTHER,
+                protocol=network.SecurityRuleProtocol.ASTERISK,
+                source_address_prefix="*",
+                source_port_range="*",
+            ),
+        ]
+
+        return [
```
jemrobinson: Same as above. Why are we allowing None in this list? Especially, why is that then part of the return type, when I think it could just be an implementation detail.

data_safe_haven/infrastructure/programs/declarative_sre.py
```diff
@@ -140,26 +144,37 @@ def __call__(self) -> None:
...
+                sre_networking_props,
+                tags=self.tags,
+            )
+        else:
+            networking = SRENetworkingComponentWithoutNexus(
+                "sre_networking",
+                self.stack_name,
+                sre_networking_props,
+                tags=self.tags,
+            )
```
jemrobinson: Not sure I like the idea of complete separate components here when the main difference is that we're removing some rules from a list. I think this would be handled better as an initialisation argument to `SRENetworkingComponent`

data_safe_haven/infrastructure/programs/sre/networking.py
```diff
@@ -399,73 +402,14 @@ def __init__(
             opts=child_opts,
             tags=child_tags,
         )
-        nsg_data_configuration = network.NetworkSecurityGroup(
+        self.nsg_data_configuration = network.NetworkSecurityGroup(
             f"{self._name}_nsg_data_configuration",
             location=props.location,
             network_security_group_name=f"{stack_name}-nsg-data-configuration",
             resource_group_name=props.resource_group_name,
             security_rules=[
```
jemrobinson: I think we'd be better off defining security rules in-line above, for example with something like:

```python
security_rules = [
    network.SecurityRuleArgs(...),
    ...
    network.SecurityRuleArgs(...),
]
if use_software_repositories:
  security_rules.extend([
    network.SecurityRuleArgs(...),
    ...
    network.SecurityRuleArgs(...),
  ])
```

data_safe_haven/infrastructure/programs/sre/networking.py
```diff
@@ -1403,168 +1259,13 @@ def __init__(
...
-                    destination_address_prefix="*",
-                    destination_port_range="*",
-                    direction=network.SecurityRuleDirection.OUTBOUND,
-                    name="DenyAllOtherOutbound",
-                    priority=NetworkingPriorities.ALL_OTHER,
-                    protocol=network.SecurityRuleProtocol.ASTERISK,
-                    source_address_prefix="*",
-                    source_port_range="*",
-                ),
+                *self.get_nsg_workspaces_outbound_rule_args(),
```
jemrobinson: Same comment as for security rules above

jemrobinson (changes requested):

data_safe_haven/infrastructure/programs/sre/desired_state.py
cptanalatriste: I was using the name suggested by @JimMadge  (see: https://github.com/alan-turing-institute/data-safe-haven/pull/2389#discussion_r1957942603) but I agree `use_software_repositories` is more widely used in the codebase. I'll make the changes.

data_safe_haven/infrastructure/programs/sre/networking.py
cptanalatriste: I thought the order NSG rules were defined on the list affected evaluation, but it appears [evaluation order  only depends on priority](https://learn.microsoft.com/en-us/azure/virtual-network/network-security-groups-overview). I'll make the changes

cptanalatriste: And regarding the messiness, some people consider filtering list comprehensions more Pythonic :) (see: https://stackoverflow.com/questions/71423159/append-to-list-if-not-none-within-a-list-comprehension )

data_safe_haven/infrastructure/programs/sre/networking.py
cptanalatriste: When the method is overwritten by a network configuration without Nexus proxies, like `SRENetworkingComponentWithoutNexus` (overwritten methods should have the same method signature. See: https://stackoverflow.com/questions/61146406/python-overriding-type-hint-on-a-methods-return-in-child-class-without-redefin )

data_safe_haven/infrastructure/programs/sre/networking.py
jemrobinson: I think that argument's more about filtering an existing sequence then defining a new one. If you want to make a list comprehension work here, I'm sure it's possible though :).

data_safe_haven/infrastructure/programs/sre/networking.py
cptanalatriste: I'll replace the list comprehension filter by the if/else as described in https://github.com/alan-turing-institute/data-safe-haven/pull/2389#discussion_r2259596202. Regarding the rationale for `None` return types, please refer to https://github.com/alan-turing-institute/data-safe-haven/pull/2389#discussion_r2259627562

data_safe_haven/infrastructure/programs/sre/networking.py
jemrobinson: Right - so if we keep this as a single class, we won't need the None here?

jemrobinson:
@cptanalatriste : I think there *is* a way of writing this logic as multiple classes if that's the way you want to go, but I think it would be cleaner to do with composability or mixins instead of inheriting functions that you override to return None.

data_safe_haven/infrastructure/programs/declarative_sre.py
cptanalatriste: I got inspired by that approach when I got an alert from Ruff, about having boolean arguments in functions (see: https://docs.astral.sh/ruff/rules/boolean-type-hint-positional-argument/ ). But if it's the convention of the team/codebase, I'll go with that approach. I might need to silence some Ruff alerts.

cptanalatriste: I'll go and silence this alert: 
image

jemrobinson: This isn't necessarily the wrong approach, but I think it's slightly inverted. If you want the two classes approach, you probably want the "without Nexus" class to be the base, and "with Nexus" to be the extension. You can do this with inheritance/mixins or you could do it by composing the class to add capabilities (see e.g. the [decorator pattern](https://refactoring.guru/design-patterns/decorator)).

N.B. I do agree with the Ruff guideline that having a single class with complicated sets of bools to turn behaviour on and off is a bad thing.

data_safe_haven/infrastructure/programs/sre/networking.py
cptanalatriste: By using functions, I was trying to partially alleviate the class/method issue that's specially pervasive in `networking.py` (see https://stackoverflow.com/questions/475675/when-is-a-function-too-long for indications that class/methods might be too long). But if in-lining is the code/team convention, I'll go ahead and write the changes.

data_safe_haven/infrastructure/programs/sre/networking.py
cptanalatriste: Will change. See https://github.com/alan-turing-institute/data-safe-haven/pull/2389#discussion_r2259778054 for original rationale

craddm:
LGTM

PR #2403, opened 2025-02-12
Revert "⬆️ Bump cryptography from 44.0.0 to 44.0.1"

craddm:
Reverts alan-turing-institute/data-safe-haven#2402

Breaks tests and doc building. `acme` is not currently compatible with `josepy=2.0.0`


jemrobinson:
For future reference - you need to close-then-reopen the dependabot PRs to get the tests to run.

Issue #2221, opened 2024-10-03
Path clash with Azure local storage


craddm:
## :white_check_mark: Checklist

- [x] I have searched open and closed issues for duplicates.
- [x] This is a problem observed when **deploying** a Data Safe Haven.
- [x] I can reproduce this with the [latest version](https://github.com/alan-turing-institute/data-safe-haven/releases).
- [x] I have read through the [documentation](https://alan-turing-institute.github.io/data-safe-haven/).
- [x] This isn't an open-ended question (open a [discussion](https://github.com/alan-turing-institute/data-safe-haven/discussions) if it is).

## :computer: System information

- **Operating System:** Debian bookworm
- **Data Safe Haven version:** develop

## :package: Packages

<details>
<summary>List of packages</summary>

```noen
acme==2.10.0
annotated-types==0.7.0
appdirs==1.4.4
Arpeggio==2.0.2
attrs==24.2.0
azure-common==1.1.28
azure-core==1.31.0
azure-identity==1.18.0
azure-keyvault-certificates==4.8.0
azure-keyvault-keys==4.9.0
azure-keyvault-secrets==4.8.0
azure-mgmt-compute==33.0.0
azure-mgmt-containerinstance==10.1.0
azure-mgmt-core==1.4.0
azure-mgmt-dns==8.1.0
azure-mgmt-keyvault==10.3.1
azure-mgmt-msi==7.0.0
azure-mgmt-rdbms==10.1.0
azure-mgmt-resource==23.1.1
azure-mgmt-storage==21.2.1
azure-storage-blob==12.23.1
azure-storage-file-datalake==12.17.0
azure-storage-file-share==12.18.0
certifi==2024.8.30
cffi==1.17.1
charset-normalizer==3.3.2
chevron==0.14.0
click==8.1.7
cryptography==43.0.1
-e git+https://github.com/craddm/data-safe-haven.git@a6a6993ad7bbcc02e8f05629fe0fd5ab7154a900#egg=data_safe_haven
debugpy==1.8.6
dill==0.3.9
dnspython==2.6.1
fqdn==1.5.1
grpcio==1.60.2
idna==3.10
isodate==0.6.1
josepy==1.14.0
markdown-it-py==3.0.0
mdurl==0.1.2
msal==1.31.0
msal-extensions==1.2.0
msrest==0.7.1
oauthlib==3.2.2
parver==0.5
portalocker==2.10.1
protobuf==4.25.5
[...]

PR #2432, opened 2025-04-08
Enabling SRE users to write into /mnt/scratch

### LINKED ISSUE #2413: Users cannot write into `/mnt/scratch`

- [x] This is a problem observed when **using** a Data Safe Haven.

- **Operating System:** MacOS
- **Data Safe Haven version:** 5.3.1
- **Browser details:** Google Chrome

Users do not have write permissions into the `/mnt/scratch` directory. See image below:

Image

As a workaround, DSH admins can invoke `sudo` and modify the folder permissions.

A more robust solution would require editing this file: https://github.com/cptanalatriste/data-safe-haven/blob/af7d7e91af505d7d8958251f89a4a225c6cdb71c/data_safe_haven/resources/workspace/workspace.cloud_init.mustache.yaml#L4

@jemrobinson suggests adding `auto, "defaults,nofail"` to the mount options on line 41 should fix it.

cptanalatriste:

At the moment, `/mnt/scratch` is only editable by `root`, without a clear purpose (besides the workaround implemented in https://github.com/alan-turing-institute/data-safe-haven/issues/2221 ). However, it would be convenient for it to accessible to SRE users similarly as `/mnt/shared`. While it is ephemeral storage, it has the advantage that read/writes are faster when compared to network storage.

This PR implements that functionality.

Closes: https://github.com/alan-turing-institute/data-safe-haven/issues/2413

### :microscope: Tests

I deployed an SRE with the current changes and two SRDs. One of them is a Linux `Standard_D2ds_v5` with a temporary disk. As per [the documentation](https://learn.microsoft.com/en-us/azure/virtual-machines/managed-disks-overview), it has a temporary disk at `/dev/sdb`:

image

We can see that the permissions of `/mnt/scratch` now allow writes from other users:

image

In contrast, the `Standard_D2s_v5`  [does not offer local storage](https://learn.microsoft.com/en-us/azure/virtual-machines/sizes/general-purpose/dsv5-series?tabs=sizebasic]). In this case, the permissions of `/mnt/scratch` were not altered:

image

### DIFF
--- a/data_safe_haven/resources/workspace/ansible/tasks/scratchpad.yaml
+++ b/data_safe_haven/resources/workspace/ansible/tasks/scratchpad.yaml
@@ -1,12 +1,10 @@
 ---
-- name: Collect hardware information
-  ansible.builtin.setup:
-    gather_subset:
-      - hardware
-
-- name: Store temporary disks in list
-  ansible.builtin.set_fact:
-    temporary_disks: "{{ hostvars[inventory_hostname].ansible_devices.keys() | map('regex_search', 'sdb.*') | select('string') | list }}"
+- name: Find temporary disk on Azure VM
+  ansible.builtin.shell: |
+    find /dev/disk/azure -name 'resource-part[0123456789]' | sort | sed -n 1p | xargs readlink -f
+  register: search_result
+  changed_when: false
+  ignore_errors: true
 
 - name: Enable user writing on scratchpad.
   ansible.builtin.file:
@@ -15,4 +13,4 @@
     owner: root
     group: root
     mode: "0777"
-  when: "temporary_disks is defined and (temporary_disks|length>0)"
+  when: "search_result.stdout != ''"
--- a/data_safe_haven/resources/workspace/ansible/tasks/scratchpad.yaml
+++ b/data_safe_haven/resources/workspace/ansible/tasks/scratchpad.yaml
@@ -6,4 +6,4 @@
     state: directory
     owner: root
     group: root
-    mode: 0777
+    mode: '0777'
--- a/data_safe_haven/resources/workspace/ansible/tasks/scratchpad.yaml
+++ b/data_safe_haven/resources/workspace/ansible/tasks/scratchpad.yaml
@@ -1,0 +1,9 @@
+---
+
+- name: Enable user writing on scratchpad (cgc6).
+  ansible.builtin.file:
+    path: /mnt/scratch
+    state: directory
+    owner: root
+    group: root
+    mode: 0777
--- a/data_safe_haven/resources/workspace/ansible/tasks/scratchpad.yaml
+++ b/data_safe_haven/resources/workspace/ansible/tasks/scratchpad.yaml
@@ -1,6 +1,6 @@
 ---
 
-- name: Enable user writing on scratchpad (cgc6).
+- name: Enable user writing on scratchpad.
   ansible.builtin.file:
     path: /mnt/scratch
     state: directory
--- a/data_safe_haven/resources/workspace/ansible/tasks/scratchpad.yaml
+++ b/data_safe_haven/resources/workspace/ansible/tasks/scratchpad.yaml
@@ -1,4 +1,12 @@
 ---
+- name: Collect hardware information
+  ansible.builtin.setup:
+    gather_subset:
+      - hardware
+
+- name: Store temporary disks in list
+  ansible.builtin.set_fact:
+    temporary_disks: "{{ hostvars[inventory_hostname].ansible_devices.keys() | map('regex_search', 'sdb.*') | select('string') | list }}"
 
 - name: Enable user writing on scratchpad.
   ansible.builtin.file:
@@ -6,4 +14,5 @@
     state: directory
     owner: root
     group: root
-    mode: '0777'
+    mode: "0777"
+  when: "temporary_disks is defined and (temporary_disks|length>0)"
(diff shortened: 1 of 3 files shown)


data_safe_haven/resources/workspace/ansible/tasks/scratchpad.yaml
```diff
@@ -0,0 +1,9 @@
+---
+
+- name: Enable user writing on scratchpad.
+  ansible.builtin.file:
+    path: /mnt/scratch
+    state: directory
+    owner: root
+    group: root
+    mode: '0777'
```
JimMadge: I think we want to make this conditional on a scratch disk existing.

JimMadge: Otherwise, we are making it a bit trivial to consume all of the root devices storage which will break user sessions.

JimMadge (changes requested):

data_safe_haven/resources/workspace/ansible/tasks/scratchpad.yaml
cptanalatriste: @JimMadge I updated the PR with the requested changes. I have also added some comments regarding the tests performed.

data_safe_haven/resources/workspace/ansible/tasks/scratchpad.yaml
```diff
@@ -0,0 +1,18 @@
+---
+- name: Collect hardware information
+  ansible.builtin.setup:
+    gather_subset:
+      - hardware
+
+- name: Store temporary disks in list
+  ansible.builtin.set_fact:
+    temporary_disks: "{{ hostvars[inventory_hostname].ansible_devices.keys() | map('regex_search', 'sdb.*') | select('string') | list }}"
```
JimMadge: This won't be reliable, the letters of sdX devices are assigned by the kernel in the order that it recognises them, so it can change on reboot.

It would be better to look in `/dev/disk/` for persistent labels. I think there is a `/dev/disk/azure` directory which should help.

JimMadge (changes requested):

data_safe_haven/resources/workspace/ansible/tasks/scratchpad.yaml
cptanalatriste: Thanks for the feedback @JimMadge . From [Azure's documentation](https://learn.microsoft.com/en-us/troubleshoot/azure/virtual-machines/linux/troubleshoot-device-names-problems), ` /dev/disk/azure-resource-part1` is a symbolic link to the resource/ephemeral disk. In this new revision, I'm relying on that information to determine when to modify `/mnt/scratch` permissions.

data_safe_haven/resources/workspace/ansible/tasks/scratchpad.yaml
```diff
@@ -0,0 +1,16 @@
...
+  ignore_errors: true
+
+- name: Enable user writing on scratchpad.
+  ansible.builtin.file:
+    path: /mnt/scratch
+    state: directory
+    owner: root
+    group: root
+    mode: "0777"
+  when: "search_result.stdout != ''"
```
JimMadge:
```suggestion
- name: Find temporary disk on Azure VM
  ansible.builtin.stat:
    path: /dev/disk/azure/resource-part1
  register: ephemeral_disk_link

- name: Set permissions of scratch directory
  ansible.builtin.file:
    path: /mnt/scratch
    state: directory
    owner: root
    group: root
    mode: "0777"
  when: ephemeral_disk_link.stat.is_link
```

I think it is a much neater/clearer to use Ansible modules than shell commands and text processing.

JimMadge: Even simpler,

```suggestion
- name: Check for scratch directory
  ansible.builtin.stat:
      path: /mnt/scratch
  register: scratch_dir

- name: Set permissions of scratch directory
  ansible.builtin.file:
    path: /mnt/scratch
    state: directory
    owner: root
    group: root
    mode: "0777"
  when: p.stat.isdir is defined and p.stat.isdir
```

As this directory is created in by cloudinit when the ephemeral disk is mounted.

JimMadge: Or

```suggestion
- name: Set permissions of scratch directory
  ansible.builtin.file:
    path: /mnt/scratch
    state: directory
    owner: root
    group: root
    mode: "0777"
  when: /mnt/scratch is mount
```

https://docs.ansible.com/ansible/latest/collections/ansible/builtin/mount_test.html#ansible-collections-ansible-builtin-mount-test

JimMadge (changes requested):

data_safe_haven/resources/workspace/ansible/tasks/scratchpad.yaml
cptanalatriste: @JimMadge that's not the case. The `/mnt/scratch`directory is created and mounted on the OS disk, when there's not temporary disk on the VM

cptanalatriste: image

cptanalatriste: image

JimMadge: Oh, that is surprising. I don't think that is what we would want ideally.

In any case, checking for the symlink to the device or checking that the dir is a mountpoint should work.

JimMadge (approved):

PR #2433, opened 2025-04-10
Documentation: Deploying SREs requires Owner role

cptanalatriste:
Indicating the necessary role for deploying SREs

### DIFF
--- a/docs/source/deployment/deploy_sre.md
+++ b/docs/source/deployment/deploy_sre.md
@@ -19,7 +19,7 @@ $ az provider register --name Microsoft.Network
 ## Requirements
 
 - An Azure subscription where you will deploy your infrastructure.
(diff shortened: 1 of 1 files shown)


docs/source/deployment/deploy_sre.md
```diff
@@ -16,6 +16,11 @@ $ az provider register --name Microsoft.Network
 
 ::::
 
+## Requirements
+
+- An Azure subscription where you will deploy your infrastructure.
+    - An account with at least [Owner](https://learn.microsoft.com/en-us/azure/role-based-access-control/built-in-roles/privileged) permissions on this subscription.
```
JimMadge:
```suggestion
- An Azure subscription where you will deploy your infrastructure.
- An account with at least [Owner](https://learn.microsoft.com/en-us/azure/role-based-access-control/built-in-roles/privileged) role over the scope of the subscription.
```

A bit of a minor language rework.

I think this makes more sense as a two item list, although I'm open to being told otherwise 😆

JimMadge (approved):

PR #2435, opened 2025-04-28
Enabling configuring the size of Nexus' persistent directory

### LINKED ISSUE #2434: Nexus fails due to lack of storage space

- [ ] This is a problem observed when **using** a Data Safe Haven.
- [ ] I can reproduce this with the [latest version](https://github.com/alan-turing-institute/data-safe-haven/releases).

- **Operating System:** MacOS
- **Data Safe Haven version:** 5.4.1
- **Browser details:** Chrome v134

Running `pip install` from within the SRE fails. After checking the logs on the `container-group-software-repositories`  we found a `No Space left on device` error:

Image

The size of the Nexus container volume is currently hardcoded to 2Gb. Manually editing the quota size of the file share fixes the  problem **if the data isn't corrupted**. Otherwise, deleting the container group and the corresponding file share and a redeploy solve the problem.

cptanalatriste:

At the moment, the size of Nexus' persistent directory ---the `nexus-data` directory which corresponds to the File Share `software-repositories-nexus`--- is hard-coded to 2Gb, which is [insufficient in many cases](https://github.com/alan-turing-institute/data-safe-haven/issues/2434).
In this PR, we enable setting-up its size on the SRE's YAML configuration file.

We are proposing the following configuration:

```
user_services:
  nexus:
    persistent_quota_gb: 3
```

We believe this format will support, in the future, configuring other user services offered by the SRE (`Gitea`, `HedgeDoc`, or other `Nexus` properties), as well enabling/disabling them.

Closes https://github.com/alan-turing-institute/data-safe-haven/issues/2434

### :microscope: Tests

- We have deployed a Tier-2 SRE with a `persistent_quota_gb` of `3`, and verified in Portal that the file share reflects the user provided size.
- We have deployed a Tier-2 SRE  without the `user_services` section, and verified that the file share has the default size of `2Gb`.

### DIFF
--- a/data_safe_haven/config/config_sections.py
+++ b/data_safe_haven/config/config_sections.py
@@ -56,7 +56,7 @@ class ConfigSubsectionNexus(BaseModel, validate_assignment=True):
 
 
 class ConfigSectionUserServices(BaseModel, validate_assignment=True):
-    nexus: ConfigSubsectionNexus = ConfigSubsectionNexus(persistent_quota_gb=2)
+    nexus: ConfigSubsectionNexus = ConfigSubsectionNexus(persistent_quota_gb=10)
 
 
 class ConfigSectionSRE(BaseModel, validate_assignment=True):
--- a/data_safe_haven/config/config_sections.py
+++ b/data_safe_haven/config/config_sections.py
@@ -46,13 +46,17 @@ class ConfigSubsectionRemoteDesktopOpts(BaseModel, validate_assignment=True):
     allow_paste: bool
 
 
-class ConfigSubsectionServiceQuotaGB(BaseModel, validate_assignment=True):
-    nexus: PositiveInt
-
-
 class ConfigSubsectionStorageQuotaGB(BaseModel, validate_assignment=True):
     home: AzurePremiumFileShareSize
     shared: AzurePremiumFileShareSize
+
+
+class ConfigSubsectionNexus(BaseModel, validate_assignment=True):
+    persistent_quota_gb: PositiveInt
+
+
+class ConfigSectionUserServices(BaseModel, validate_assignment=True):
+    nexus: ConfigSubsectionNexus | None = ConfigSubsectionNexus(persistent_quota_gb=2)
 
 
 class ConfigSectionSRE(BaseModel, validate_assignment=True):
@@ -65,9 +69,6 @@ class ConfigSectionSRE(BaseModel, validate_assignment=True):
     data_provider_ip_addresses: list[IpAddress] = []
     remote_desktop: ConfigSubsectionRemoteDesktopOpts
     research_user_ip_addresses: list[IpAddress] | AzureServiceTag = []
-    service_quota_gb: ConfigSubsectionServiceQuotaGB = ConfigSubsectionServiceQuotaGB(
-        nexus=2
-    )
     storage_quota_gb: ConfigSubsectionStorageQuotaGB
     software_packages: SoftwarePackageCategory = SoftwarePackageCategory.NONE
     timezone: TimeZone = "Etc/UTC"
--- a/data_safe_haven/config/config_sections.py
+++ b/data_safe_haven/config/config_sections.py
@@ -56,7 +56,7 @@ class ConfigSubsectionNexus(BaseModel, validate_assignment=True):
 
 
 class ConfigSectionUserServices(BaseModel, validate_assignment=True):
-    nexus: ConfigSubsectionNexus | None = ConfigSubsectionNexus(persistent_quota_gb=2)
+    nexus: ConfigSubsectionNexus = ConfigSubsectionNexus(persistent_quota_gb=2)
 
 
 class ConfigSectionSRE(BaseModel, validate_assignment=True):
--- a/data_safe_haven/config/config_sections.py
+++ b/data_safe_haven/config/config_sections.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 from ipaddress import ip_network
 from itertools import combinations
 
-from pydantic import BaseModel, field_validator
+from pydantic import BaseModel, PositiveInt, field_validator
 
 from data_safe_haven.types import (
     AzureLocation,
@@ -46,6 +46,10 @@ class ConfigSubsectionRemoteDesktopOpts(BaseModel, validate_assignment=True):
     allow_paste: bool
 
 
+class ConfigSubsectionServiceQuotaGB(BaseModel, validate_assignment=True):
+    nexus: PositiveInt
+
+
 class ConfigSubsectionStorageQuotaGB(BaseModel, validate_assignment=True):
     home: AzurePremiumFileShareSize
     shared: AzurePremiumFileShareSize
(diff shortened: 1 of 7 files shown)


JimMadge (commented):
Looks good.

Did you check what the source of the problem was? _I.e._ was it that the one package Nexus was trying to download was >2GB or was it that Nexus wasn't cleaning up old packages?

cptanalatriste:
> Looks good.
> 
> Did you check what the source of the problem was? _I.e._ was it that the one package Nexus was trying to download was >2GB or was it that Nexus wasn't cleaning up old packages?

Hi @JimMadge , the original error was triggered by installing `llvmlite`, but given its size it's unlikely to be the culprit. The `No space left on device` error message suggest [lack of space in the mounted volume](https://support.sonatype.com/hc/en-us/articles/360000096228-What-to-Do-When-the-Blobstore-is-Out-of-Disk-Space) or the [OS Disk](https://stackoverflow.com/questions/24300087/is-it-safe-to-remove-files-in-nexus-tmp-folder). Nexus supports some [cleanup tasks](https://www.sonatype.com/blog/2009/09/nexus-scheduled-tasks) for minimising the chances for this happening again, that I can have a look in another PR.

craddm:
Just another thought - Nexus is caching requested images/packages, and you might be able to shorten the lifespan of the cache items so they get cleaned up more often using some of the cleanup tasks https://support.sonatype.com/hc/en-us/articles/115010182627-How-should-I-configure-Proxy-Repository-Cache-setting

cptanalatriste:
> Just another thought - Nexus is caching requested images/packages, and you might be able to shorten the lifespan of the cache items so they get cleaned up more often using some of the cleanup tasks https://support.sonatype.com/hc/en-us/articles/115010182627-How-should-I-configure-Proxy-Repository-Cache-setting

Hi @craddm! Yeah, definitely there are many options for making Nexus behave better. Sadly, the standard way of enabling them is [via the Web GUI](https://community.sonatype.com/t/how-to-configure-nexus-using-the-approach-infrastructure-as-code/4443). Nexus also offers a [REST API and a Scripting API](https://help.sonatype.com/en/automation.html) that shows promise, but this will require further exploration and way more code. Maybe I can tackle this approach in another PR.

craddm:
One way we interact with the API at the moment is to accept the EULA. We do that with the Nexus Allowlist container, so if you do go that route maybe look at there as a place to do it - https://github.com/alan-turing-institute/nexus-allowlist

JimMadge:
All of the Nexus configuration is done through the API. 2GB is possible too small for a stable installation. I worry that this might be working around the actual issue.

cptanalatriste:
@JimMadge changing the size of quota seems to address the issue every time it presents. And by having the size in the configuration file, we avoid having it restored manually after every redeployment (that we do quite often due to other issues). If you require more evidence, I'm happy to close this PR and continue with the investigation.

craddm:
It seems most likely to me that it's just the cache outgrowing the file share (which'll happen the more and more packages people use!) rather than anything inherently unstable about having 2gb drive for installation

cptanalatriste:
> It seems most likely to me that it's just the cache outgrowing the file share (which'll happen the more and more packages people use!) rather than anything inherently unstable about having 2gb drive for installation

I checked the container logs, and the `No space left on device` errors are triggered when writing to:

* `/nexus-data/elasticsearch/nexus/nodes`,  related to Elaticsearch index size. This can be addressed by expanding the quota.
* The Blobstore (triggering a `BlobStoreException`). The blob store is also within `nexus-data`.
* The JVMs temporary directory (`../sonatype-work/nexus3/tmp`). Probably stored in the OS Disk, and not sure we can modify it when working with ACIs.

@craddm , what are the signals that it is a cache issue?

craddm:
The blob store is the cache and elasticsearch is indexing it (AFAIK), so both of those would be issues with the size of the cache

cptanalatriste:
@craddm [in the link you shared](https://support.sonatype.com/hc/en-us/articles/115010182627-How-should-I-configure-Proxy-Repository-Cache-setting) I cannot find a parameter regarding the total cache size, only component lifecycle. Components requested can be invalidated and deleted (and potentially [soft-deleted](https://community.sonatype.com/t/when-nexus-decides-to-delete-blobs-and-which-blobs-if-nothing-configured-to-do-so/7796/3)) but they will be replaced with the component's current version.

And if we use Cache and Blob Store interchangeably, there's a [soft-quota](https://help.sonatype.com/en/configuring-blob-stores.html#soft-quota-162038) system that's aware of its size and availability (it appears we're using it at DSH, the screenshot shows an SRE with a `nexus-data` of size 3Gb). However, it cannot block write operations, only generate warnings.

Anyway, my point is that I believe it's useful for DSH admins to configure the size of Nexus' persistent directory (including Blob Store size), and they would benefit from the functionality of this PR.

image

data_safe_haven/config/config_sections.py
```diff
@@ -51,6 +51,14 @@ class ConfigSubsectionStorageQuotaGB(BaseModel, validate_assignment=True):
     shared: AzurePremiumFileShareSize
 
 
+class ConfigSubsectionNexus(BaseModel, validate_assignment=True):
+    persistent_quota_gb: PositiveInt
+
+
+class ConfigSectionUserServices(BaseModel, validate_assignment=True):
+    nexus: ConfigSubsectionNexus = ConfigSubsectionNexus(persistent_quota_gb=2)
```
JimMadge: If this value seems to be a problem in production, and we provide a default value, we should make the default something that is more likely to be suitable.

JimMadge:
> @JimMadge changing the size of quota seems to address the issue every time it presents. And by having the size in the configuration file, we avoid having it restored manually after every redeployment (that we do quite often due to other issues). If you require more evidence, I'm happy to close this PR and continue with the investigation.

I think it is a useful addition. Happy to approve when we have set the default value to something sensible. I'm just still not quite sure what is happening.

For example, would setting the quota to 100GB fix the problem, or would it mean you just encounter it last and with a greater cost for storage?

data_safe_haven/config/config_sections.py
craddm: Sonatype recommend 20 Gb for a small setup - https://help.sonatype.com/en/nexus-repository-reference-architecture-1.html

Our usage profile is way smaller than that in terms of requests/hour, etc, but I don't think that in itself is relevant to how much space the cache packages and associated metadata/indices will take up

craddm: I also note on that page it says "Required disk space for Nexus Repository varies by deployment size and complexity. You must have at least 4GB available disk space at all times; if available disk space drops below 4GB, the database will switch to read-only mode." so I assume there's something I'm missing here as that doesn't match the behaviour with a 2Gb file share (is it counting an ephemeral/OS disk attached to the container as well?)

craddm:
Yep, fair point about the cache invalidation. "Compact blob stores" is probably what we want it do -  https://help.sonatype.com/en/cleanup-policies.html

(anyway just to be clear am on board with making this configurable but also a sensible default to start with would be good!)

data_safe_haven/config/config_sections.py
JimMadge: The quota on the object storage might not be visible to Nexus (it is object storage right?).

cptanalatriste: We're using 4GB of RAM in the container instead of the recommended 8GB for a small setting. Maybe we can start with 10GB of blob storage instead of the recommended 20GB? And if it's inadequate, admins can always change it via the YAML file (FYI, we're currently at 8GB for `nexus-data` in Prod).

cptanalatriste:
> For example, would setting the quota to 100GB fix the problem, or would it mean you just encounter it last and with a greater cost for storage?

I do not have an answer to that question at the moment. From the evidence I gathered so far, increasing `nexus-data` size seems to solve the problem. In Prod, I setup it size to `8Gb` and it's been going on well without issues (so far!).

If the issue reappears, I can revisit and do a deeper dive (I still suspect on the OS disk size for ACI containers...). But in either case, I think this feature is useful.

cptanalatriste:
I changed the default quota to 10GB: It's half the recommended blob store size for small settings, and closer to the 8Gb we're currently using in Prod.

JimMadge (approved):

Issue #2099, opened 2024-08-06
ClamAV: clamd.ctl file is not created


jemrobinson:
## :white_check_mark: Checklist

- [x] I have searched open and closed issues for duplicates.
- [x] This is a problem observed when **managing** a Data Safe Haven.
- [x] I can reproduce this with the [latest version](https://github.com/alan-turing-institute/data-safe-haven/releases).
- [x] I have read through the [documentation](https://alan-turing-institute.github.io/data-safe-haven/).
- [x] This isn't an open-ended question (open a [discussion](https://github.com/alan-turing-institute/data-safe-haven/discussions) if it is).

## :computer: System information

- **Operating System:** 
- **Data Safe Haven version:** 

## :package: Packages

<details>
<summary>List of packages</summary>

```none
Paste list of packages here
```

</details>

## :no_entry_sign: Describe the problem

The `clamav-daemon` service is not creating `/var/run/clamav/clamd.ctl`. This causes the `clamav-clamonacc` service to timeout waiting for this file to exist. 

## :steam_locomotive: Workarounds or solutions

See [this StackOverflow thread](https://askubuntu.com/questions/1170774/clamav-clamd-ctl-file-is-not-getting-created-on-ubuntu)

craddm:
Note that this stops subsequent steps from running, and `desired-state.sh` needs to be run manually complete the process. On our deployments for pentesting, this meant that smoke tests were not copied to the appropriate locations.

jemrobinson:
What's the best fix for this? My first guesses would be either:

- disable on-access virus scanning (I'm not sure this is a requirement of DSPT and it does add a lot of compute overhead)
- allow this ansible task to fail with `ignore_errors: true`

What do you think @JimMadge @craddm?

JimMadge:
I think we should be able to wait until the file is created.

JimMadge:
https://docs.ansible.com/ansible/latest/collections/ansible/builtin/wait_for_module.html

craddm:
> What's the best fix for this? My first guesses would be either:
> 
> * disable on-access virus scanning (I'm not sure this is a requirement of DSPT and it does add a lot of compute overhead)
> * allow this ansible task to fail with `ignore_errors: true`
> 
> What do you think @JimMadge @craddm?

On-access virus scanning is a DSPT requirement - 

"Antivirus/anti-malware software scans files automatically upon access. This includes when files are downloaded and opened, and when they are accessed from a network folder."

[...]

PR #2460, opened 2025-07-16
Using the correct socket on the unit file for clamav-clamonacc

### LINKED ISSUE #2458: Cannot log into SRDs due to missing nslcd restart.

- [ ] This is a problem observed when **deploying** a Data Safe Haven.
- [ ] I can reproduce this with the [latest version](https://github.com/alan-turing-institute/data-safe-haven/releases).

- **Operating System:** macOS 
- **Data Safe Haven version:** 5.5.0

## :package: Packages

<details>

```none
Paste list of packages here
```

</details>

After a fresh deployment of Tier-2 and Tier-3 SREs, is not possible to log into the SRDs with registered users. Evidence at: https://alan-turing-institute.slack.com/archives/C0428BBM94M/p1751618514209919

After a debbuging session with @craddm , we discovered in that in both SREs the ansible playbook failed before reaching LDAP configuration. In the following, notice that the the last task is `Enable and start ClamAV on access scan` on 03/07/2025, before the next Ansible run on 04/07/2025.

Image

The `nlscd` deamon is restarted only once, after `cloud-init` finishes (see https://github.com/alan-turing-institute/data-safe-haven/blob/develop/data_safe_haven/resources/workspace/workspace.cloud_init.mustache.yaml), and at this point LDAP is not properly setup. Although Ansible fixes the problem later, the `nlscd` is still in an invalid state.

@craddm believes is related to https://github.com/alan-turing-institute/data-safe-haven/issues/2221 . We discussed implementing a fix where daemon restart happens within an Ansible Playbook (if possible), so the next Ansible run would fix the problem , if this reappears.

**WORKAROUND:** Restartin `nlscd` manually with `systemctl restart nslcd` solves the problem.

<details>

Partial logs of ClamAV Ansible playbook failing:

Image

LDAP errors on Syslog, after a failed log attempt:

Image

ClamaAV failing also on Tier-3:

Image

</details>

cptanalatriste:

N/A

At the moment, `clamav.conf` states that the UNIX domain socket should be created at `/tmp/clamd.socket` and the unit file for `clamav-clamonacc` requires the socket file to start this service, otherwise it fails to start. When `clamav-clamonacc` fails to start, the Ansible playbook aborts and many services end misconfigured (like LDAP).

We verified, and other users have reported, that for the socket configuration in `clamav.conf` to take place  you need to restart or reconfigure the ClamAV daemon (see: https://github.com/alan-turing-institute/data-safe-haven/issues/2458  for evidence). However, a closer inspection of the ClamAV unit file shows that out-of-the-box a socket file is created at `/run/clamav/clamd.ctl` (managed by the `clamav-daemon.socket` unit file). The socket file is created before the daemon starts, and its ready to be used by the `clamav-clamonacc` service.

In this PR: 1) We indicate ClamaAV to use the `/run/clamav/clamd.ctl` file as local socket, 2) Remove the `ExecStartPre` logic from the `clamav-clamonacc` unit file, since this dependency is managed by `systemd`.

Closes https://github.com/alan-turing-institute/data-safe-haven/issues/2458 and https://github.com/alan-turing-institute/data-safe-haven/issues/2099

### :microscope: Tests

These are the results in at T2 SRE. After this changes were implemented:

`clamav-clamonacc` is active and running:

image

`clamav-daemon` is active and running:

image

The socket filed configured by `clamav-daemon.socket` is present:

image

### DIFF
--- a/data_safe_haven/resources/workspace/ansible/files/etc/systemd/system/clamav-clamonacc.service
+++ b/data_safe_haven/resources/workspace/ansible/files/etc/systemd/system/clamav-clamonacc.service
@@ -6,7 +6,6 @@ After=clamav-daemon.service syslog.target network.target
 [Service]
 Type=simple
 User=root
-ExecStartPre=/bin/bash -c "while [ ! -S /tmp/clamd.socket ]; do sleep 1; done"
 ExecStart=/usr/sbin/clamonacc --foreground=true
 Restart=on-failure
 RestartSec=30
--- a/data_safe_haven/resources/workspace/ansible/files/etc/clamav/clamd.conf
+++ b/data_safe_haven/resources/workspace/ansible/files/etc/clamav/clamd.conf
@@ -1,5 +1,5 @@
 # Path to a local socket file the daemon will listen on.
-LocalSocket /tmp/clamd.socket
+LocalSocket /run/clamav/clamd.ctl
 # Sets the permissions on the unix socket to the specified mode.
 LocalSocketMode 660
 # Prevent access to infected files for normal users


jemrobinson (approved):
LGTM but I'd like to hear @JimMadge and/or @craddm's thoughts on whether this was set to `/tmp` for a reason.

data_safe_haven/resources/workspace/ansible/files/etc/systemd/system/clamav-clamonacc.service
```diff
@@ -6,7 +6,6 @@ After=clamav-daemon.service syslog.target network.target
 [Service]
 Type=simple
 User=root
-ExecStartPre=/bin/bash -c "while [ ! -S /tmp/clamd.socket ]; do sleep 1; done"
 ExecStart=/usr/sbin/clamonacc --foreground=true
```
JimMadge, line 9:
```suggestion
ExecStartPre=/bin/bash -c "while [ ! -S /run/clamav/clamd.ctl ]; do sleep 1; done"
ExecStart=/usr/sbin/clamonacc --foreground=true
```

JimMadge: This is still important as clamonacc requires clamd to be running.

JimMadge (changes requested):
> LGTM but I'd like to hear @JimMadge and/or @craddm's thoughts on whether this was set to /tmp for a reason.

IIRC simply because that was the example given in the ClamAV docs.
`/tmp` has the sticky bit set, which helps as users can access the socket but not delete it.

`/run` seems like a sensible place, actually the [FHS](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s15.html) says that transient sockets must be in `/run` or a subdirectory.

data_safe_haven/resources/workspace/ansible/files/etc/systemd/system/clamav-clamonacc.service
JimMadge: Or... Is it covered by the Requires and After?

JimMadge: [Should be](https://www.man7.org/linux/man-pages/man5/systemd.unit.5.html)

JimMadge (approved):

PR #2463, opened 2025-07-18
Update Azure SDK packages

jemrobinson:
## :white_check_mark: Checklist

- [x] You have given your pull request a meaningful title (_e.g._ `Enable foobar integration` rather than `515 foobar`).
- [x] You are targeting the appropriate branch. If you're not certain which one this is, it should be **`develop`**.
- [x] Your branch is up-to-date with the **target branch** (it probably was when you started, but it may have changed since then).

### :arrow_heading_up: Summary

Several Azure SDK packages have updates that drop support for particular `vYYYY-MM-DD` versions that we've been using in `azure_sdk.py`. This PR updates all the `azure-X-Y` packages to the latest version and removes dependencies on specific SDK versions.

### :closed_umbrella: Related issues

Merging this should fix the currently-broken dependabot PRs (e.g. #2462).

### :microscope: Tests

I **have not made a test deployment** but I have run the automated tests after making each change.

Issue #2270, opened 2024-10-29
Backup not functional


JimMadge:
## :white_check_mark: Checklist

- [x] I have searched open and closed issues for duplicates.
- [x] This is a problem observed when **managing** a Data Safe Haven.
- [x] I can reproduce this with the [latest version](https://github.com/alan-turing-institute/data-safe-haven/releases).
- [x] I have read through the [documentation](https://alan-turing-institute.github.io/data-safe-haven/).
- [x] This isn't an open-ended question (open a [discussion](https://github.com/alan-turing-institute/data-safe-haven/discussions) if it is).

## :computer: System information

- **Operating System:** 
- **Data Safe Haven version:** 5.0.1

## :package: Packages

<details>
<summary>List of packages</summary>

```none
Paste list of packages here
```

</details>

## :no_entry_sign: Describe the problem

Backup not functional, workaround in docs doesn't work.

A major problem here is, the storage account types we have chosen (for good reasons of performance, ability to upload/download to/from) **are not** compatible with Azure backup services.

## :steam_locomotive: Workarounds or solutions

Because of the incompatibility, I think our two options are

1. Change the type of storage accounts we use to ones which are compatible
    - This will mean we loose nice features like performance, nice integration with NFS, upload/download using Azure storage explorer
    - Quite a significant infrastructure change to make
2. Implement our own backup
    - Requires writing/supporting our own solution
    - There are off-the-shelf tools we could use (I think Borg/Borgmatic)
    - Should be fairly simple to implement with a VM or container, as we can mount all of our storage as NFS
    - I have a proposal for how this would look using Borg/Borgmatic in the [comments below](https://github.com/alan-turing-institute/data-safe-haven/issues/2270#issuecomment-2449367024)

JimMadge:
Following the docs, fixing the backup instance fails (after 22 minutes :scream:)

> Fixing protection error for BlobBackupSensitiveData
UserErrorMissingRequiredPermissions: Appropriate permissions to perform the operation is missing.

Possibly because my user is missing some backup roles, I am an Owner and Storage Blob Data Owner though.

JimMadge:
The problem may be that the backup vault doesn't have the correct role assigned in the target storage account.

JimMadge:
[...]

PR #2466, opened 2025-07-29
Fix library update

cptanalatriste:

N/A

A recent library update (see: https://github.com/alan-turing-institute/data-safe-haven/pull/2463/files ) broke the SRE deployment process. This PR fixes it.

Note that using updated versions of the library made deployment fail due to well-know problems with our backup system (see: https://github.com/alan-turing-institute/data-safe-haven/issues/2270 ) . As part of this PR, we remove the deployment of backup.

### :microscope: Tests

We deployed a Tier-2 SRE from the current branch, and deployment succeeded: 

image

### DIFF
--- a/data_safe_haven/infrastructure/programs/sre/backup.py
+++ b/data_safe_haven/infrastructure/programs/sre/backup.py
@@ -179,7 +179,7 @@ class SREBackupComponent(ComponentResource):
             principal_type=authorization.PrincipalType.SERVICE_PRINCIPAL,
             role_assignment_name=str(
                 seeded_uuid(
-                    f"{stack_name} Storage Account Backup Contributor for Backup Value"
+                    f"{stack_name} Storage Account Backup Contributor for Backup Vault"
                 )
             ),
             role_definition_id=Output.concat(
--- a/data_safe_haven/infrastructure/programs/sre/backup.py
+++ b/data_safe_haven/infrastructure/programs/sre/backup.py
@@ -8,6 +8,8 @@ from pulumi_azure_native import authorization, dataprotection
 
 from data_safe_haven.functions import seeded_uuid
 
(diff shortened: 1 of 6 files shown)


data_safe_haven/infrastructure/programs/sre/backup.py
```diff
@@ -162,6 +173,25 @@ def __init__(
             ),
         )
 
+        authorization.RoleAssignment(
+            f"{self._name}_backup_instance_dns_zone_contributor_role_assignment",
+            principal_id=backup_vault.identity.principal_id,
+            principal_type=authorization.PrincipalType.SERVICE_PRINCIPAL,
+            role_assignment_name=str(
+                seeded_uuid(
+                    f"{stack_name} Storage Account Backup Contributor for Backup Value"
```
jemrobinson:
```suggestion
                    f"{stack_name} Storage Account Backup Contributor for Backup Vault"
```

data_safe_haven/infrastructure/programs/sre/backup.py
```diff
@@ -162,6 +173,25 @@ def __init__(
             ),
         )
 
+        authorization.RoleAssignment(
```
jemrobinson, line 176: Does adding this role fix anything? I see you've removed the backup component from our deployment workflow in any case.

jemrobinson (approved):
LGTM

data_safe_haven/infrastructure/programs/sre/backup.py
cptanalatriste: Without the role-assignment, deployment fails due to permissions. With this role assignment, it fails due to unsupported account type. I guess it makes us closer to back up deployment when/if Azure supports our storage accounts.

PR #2476, opened 2025-09-04
Hotfix: Using the correct socket on the unit file for clamav-clamonacc

cptanalatriste:


cptanalatriste:
Here's the review when merged in Develop: https://github.com/alan-turing-institute/data-safe-haven/pull/2460

JimMadge (approved):

PR #2482, opened 2025-09-23
Automatically ingressing code from pre-approved GitHub repositories

### LINKED ISSUE #1995: Gitea mirroring security requirements

cptanalatriste:

N/A

### Expected behaviour
TRE admins will include in the config file the information of the external repository they want to ingress into the TRE. The config file looks like this:

```yaml
user_services:
  dns_sidecar:
    cron_expression: "*/30 * * * *"
    replica_timeout: 600
    retry_limit: 0
  nexus:
    persistent_quota_gb: 10
  gitea_mirror:
    repositories:
      - repository_name: data-safe-haven
        repository_url: https://github.com/cptanalatriste/data-safe-haven
        repository_auth_token: TOKEN_ACCESS_TO_THE_REPO_TO_INGRESS
```
Once the deployment is finished, the Gitea server currently available from workspaces will have a new user `workspaceuser`, containing mirrors to the external repositories defined in the YAML config file (see screenshots in the "Test" section). This repository will be automatically updated **every 10 minutes**.

In a similar spirit as [with Nexus proxies](https://github.com/alan-turing-institute/data-safe-haven/pull/2389), we're only enabling this functionality for SREs with Tier-2 and Tier-3.  For SREs with Tier-0/Tier-1, we're disabling this functionality, as they already have access to the Internet. 

### Implementation details

We have added a new Container Group, `-container-group-mirror-manager`, to be deployed in its own subnet, `UserServicesGiteaMirrorSubnet`. This container group has the following containers:

1. The `gitea` container, that create a migration to the external GitHub repository.
2. The `mirrormanager` container, that provides a Python program that does the following:
    -  Per each external repository in the config file, it creates a [repository mirror ](https://docs.gitea.com/next/usage/repo-mirror)in the `gitea` container in Step 1. using [Gitea's REST API.](https://docs.gitea.com/api/1.24/#tag/repository/operation/repoMigrate).
    - Per each repository mirror created in the previous step, we also add [a push mirror ](https://docs.gitea.com/next/usage/repo-mirror#pushing-to-a-remote-repository)to the Gitea server accessible from the workspaces. We also rely on [Gitea's REST API to accomplish this](https://docs.gitea.com/api/1.24/#tag/repository/operation/repoAddPushMirror).

Closes https://github.com/alan-turing-institute/data-safe-haven/issues/1995

### :microscope: Tests

We have deployed a Tier-2 SRE with a single repository to migrate. We checked that the migration is completed, and a push mirror is available from the Workspaces under the user we created:

image

We have also seen the repository been updated as new commits were added.

Also, we have deployed a Tier-1 SRE. We have checked it is still operational, and none of the components of the Gitea mirror are deployed:

image


data_safe_haven/config/sre_config.py
```diff
@@ -135,5 +137,14 @@ def template(cls: type[Self], tier: int | None = None) -> SREConfig:
                     replica_timeout="Maximum number of seconds a DNS sidecar job is allowed to run.",  # type: ignore
                     retry_limit="Maximum number of retries before failing the DNS sidecar job.",  # type: ignore
                 ),
+                gitea_mirror=ConfigSubsectionGiteaMirror.model_construct(
+                    repositories=[
+                        GitRepository.model_construct(
+                            repository_name="An identifier for the GitHub repository to mirror.",
+                            repository_url="The URL of the GitHub repository to mirror.",  # type: ignore
+                            repository_auth_token="A GitHub personal access token, with access to the repository to mirror.",
```
jemrobinson: Do we suggest making this a read-only token? If not, then we should.

data_safe_haven/infrastructure/programs/sre/mirror_manager.py
```diff
@@ -0,0 +1,434 @@
...
+    LocalDnsRecordProps,
+    PostgresqlDatabaseComponent,
+    PostgresqlDatabaseProps,
+    WrappedLogAnalyticsWorkspace,
+)
+from data_safe_haven.resources import resources_path
+from data_safe_haven.utility import FileReader
+
+
+class SREGiteMirrorManagerProps:
```
jemrobinson:
```suggestion
class SREGiteaMirrorManagerProps:
```

Also, I suggest calling the file `gitea_mirror_manager.py`

jemrobinson:
@cptanalatriste : I haven't had time to review this properly (and probably won't have time) but this looks very nice! :tada:

To confirm, does this PR do the following:
- add a second Gitea server which has outbound internet access
- add a container to perform the (two?) synchronisation jobs (internet -> external server + external -> internal server)

A couple of things to think about/document:
- what new data breach opportunities might this open up?
- what are the cost implications of this additional infrastructure?

cptanalatriste:
> @cptanalatriste : I haven't had time to review this properly (and probably won't have time) but this looks very nice! 🎉
> 
> To confirm, does this PR do the following:
> 
>     * add a second Gitea server which has outbound internet access
> 
>     * add a container to perform the (two?) synchronisation jobs (internet -> external server + external -> internal server)
> 
> 
> A couple of things to think about/document:
> 
>     * what new data breach opportunities might this open up?
> 
>     * what are the cost implications of this additional infrastructure?

Hi @jemrobinson. Thanks! The CVD-Net team is really excited about these features. Regarding your questions:

-  Yes, we have added a second Gitea server (the `gitea` container at the `container-group-mirror-manager ` container group). The container group is deployed to a new subnet (`UserServicesGiteaMirrorSubnet`), and we allow outbound connections to `github.com` at NSG and Firewall level.
- The synchronisation is handled entirely by the new Gitea server, using out-of-the-box features (pull and push mirrors). We have deployed a new container (`mirrormanager`) with a Python program to set up this exchange using the Gitea API. This program does the following: 1) Creates a pull-mirror for each external GitHub repository, and 2) Creates a push mirror from the pull mirror to the Gitea server accessible from the workspaces.
- I think the biggest security concern is that, once the self-code-ingress is set up, you can bring whatever you want into the SRE. For example, you can bring a dataset for linkage attacks, or even malware. I think we would need some process controls in place, to regularly check what's been ingressed.
- We have an additional container group, which necessarily increases the cost. I will let my sandbox running for a few days to have a sense of the impact.

data_safe_haven/config/sre_config.py
cptanalatriste: Will do!

data_safe_haven/infrastructure/programs/sre/mirror_manager.py
cptanalatriste: Thanks for catching the typo! I will rename the file too

cptanalatriste:
@jemrobinson here some cost numbers for reference. A Tier-1 SRE at the moment is costing about 44GBP per day. A Tier-2 one, that has the Nexus proxy and the Gitea mirror, is costing about 63GBP

JimMadge (commented):
A few questions,

- Does this _only_ support mirroring repositories from GitHub? My understanding was the Gitea migrate function supported any git repository
- I would get rid of the push mirror and just pull twice, external gitea pulls from the internet, internal gitea pulls from external.
  That way we don't give any gitea instance permissions to push to another.
  It should be simplier to implement (using one API endpoint twice).

data_safe_haven/infrastructure/programs/sre/gitea_mirror_manager.py
```diff
@@ -0,0 +1,434 @@
...
+        )
+
+        # Define the container group.
+        self.container_group_name = f"{stack_name}-container-group-mirror-manager"
+        self.container_group = containerinstance.ContainerGroup(
+            f"{self._name}_container_group",
+            container_group_name=self.container_group_name,
+            containers=[
+                containerinstance.ContainerArgs(
+                    image="xr09/python-requests:3.11",
```
JimMadge: Might be sensible here to make our own container image based on Python.

cptanalatriste:
> Does this _only_ support mirroring repositories from GitHub? My understanding was the Gitea migrate function supported any git repository

Hi @JimMadge, If you go to the migration GUI on Gitea there's a list of supported services (see image below). Since I have only tested GitHub, I thought I would limit documentation/code to GitHub only, considering is the service we plan to use in CVD-Net.

> I would get rid of the push mirror and just pull twice, external gitea pulls from the internet, internal gitea pulls from external.

Using only the pull mirror functionality makes sense, I'll start working on those changes.


image

data_safe_haven/infrastructure/programs/sre/gitea_mirror_manager.py
cptanalatriste: Will do! I'll post some changes shortly

cptanalatriste: @JimMadge I have moved this logic to its own container, using only the Pull Mirror endpoint. You can have a look at: https://github.com/alan-turing-institute/gitea-mirror-manager

craddm (approved):
LGTM, not sure if @JimMadge has any further comments

llewelld:
For future reference, configuring repository mirroring requires a per-repository personal access token (PAT). The following instructions can be used to generate one:

1. GitHub > Select Avatar > Settings > Developer Settings > Personal access tokens > Fine-grained tokens > Generate new token.
2. Then enter name, description, an appropriate expiration.
3. Then set "Repository access" to "Only select repositories" and ensure the relevant repository is selected.
4. Then "Add permissions" > select "Contents" > set to "Read-only"
5. Then "Generate token".

PR #2486, opened 2025-10-17
The Gitea mirror should take a new ip range for easy TRE update

cptanalatriste:

N/A

Deploying the [self-code ingress feature](github.com/alan-turing-institute/data-safe-haven/pull/2482) on a TRE deployed with an old DSH version can be problematic since, at the moment, the container group is using a VNet already in use. As part of this change, we're making sure the new components are deployed to a brand-new subnet.

Address a bug introduced in https://github.com/alan-turing-institute/data-safe-haven/pull/2482

### :microscope: Tests

As you can see from this Tier-2 SRE, the Gitea mirror subnet is now after the DNS Sidecar (the last new VNet added):

image

And the mirror is still working:

image


craddm (approved):
LGTM

Issue #2489, opened 2025-10-23
The Virtual Network has the wrong IP address of the DNS Server


cptanalatriste:
## :white_check_mark: Checklist

- [x] I have searched open and closed issues for duplicates.
- [x] This is a problem observed when **deploying** a Data Safe Haven.
- [x] I can reproduce this with the [latest version](https://github.com/alan-turing-institute/data-safe-haven/releases).
- [x] I have read through the [documentation](https://alan-turing-institute.github.io/data-safe-haven/).
- [x] This isn't an open-ended question (open a [discussion](https://github.com/alan-turing-institute/data-safe-haven/discussions) if it is).

## :computer: System information

- **Operating System:** macOS
- **Data Safe Haven version:** v5.6.0

## :package: Packages

<details>
<summary>List of packages</summary>

```none
Paste list of packages here
```

</details>

## :no_entry_sign: Describe the problem

When users attempt to log into the SRE via Guacamole they face an error message after providing credentials:

<img width="812" height="359" alt="Image" src="https://github.com/user-attachments/assets/4457c0d8-5f4a-41df-a1fc-da608783c7e1" />

After checking Guacamole logs, we found it has problem resolving the IP of the Identity container, with the error `java.net.UnknownHostException: login.microsoftonline.com` (see below for full stack trace).

We verify the DNS Server settings for the SRE virtual network:

## :deciduous_tree: Log messages

<details>
<summary>Relevant log messages</summary>

```none
15:35:00.384 [http-nio-8080-exec-6] DEBUG org.jose4j.http.Get - Waiting 720ms before retrying (3 of 3) HTTP GET of https://login.microsoftonline.com/cb94a6f6-ef7a-42ab-bcad-4f0b887cfd3e/discovery/v2.0/keys after failed attempt: {}
java.net.UnknownHostException: login.microsoftonline.com
	at java.base/sun.nio.ch.NioSocketImpl.connect(NioSocketImpl.java:567)
	at java.base/java.net.SocksSocketImpl.connect(SocksSocketImpl.java:327)
	at java.base/java.net.Socket.connect(Socket.java:751)
	at java.base/sun.security.ssl.SSLSocketImpl.connect(SSLSocketImpl.java:304)
	at java.base/sun.net.NetworkClient.doConnect(NetworkClient.java:178)
	at java.base/sun.net.www.http.HttpClient.openServer(HttpClient.java:531)
[...]

PR #2500, opened 2025-11-20
Replacing the DNS Container Instance with a Virtual Machine

### LINKED ISSUE #2488: The workstation has the wrong IP address of the DNS Server

- [x] This is a problem observed when **using** a Data Safe Haven.

- **Operating System:** macOS
- **Data Safe Haven version:** v5.5.1
- **Browser details:** Google Chrome

Users reported being unable to log into a workstation after successfully logging into Guacamole:

Image

Containers involved in logging seemed up and running, and their DNS entries are accurate. When accessing the workstation via serial console. we discovered that `nslcd` had trouble resolving the name `identity....`. When using `resolvectl status` to check the workstation DNS configuration, we obtained the following:

Image

Which is not the IP of the DNS Server used at the TRE:

Image

We believe it's another consequence of issue https://github.com/alan-turing-institute/data-safe-haven/issues/1802, that sadly is not addressed by https://github.com/alan-turing-institute/data-safe-haven/pull/2442.

Redeploying the SRE doesn't not fix it. However, re-running `cloud-init` does. @craddm believes even a VM restart could fix it too.

cptanalatriste:

N/A

As seen in the Related Issues section, the current implementation of our custom DNS Server, using Azure Container Instances, can change its IP number, which produces malfunctions at both workspace at user service level. To address this issue, we replaced this container instance with a **virtual machine with a static IP**, that acts as a container host for our AdGuard Home container.

In this PR, **we have not changed in any way** the AdGuard Home container configuration or functionality. We are just replacing its runtime environment for a VM container host, instead of Azure's container instances. To accomplish this, we did the following:

- We defined a new `SREDnsServerVMComponent`: a virtual machine capable of hosting docker containers (see `dns_server_vm.py`).
- Using `cloud-init` (see: `dns_server_vm.cloud_init.mustache.yaml`) we perform the following setup steps : 
    - Define AdGuardHome configuration files, to later mount them in a  volume when running the container.
    - Disable the `systemd-resolved` daemon, since it is binded to port 53 required to handle DNS requests (see AdGuardHome documentation at:  https://adguard-dns.io/kb/adguard-home/faq/#bindinuse)
    - Install Docker within the VM.
    - Create a service unit configuration file to log into Docker Hub, download the AdGuardHome image, and run the container.
- We also modified the original `SREDnsServerComponent` (see: `dns_server.py`) replacing the container instance for the virtual machine defined in the previous step.
- We also modified the DNS virtual network (also in`dns_server.py`) . In specific, we did the following:
    - In the Network Security Group, we allowed outbound connections to the internet, so the virtual machine can download and install Docker.
    - We removed the delegation to Container Groups. since we won't be using them anymore. 

Fixes https://github.com/alan-turing-institute/data-safe-haven/issues/2488 and https://github.com/alan-turing-institute/data-safe-haven/issues/2489

### :microscope: Tests

To see this feature in action you can check the TRE available at: `https://t2sre.cvdnetdev.develop.turingsafehaven.ac.uk`. There, you can verify:

- The `adguardhome` service is up-and-running:
image

- The `adguardhome` container is handling DNS requests:

image

- Within the workspace, we haven't lost connectivity to services like Nexus or Gitea. And Internet connectivity is still restricted:

image


data_safe_haven/infrastructure/components/composite/virtual_machine.py
```diff
@@ -231,48 +231,52 @@ def __init__(
...
-            provider_name="Microsoft.Compute",
-            resource_group_name=props.resource_group_name,
-            resource_name_=virtual_machine.name,
-            resource_type="VirtualMachines",
-            opts=ResourceOptions.merge(
-                child_opts,
-                ResourceOptions(parent=virtual_machine),
-            ),
-        )
+        if props.maintenance_configuration_id is not None:
```
JimMadge:
```suggestion
        if props.maintenance_configuration_id:
```

data_safe_haven/infrastructure/components/composite/virtual_machine.py
```diff
@@ -231,48 +231,52 @@ def __init__(
...
-                props.data_collection_rule_name, "-association"  # this name is required
-            ),
-            data_collection_rule_id=props.data_collection_rule_id,
-            resource_uri=virtual_machine.id,
-            opts=ResourceOptions.merge(
-                child_opts,
-                ResourceOptions(parent=virtual_machine),
-            ),
-        )
+        if props.data_collection_rule_id is not None:
```
JimMadge:
```suggestion
        if props.data_collection_rule_id:
```

data_safe_haven/infrastructure/components/composite/virtual_machine.py
```diff
@@ -31,6 +29,8 @@ def __init__(
         vm_name: Input[str],
         vm_size: Input[str],
         admin_username: Input[str] | None = None,
+        data_collection_rule_id: Input[str] | None = None,
+        data_collection_endpoint_id: Input[str] | None = None,
```
JimMadge, line 33: Is it the case that these must both be defined or both be `None`? If so, we can add a check, or combine the if blocks somehow.

data_safe_haven/infrastructure/components/composite/virtual_machine.py
```diff
@@ -231,48 +231,52 @@ def __init__(
...
-            f"{name_underscored}_dcra_to_dce",
-            association_name="configurationAccessEndpoint",  # this name is required
-            data_collection_endpoint_id=props.data_collection_endpoint_id,
-            resource_uri=virtual_machine.id,
-            opts=ResourceOptions.merge(
-                child_opts,
-                ResourceOptions(parent=virtual_machine),
-            ),
-        )
+        if props.data_collection_endpoint_id is not None:
```
JimMadge:
```suggestion
        if props.data_collection_endpoint_id:
```

data_safe_haven/infrastructure/programs/sre/dns_server.py
```diff
@@ -194,7 +205,9 @@ def __init__(
             opts=ResourceOptions.merge(
                 child_opts,
                 ResourceOptions(
-                    ignore_changes=["virtual_network_peerings"]
+                    ignore_changes=["virtual_network_peerings"],
+                    delete_before_replace=True,
+                    replace_on_changes=["subnets[*].delegations"],
```
JimMadge, line 240:
```suggestion
```

I _think_ delegations have been removed so this should be needed.

data_safe_haven/infrastructure/programs/sre/dns_server.py
```diff
@@ -206,96 +219,61 @@ def __init__(
...
-                            mount_path="/opt/adguardhome/custom",
-                            name="adguard-opt-adguardhome-custom",
-                            read_only=True,
-                        ),
-                    ],
+        # Deploy maintenance configuration
+        # See https://learn.microsoft.com/en-us/azure/update-manager/scheduled-patching
+        maintenance_configuration = maintenance.MaintenanceConfiguration(
+            f"{self._name}_dns_server_maintenance_configuration",
+            duration="03:55",  # Maximum allowed value for this parameter
```
JimMadge: What does this correspond to, is the maximum what we what?

Do you remember @craddm?

data_safe_haven/infrastructure/programs/sre/dns_server.py
```diff
@@ -206,96 +219,61 @@ def __init__(
...
-            ],
+            resource_name_=f"{stack_name}-dns-server-maintenance-configuration",
+            start_date_time=Output.from_input(props.timezone).apply(
+                lambda timezone: next_occurrence(
+                    hour=2,
+                    minute=4,
+                    timezone=timezone,
+                    time_format="iso_minute",
+                )  # Run maintenance at 02:04 local time every night
+            ),
```
JimMadge: It might be nice to stagger the updates for different components

data_safe_haven/resources/dns_server/dns_server_vm.cloud_init.mustache.yaml
```diff
@@ -0,0 +1,77 @@
...
+      Description=Starts the AdGuardHome container.
+      After=docker.service
+      Requires=docker.service
+
+      [Service]
+      TimeoutStartSec=0
+      Restart=always
+      Environment="HOME=/home/adguardhome"
+      ExecStartPre=/usr/bin/docker login -u {{dockerhub_username}} -p {{dockerhub_access_token}}
+      ExecStart=/usr/bin/docker run --rm --name adguardhome -v /opt/adguardhome/custom:/opt/adguardhome/custom -p 53:53/tcp -p 53:53/udp -p 80:80/tcp --entrypoint /bin/sh {{container_image}} /opt/adguardhome/custom/entrypoint.sh
```
JimMadge:
```suggestion
      ExecStart=/usr/bin/docker run --rm --name adguardhome -v /opt/adguardhome/custom:/opt/adguardhome/custom -p 53:53/tcp -p 53:53/udp --entrypoint /bin/sh {{container_image}} /opt/adguardhome/custom/entrypoint.sh
```

I think this may have been because container instances required exposing port 80, we probably don't need it here.

data_safe_haven/resources/dns_server/dns_server_vm.cloud_init.mustache.yaml
```diff
@@ -0,0 +1,77 @@
...
+  - unattended-upgrades
+package_update: true
+package_upgrade: true
+
+runcmd:
+  # Mount all external volumes
+  # --------------------------
+  - echo ">=== Mounting all external volumes... ===<"
+  - grep -v -e '^[[:space:]]*$' /etc/fstab | sed 's|^|  /etc/fstab  |'
+  - mount -fav
```
JimMadge: Are there external volumes? We might be able to remove this

JimMadge (commented):
Looks good.

In general, deploying a VM to just run docker feels like a bad idea. In the long term, switching to K8s would be better, but that is a big task. As it is important for the DNS to have a static IP address this feel like a sensible workaround.

What is the logging situation here? Having logs of how DNS requests were handled feels important. Can we push logs from the docker process to out central logs, or to SYSLOG which then pushes?

data_safe_haven/infrastructure/components/composite/virtual_machine.py
cptanalatriste: Done!

data_safe_haven/infrastructure/components/composite/virtual_machine.py
cptanalatriste: Done!

cptanalatriste:
> Looks good.
> 
> In general, deploying a VM to just run docker feels like a bad idea. In the long term, switching to K8s would be better, but that is a big task. As it is important for the DNS to have a static IP address this feel like a sensible workaround.
> 
> What is the logging situation here? Having logs of how DNS requests were handled feels important. Can we push logs from the docker process to out central logs, or to SYSLOG which then pushes?

Thanks for all the feedback @JimMadge . After a few refactorings, I managed to connect the DNS server VM to the existing Monitoring component. Docker logs are sent to Syslog, and we can query them in the log analytics workspace:  
image

data_safe_haven/infrastructure/components/composite/virtual_machine.py
cptanalatriste: At the moment, `data_collection_rule_id ` and `data_collection_endpoint_id ` do not depend on each other. Not every data collection rule requires a collection endpoint (see: https://learn.microsoft.com/en-us/azure/azure-monitor/data-collection/data-collection-endpoint-overview?source=recommendations&tabs=portal ).

data_safe_haven/infrastructure/components/composite/virtual_machine.py
cptanalatriste: Done!

data_safe_haven/infrastructure/programs/sre/dns_server.py
cptanalatriste: If we have a TRE deployed with v5.6.0, upgrading will require a change of delegations (i.e. removing them) and deployment will fail.

data_safe_haven/infrastructure/programs/sre/dns_server.py
cptanalatriste: According to Pulumi docs, it's the duration of the maintenance window, in hours. I think we need it to be long enough so all updates are installed during it.

data_safe_haven/infrastructure/programs/sre/dns_server.py
cptanalatriste: We do not need this anymore, since now I'm reusing the already existing Maintenance Configuration

data_safe_haven/resources/dns_server/dns_server_vm.cloud_init.mustache.yaml
cptanalatriste: I tried it, and the DNS server is working fine without exposing port 80. I'll make the change

data_safe_haven/resources/dns_server/dns_server_vm.cloud_init.mustache.yaml
cptanalatriste: Just tested this, and after removal the DNS server works fine. I will make the changes

PR #2542, opened 2026-01-28
Container groups depend on Log Analytics workspace

cptanalatriste:

For each container instance, we're making explicit that they depend on the Log Analytics Workspace. That way, the shared key will be available at the container instance creation time.

### DIFF
--- a/data_safe_haven/infrastructure/programs/sre/apt_proxy_server.py
+++ b/data_safe_haven/infrastructure/programs/sre/apt_proxy_server.py
@@ -1,7 +1,7 @@
 from collections.abc import Mapping
 
 from pulumi import ComponentResource, Input, Output, ResourceOptions
-from pulumi_azure_native import containerinstance, network, storage
+from pulumi_azure_native import containerinstance, network, operationalinsights, storage
 
 from data_safe_haven.infrastructure.common import (
     get_id_from_subnet,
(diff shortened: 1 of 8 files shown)


data_safe_haven/infrastructure/programs/sre/apt_proxy_server.py
```diff
@@ -129,7 +129,10 @@ def __init__(
             diagnostics=containerinstance.ContainerGroupDiagnosticsArgs(
                 log_analytics=containerinstance.LogAnalyticsArgs(
                     workspace_id=props.log_analytics_workspace.workspace_id,
-                    workspace_key=props.log_analytics_workspace.workspace_key,
+                    workspace_key=operationalinsights.get_shared_keys_output(
+                        resource_group_name=props.log_analytics_workspace.resource_group_name,
+                        workspace_name=props.log_analytics_workspace.name,
+                    ).apply(lambda keys: keys.primary_shared_key),
```
JimMadge, line 135: This is repeated quite a few times.

I think it would be better to have this as an `Output` somewhere and pass it as an argument.

I think that way Pulumi should recognise the dependency too so there is no need to add the explicit `depends_on`.

JimMadge (changes requested):
I think we should try to be more DRY here and retrieve the key once.

Will this stop a race condition which can cause SRE deployment to fail?

JimMadge (approved):
Fine to merge as is but lets open an issue for the improvement to be a future patch release.

data_safe_haven/infrastructure/programs/sre/apt_proxy_server.py
cptanalatriste: I have added issue https://github.com/alan-turing-institute/data-safe-haven/issues/2543 with your suggestions. I wIll address them on a following PR

PR #2546, opened 2026-01-30
Adding support for OS disk size specification on TRE config file.

### LINKED ISSUE #2545: The workspace's hard disk runs out of space

- **Operating System:**  macOs
- **Data Safe Haven version:**  v5.6.0

## :package: Packages

<details>

```none
Paste list of packages here
```

</details>

When try to log into a workspace using RDP, we only observe a grey screen after entering credentials:

Image

It's worth mentioning that SSH works fine. Via the serial console, we observe various error messages regarding lack of space (`OSError: [Errno 28] No space left on device`). Using `df` confirms this:

Image

<details>

```none
Your log details here
```

</details>

cptanalatriste:

N/A

As part of this PR:
- Users are able to define the size of the OS Disk in the SRE config file. For example, defining a 64GB size OS disk is done with the following:
```yaml
        storage_quota_gb:
            home: 100
            shared: 100
            data_disk: 0
            os_disk: 64
```
- The default OS disk size for workspaces in now 64GB, since the default 30GB have proven to be insufficient.
- We're disabling updating the size of the data disk, since at the moment it triggers the creation of a new VM/Disk, erasing the previously present data.

**IMPORTANT!**: It is not possible to update the OS Disk size using the DSH CLI, due to Pulumi limitations. Attempts of doing it will fail with the following error:

```
Pulumi error:  ~  azure-native:compute:VirtualMachine sre_workspaces_vm_workspace_01 **updating failed** [diff: ~storageProfile];    
error: Status=400 Code="Conflict" Message="Disk resizing is allowed only when creating a VM or when the VM is deallocated."                             
```
We expect users to manually update OS Disk size (using the Portal or Azure CLI), and normalise Pulumi state later.

Closes: https://github.com/alan-turing-institute/data-safe-haven/issues/2545

### :microscope: Tests

The workspaces are created with the right disk size:

image

### DIFF
--- a/data_safe_haven/config/config_sections.py
+++ b/data_safe_haven/config/config_sections.py
@@ -51,6 +51,7 @@ class ConfigSubsectionStorageQuotaGB(BaseModel, validate_assignment=True):
     home: AzurePremiumFileShareSize
     shared: AzurePremiumFileShareSize
     data_disk: AzureDataDiskSize = 0
+    os_disk: AzureDataDiskSize = 64
 
 
 class ConfigSubsectionNexus(BaseModel, validate_assignment=True):
--- a/data_safe_haven/infrastructure/components/composite/virtual_machine.py
+++ b/data_safe_haven/infrastructure/components/composite/virtual_machine.py
@@ -31,9 +31,10 @@ class VMComponentProps:
         admin_username: Input[str] | None = None,
         data_collection_rule_id: Input[str] | None = None,
         data_collection_endpoint_id: Input[str] | None = None,
-        data_disk_size: int = 0,
+        data_disk_size: Input[int] = 0,
         ip_address_public: Input[bool] | None = None,
         maintenance_configuration_id: Input[str] | None = None,
+        os_disk_size: Input[int] = 64,
     ) -> None:
         self.admin_password = admin_password
         self.admin_username = admin_username if admin_username else "dshvmadmin"
@@ -48,6 +49,7 @@ class VMComponentProps:
         self.ip_address_public = ip_address_public
         self.location = location
         self.maintenance_configuration_id = maintenance_configuration_id
+        self.os_disk_size = os_disk_size
         self.os_profile_args = None
         self.resource_group_name = resource_group_name
         self.subnet_name = subnet_name
@@ -198,13 +200,14 @@ class VMComponent(ComponentResource):
                 ],
             ),
             os_profile=props.os_profile,
-            resource_group_name=props.resource_group_name,
+        resource_group_name=props.resource_group_name,
             storage_profile=compute.StorageProfileArgs(
                 data_disks=data_disks,
                 image_reference=props.image_reference,
                 os_disk=compute.OSDiskArgs(
                     caching=compute.CachingTypes.READ_WRITE,
                     create_option=compute.DiskCreateOptionTypes.FROM_IMAGE,
+                    disk_size_gb=props.os_disk_size,
                     delete_option=compute.DiskDeleteOptionTypes.DELETE,
                     managed_disk=compute.ManagedDiskParametersArgs(
                         storage_account_type=compute.StorageAccountTypes.PREMIUM_LRS,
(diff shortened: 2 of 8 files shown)


data_safe_haven/infrastructure/components/composite/virtual_machine.py
```diff
@@ -31,9 +31,10 @@ def __init__(
         admin_username: Input[str] | None = None,
         data_collection_rule_id: Input[str] | None = None,
         data_collection_endpoint_id: Input[str] | None = None,
-        data_disk_size: int = 0,
+        data_disk_size: Input[int] = 0,
         ip_address_public: Input[bool] | None = None,
         maintenance_configuration_id: Input[str] | None = None,
+        os_disk_size: Input[int] = 64,
```
martintoreilly, line 37: Can we add a unit tests that verifies that this default is successfully overridden when the user specified config sets a different value? Do the current unit tests validate that the value of the VM properties in the Pulumi definition that deploys the VM are as specified in the user-specified config? If so, it should just be a case of adding a test case that sets the VM size differently (e.g. to 128GB).

data_safe_haven/config/config_sections.py
```diff
@@ -51,6 +51,7 @@ class ConfigSubsectionStorageQuotaGB(BaseModel, validate_assignment=True):
     home: AzurePremiumFileShareSize
     shared: AzurePremiumFileShareSize
     data_disk: AzureDataDiskSize = 0
+    os_disk: AzureDataDiskSize = 64
```
martintoreilly, line 54: Can we add a unit tests that verifies that this default is successfully overridden when the user specified config sets a different value? Do the current unit tests validate that the value of the VM properties in the Pulumi definition that deploys the VM are as specified in the user-specified config? If so, it should just be a case of adding a test case that sets the VM size differently (e.g. to 128GB).

martintoreilly (changes requested):
Code changes LGTM. Note that I was at the desk with Carlos while these changes were made so we applied the second pair of eyes at code change time.

@cptanalatriste Please could you re-run the end to end VM deployment using an OS disk size that is different to the 64GB default specified in the code, to validate that this does override the defaults and propagate the user-specified OS disk size to the deployed VM.

I added a couple of inline notes where the defaults are set asking if we can easily add unit tests that validate the defaults are overridden when a different value is provided as input.

If the end to end test shows the user-specified OS disk value propagating to the deployed VM, I'm happy to merge this PR. If its easy to incorporate unit tests baking this validation into the test suite, it would be great to see these added but I would not consider the lack of these a blocker to merging.

cptanalatriste:
> Code changes LGTM. Note that I was at the desk with Carlos while these changes were made so we applied the second pair of eyes at code change time.
> 
> @cptanalatriste Please could you re-run the end to end VM deployment using an OS disk size that is different to the 64GB default specified in the code, to validate that this does override the defaults and propagate the user-specified OS disk size to the deployed VM.
> 
> I added a couple of inline notes where the defaults are set asking if we can easily add unit tests that validate the defaults are overridden when a different value is provided as input.
> 
> If the end to end test shows the user-specified OS disk value propagating to the deployed VM, I'm happy to merge this PR. If its easy to incorporate unit tests baking this validation into the test suite, it would be great to see these added but I would not consider the lack of these a blocker to merging.

Hi @martintoreilly , I did the requested end-to-end test. Given the new TRE configuration (see the first screenshot), we can see that the deployed OS disk has the size specified (see second screenshot).

image
image

data_safe_haven/infrastructure/components/composite/virtual_machine.py
cptanalatriste: I performed the test, and it succeeded. See comment below

data_safe_haven/config/config_sections.py
cptanalatriste: Sadly, I do not have bandwidth at the moment to add the unit test. But I verified the functionality works as expected with an end-to-end test (see below)

martintoreilly (approved):
Approved.

PR #2558, opened 2026-02-12
Fix dotnet package conflict

JimMadge:

`dotnet-sdk-8.0` and `dotnet-runtime-8.0` conflict between the Ubuntu repositories and Microsoft repository.
This conflicts prevent apt upgrades.

Here, the packages providing dotnet are changed (to more recent releases) which don't conflict.
Also, the apt source and apt key configuration is moved into Ansible to be more visible and make use of Ansible facts.

### :microscope: Tests

Tested on a workspace.
Before and error like,

```
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 dotnet-host-8.0 : Conflicts: dotnet-host
E: Error, pkgProblemResolver::Resolve generated breaks, this may be caused by held packages.
```

After removing the source file `apt upgrade` is successful.

### DIFF
--- a/data_safe_haven/resources/workspace/ansible/host_vars/localhost.yaml
+++ b/data_safe_haven/resources/workspace/ansible/host_vars/localhost.yaml
@@ -37,8 +37,8 @@ package_categories:
       - clojure  # Clojure functional programming language for JVM
       - cmake  # CMake build system
       - default-jre  # Java runtime environment
-      - dotnet-runtime-8.0  # .Net runtime
-      - dotnet-sdk-8.0  # .Net SDK
+      - dotnet-runtime  # .Net runtime
+      - dotnet-sdk  # .Net SDK
       - fsharp  # F# functional-first programming language
       - g++  # GNU C++ compiler
       - gcc  # GNU C compiler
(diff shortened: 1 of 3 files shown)


data_safe_haven/resources/workspace/ansible/host_vars/localhost.yaml
```diff
@@ -37,8 +37,8 @@ package_categories:
       - clojure  # Clojure functional programming language for JVM
       - cmake  # CMake build system
       - default-jre  # Java runtime environment
-      - dotnet-runtime-8.0  # .Net runtime
```
cptanalatriste, line 40: Any idea why it was locked into this version before? Seems a bit strange...

cptanalatriste (approved):
Looks good to me! Only a minor question

data_safe_haven/resources/workspace/ansible/host_vars/localhost.yaml
jemrobinson: It came from these commits (63fa697, 4e2df0a, e0569a5) when we first started configuring these VMs with ansible.

JimMadge: I can't remember. Maybe to pin to a particular version, although I don't think that will change within an Ubuntu release anyway. So, picking one of the numbered packages is just providing an old version of dotnet.

PR #2559, opened 2026-02-12
Ensure dshadmin home directory exists

JimMadge:
Ensures appropriate ownership and permissions

The `dshadmin` home directory isn't created until the user logs in. If we use the serial console the directory is created with ownership `root:root` and so dshadmin is unable to write files. This is possibly because it bypasses `pam_mkhomedir`. In any case, it felt easier to make the directory and it's permissions part of the regular desired state.


cptanalatriste (approved):
Looks good to me!

PR #2565, opened 2026-02-16
Fix ansible error due to apt cache update before key imported

JimMadge:
- Update dotnet and msodbcsql packages
- ~~Replace deprecated method of importing package signing keys~~ Don't do this as 22.04 seems to have old OpenSSL/Curl preventing this.
- Add apt preference to hopefully prevent #2558 in the future


JimMadge:
@cptanalatriste should be ready now. Successful deployment with cloud-init and Ansible.

I'm not keen on leaving a comment in the desired state but I think it is worthwhile as that change will have to be made in the near future.

cptanalatriste (approved):
Looks good to me!

PR #2569, opened 2026-02-25
Refactor access to the shared keys of the Log Analytics Workspace

### LINKED ISSUE #2543: Refactor the dependency between Log Analytics Workspace and Container instances

As par of PR https://github.com/alan-turing-institute/data-safe-haven/pull/2542 @JimMadge made the following suggestions:

> I think we should try to be more DRY here and retrieve the key once.
> Will this stop a race condition which can cause SRE deployment to fail?

> This is repeated quite a few times.
> I think it would be better to have this as an Output somewhere and pass it as an argument.
> I think that way Pulumi should recognise the dependency too so there is no need to add the explicit depends_on

This issue tracks the implementation of these suggestions.

cptanalatriste:

PR https://github.com/alan-turing-institute/data-safe-haven/pull/2572

As stated in https://github.com/alan-turing-institute/data-safe-haven/issues/2543, there was a lot of duplicated code for accessing the Log Analytics Workspace shared keys. That redundancy was introduced to enable the update of TREs from version `v5.6.1` to `v5.7.0`. In this PR, we manage to refactor and centralise that logic, **while making it possible to update TREs from `v5.7.0`** to the release that includes this change. To accomplish that:

1. We replaced the original component with a new one (see `workspace.py`), implemented as a Component Resource following Pulumi's guidelines (see: https://www.pulumi.com/docs/iac/concepts/components/ ). This component will expose the workspace shared key as an output.
2. To ensure TRE update succeeds, we modified some of the components in `monitoring.py` and `monitoring_elements.py`. Some of the changing include: 1) Re-naming resources, 2) Add an alias, 3) Force re-creation instead of update.

Closes https://github.com/alan-turing-institute/data-safe-haven/issues/2543

### :microscope: Tests

1. We deployed a TRE with `data-safe-haven==5.7.0`.
3. We run `dsh sre deploy` from the branch including these changes.
4. The TRE is still up and usable, and the logs are still sent to the Workspace.


JimMadge:
@cptanalatriste does this supersede #2567?

.github/workflows/build_documentation.yaml
```diff
@@ -21,8 +21,8 @@ jobs:
           python-version: 3.12
 
       - name: Install hatch
-        run: pip install hatch
-
+        run: pip install virtualenv==20.39.0 hatch==1.16.3
+        # Temporary workaround. virtualenv==21.0.0 breaks the GH Action.
```
JimMadge: Let's put these changes into a different PR. It will be easier to revert them when there is a patch release of hatch to fix the dependency issue.

data_safe_haven/infrastructure/components/wrapped/workspace.py
```diff
@@ -0,0 +1,75 @@
...
+        workspace_name: Input[str],
+    ) -> None:
+        self.location = location
+        self.resource_group_name = resource_group_name
+        self.retention_in_days = retention_in_days
+        self.sku = sku
+        self.workspace_name = workspace_name
+
+
+class LogAnalyticsWorkspace(ComponentResource):
```
JimMadge: So the advantage of making this a `ComponentResource` is that we can create proper `Output`s, rather than just properties?
That seems worthwhile.

Can we then move this to the `composite` module then.

I think the name shouldn't be `workspace` as that clashes with our user workspaces.

JimMadge:
```suggestion
class OperationalInsightsWorkspace(ComponentResource):
```

Given that Azure has updated the name, we probably should too to avoid confusion.

data_safe_haven/infrastructure/programs/sre/apt_proxy_server.py
```diff
@@ -129,10 +129,7 @@ def __init__(
             diagnostics=containerinstance.ContainerGroupDiagnosticsArgs(
                 log_analytics=containerinstance.LogAnalyticsArgs(
                     workspace_id=props.log_analytics_workspace.workspace_id,
-                    workspace_key=operationalinsights.get_shared_keys_output(
-                        resource_group_name=props.log_analytics_workspace.resource_group_name,
-                        workspace_name=props.log_analytics_workspace.name,
-                    ).apply(lambda keys: keys.primary_shared_key),
+                    workspace_key=props.log_analytics_workspace.workspace_key,
```
JimMadge, line 132: Much neater :+1:

JimMadge (changes requested):
Looks great. If I have understood, moving to a `ComponentResource` feels like the right answer as it gives us `Output`s naturally, which play nicely with the rest of the Pulumi :rocket:.

I would want the CI fixes in a different PR, and a little tidying of the package's module structure though.

cptanalatriste:
> #2567

Hi @JimMadge , it actually replaces it: PR 2567 was a failed attempt :(

.github/workflows/build_documentation.yaml
cptanalatriste: Makes sense! Here's the PR:  https://github.com/alan-turing-institute/data-safe-haven/pull/2572

data_safe_haven/infrastructure/components/wrapped/workspace.py
cptanalatriste: Yes, the previous implementation ----just extending a `pulumi-azure-native` type--- was causing problems on update, and I didn't find any reference to it on Pulumi documentation. Using standard component resources seem to solve the issue.

cptanalatriste: Makes sense. Will make the change

JimMadge (approved):
Looks good :+1:

Issue #2568, opened 2026-02-19
Implement automated SRE healthchecks after deployment


cptanalatriste:
## :white_check_mark: Checklist

- [x] I have searched open and closed issues for duplicates.
- [x] This is a request for a new feature in the Data Safe Haven or an upgrade to an existing feature.
- [x] The feature is still missing in the [latest version](https://github.com/alan-turing-institute/data-safe-haven/releases).
- [x] I have read through the [documentation](https://alan-turing-institute.github.io/data-safe-haven/).
- [x] This isn't an open-ended question (open a [discussion](https://github.com/alan-turing-institute/data-safe-haven/discussions) if it is).

## :strawberry: Suggested change

At the moment, checking the health of a deployed TRE requires logging into the SRE and manually run the smoke tests script, or performing a manual validation to cover features outside the smoke tests scope. It would be great if this can be automated.

I was envisioning something like this:

```
dsh sre deploy my-tre --healthcheck
```

To perform the health check right after deploying. Or

```
dsh sre healthcheck my-tre
```

To run the health checks to an already deployed TRE. The checks can include:

- If `cloud-init` successfully finished in all workspaces.
- If the `identity` container is up-and-running. Or any other container.
- If it's possible to `pip install` within a workspace.
- If it's possible to `git clone` within a workspace.
- If the DNS server is up and running.
- If DNS is properly configured in workstations and containers.
- If internet access is actually restricted in T2/T3.
- If the DNS Sidecar is running successfully.

And any other scenario worth checking.

## :steam_locomotive: How could this be done?

As an initial approximation, I would just use Pytest + Azure Python SDK to perform the checks. Or as @JimMadge suggested we can even use Selenium. Ideally, this good be a good start for a full CI/CD pipeline in the near future.

Issue #2557, opened 2026-02-12
Updating Nexus to v3.88.0 breaks the container


cptanalatriste:
## :white_check_mark: Checklist

- [x] I have searched open and closed issues for duplicates.
- [x] This is a problem observed when **deploying** a Data Safe Haven.
- [x] I can reproduce this with the [latest version](https://github.com/alan-turing-institute/data-safe-haven/releases).
- [x] I have read through the [documentation](https://alan-turing-institute.github.io/data-safe-haven/).
- [x] This isn't an open-ended question (open a [discussion](https://github.com/alan-turing-institute/data-safe-haven/discussions) if it is).

## :computer: System information

- **Operating System:**  macOS
- **Data Safe Haven version:**  develop@7cb3e36
 

## :package: Packages

<details>
<summary>List of packages</summary>

```none
Paste list of packages here
```

</details>

## :no_entry_sign: Describe the problem

When using  the image`nexus==3.88.0` (introduced by PR https://github.com/alan-turing-institute/data-safe-haven/pull/2550 ), it is not possible to install Python packages in TREs that rely on Nexus.

Further investigation in the Azure Portal shows that the `nexus` container fails to start, and checking its logs (see below) suggest a problem when connecting to PostgreSQL.

## :deciduous_tree: Log messages

Found in the container logs:

```
Caused by: java.lang.RuntimeException: org.sonatype.nexus.upgrade.datastore.UpgradeException: SQL State : 42704
Message : ERROR: operator class "gist_trgm_ops" does not exist for access method “gist"
```

<details>
<summary>Relevant log messages</summary>

```none
Your log details here
```

</details>

llewelld:
The error is being generated because the Trigram Module is a requirement for 3.88.0, as can be seen in the [Release Notes for 3.88.0](https://help.sonatype.com/en/sonatype-nexus-repository-3-88-0-release-notes.html).

Instructions for how to enable the extension can be found in the [Nexus documentation](https://help.sonatype.com/en/installing-the-trigram-module.html).

However the situation is a little nuanced because we do have the Trigram Module enabled on the database, as can be seen in [software_repository.py](https://github.com/alan-turing-institute/data-safe-haven/blob/212a4b6ea0a62e9fd0b3bc4c63d4f97cf1539c62/data_safe_haven/infrastructure/programs/sre/software_repositories.py#L388).

[...]

Issue #2502, opened 2025-11-25
OAuth client secret starting with dash breaks identity container


cptanalatriste:
## :white_check_mark: Checklist

- [x] I have searched open and closed issues for duplicates.
- [x] This is a problem observed when **deploying** a Data Safe Haven.
- [x] I can reproduce this with the [latest version](https://github.com/alan-turing-institute/data-safe-haven/releases).
- [x] I have read through the [documentation](https://alan-turing-institute.github.io/data-safe-haven/).
- [x] This isn't an open-ended question (open a [discussion](https://github.com/alan-turing-institute/data-safe-haven/discussions) if it is).

## :computer: System information

- **Operating System:** macOS
- **Data Safe Haven version:** v5.6.0

## :package: Packages

<details>
<summary>List of packages</summary>

```none
Paste list of packages here
```

</details>

## :no_entry_sign: Describe the problem

If the OAuth Client Secret passed to `shm-cvdnetdev-sre-per-container-group-identity` happens to start with dash (`-`), the `apricot` container won't start (see Log messages).

<img width="937" height="370" alt="Image" src="https://github.com/user-attachments/assets/f7378033-e731-4383-b98e-f73e5a9cdb94" />

This is because `apricot` uses `argparse` for argument processing, and these kind of arguments is not support (see: https://bugs.python.org/issue9334  and https://github.com/python/cpython/issues/91158)

<img width="303" height="527" alt="Image" src="https://github.com/user-attachments/assets/e4ba13fe-6969-4797-a085-22ec9cf04262" />

## :deciduous_tree: Log messages

<details>
<summary>Relevant log messages</summary>

In container instance messages, you should see:

```none
Apricot: error: argument -s/--client-secret: expected one argument
```

</details>

cptanalatriste:
You can also verify in Entra that the secret will break apricot. As a workaround, you can delete the Client Secret in Entra and do a redeploy: that will force re-creating the secret along the identity container. 

<img width="681" height="652" alt="Image" src="https://github.com/user-attachments/assets/a101c356-e315-4c5d-afab-1c6e190f33a7" />

PR #2570, opened 2026-02-26
Health check on container instances

cptanalatriste:

N/A

* We have added a new CLI command: `dsh sre healthcheck` (see `sre.py`).
* We are also extending the `AzureContainerInstane` type, so it can now return all the containers in the container group (see `azure_container_instance.py`). We'll use this method in assertions (see `healthcheck_utils.py`).
* The health check logic is built using `pytest` (see `sre_healtchcheck_runner.py`). To that end, we have developed a plugin to provide tests with information of the stack, the subscription,  and the SRE config file (see `healthcheck_plugin.py`).
* This first round of tests just check that every SRE container is in running state (see `test_container_instance.py` and `pyproject.toml`). We need to add some logic to handle Nexus containers, since it's not deployed in T0/T1.
* We need to modify the Pulumi code of all the container instance resources, to expose the container group name and resource group (see `infrastructure/programs/sre/*.py`). We use this information to gather container group information.

Contributes towards https://github.com/alan-turing-institute/data-safe-haven/issues/2568

### :microscope: Tests

Tested on a Tier 2 and a Tier 0 SREs (below):

image


JimMadge (commented):
Oh, interesting 🤔.

Where were you seeing this going in the medium/long term?

Some things feel a bit odd,

- using pytest for these checks
- A CLI for this rather than a dashboard

Those things feel a bit more reasonable when we think about scaling this to other tests though. That said, we are probably only using 1% of pytest and it might be simpler to write test class of our own.

When talking about containers, it would be good to add a health check too, on top of checking the containers are running. I'm not sure if all of the containers we use have an appropriate `HEALTHCHECK` defined, but we could add them. Looks like in ACI this is called [liveness probe](https://learn.microsoft.com/en-us/azure/container-instances/container-instances-liveness-probe).

cptanalatriste:
Hi @JimMadge , thanks for your comments. Regarding your questions:

1. In the **short term**, I see the `healthcheck` feature as quick holistic quality check on deployed SRE. For example, the tests contained in this PR would detect if an automatic PR broke a container (like https://github.com/alan-turing-institute/data-safe-haven/issues/2557 ). Or if we're facing an already known container-related issue, like https://github.com/alan-turing-institute/data-safe-haven/issues/2502. I envision this tool will be really helpful for the admin teams working with production TREs (maybe @helen-d-little and @llewelld can comment).
2. This PR is meant to be just a basic scaffolding setup, and further tests/checks are coming. On the **medium term**, I would like to port most of the tests in the [security checklist](https://data-safe-haven.readthedocs.io/en/latest/deployment/security_checklist.html) to the healtcheck CLI. That would also speed-up DSH releases.
3. On the long term, [as discussed before](https://alan-turing-institute.slack.com/archives/C0428BBM94M/p1771493795254339?thread_ts=1771321948.465139&cid=C0428BBM94M), the logic built for this feature can serve as basis for a CI/CD pipeline that deploys automatically to a Development TRE. It can even, for example, run periodically and populate a dashboard. 
4. I also felt weird using `pytest` outside a testing context, but it had so many convenient features that made working on this PR very easy. To mention a few: 1) Using plugins to deliver testing dependencies, 2) Parametrised tests, 3) Configurable output and reporting. I'll argue building those features is not trivial, and I don't see a strong argument for not using a mature library that already offers them.
5. It would definitely be great to implement liveness probes to our container instances! We would need to check  into each image documentation to see what's already on offer, or maybe setup a custom one ourselves. However, at the moment the container state gaves me all the signal I need for the issues I'm aware of (see item 1.). Happy to add this to backlog, and work on it later.

Hopefully this makes sense. Please let me know if you have further questions.

JimMadge:
I think the longer term plan makes sense, but we need to do more now to plan what that will look like.

Container status/healthchecks feels like a sensible place to start. Running pulumi refresh in check mode also feels like an obvious starting point.

I feel pytest is the wrong library to use here. It is really focused on code testing, which isn't what we are doing. Parametrising test functions and formatting the output shouldn't be very difficult, and I think going with pytest we are likely to run in to problems in the future with the way it handles exceptions and reports errors.

I think what we really want is a function with returns a test result (or even just a boolean) and then format sensible output from that.

cptanalatriste:
@JimMadge , I'm still not entirely clear on the concrete technical problems we would encounter in a hypothetical future by using `pytest` (or if they're unavoidable), and I feel what you're suggesting we'll require re-writing/re-implementing mature features already available and battle-tested.

However, this feature is not been progressed for two weeks and I'm feeling some urgency for getting closer to the medium/long term plan. Also, as you mentioned, the `pytest` features I'm using at the moment are not that complicated. So, I'll prepare a `pytest`-free version of this PR shortly. I'll let you know when it's ready.

cptanalatriste:
@JimMadge here's a pytest-free version of this PR. Still working as expected: 
image

data_safe_haven/external/interface/azure_container_instance.py
```diff
@@ -89,6 +92,20 @@ def restart(self, target_ip_address: str | None = None) -> None:
...
+        aci_client = ContainerInstanceManagementClient(
+            self.azure_sdk.credential(), self.azure_sdk.subscription_id
+        )
+
+        container_group: ContainerGroup = aci_client.container_groups.get(
+            resource_group_name=self.resource_group_name,
+            container_group_name=self.container_group_name,
+        )
+
+        return typing.cast(list[Container], container_group.containers)
```
JimMadge, line 107: Is this because `container_group.containers` isn't actually a `list[Container]`?

data_safe_haven/healthcheck/check_container_instance.py
```diff
@@ -0,0 +1,70 @@
+from attrs import define
```
JimMadge, line 1: Any reason to use this over `dataclasses.dataclass`?
`attrs` isn't an explicit dependency, and we are already using dataclasses elsewhere.

data_safe_haven/provisioning/sre_provisioning_manager.py
```diff
@@ -51,9 +51,14 @@ def __init__(
...
-            "software_repositories"
-        )
+        self.software_repository_params: dict[str, str] | None = None
+
+        try:
+            self.software_repository_params = sre_stack.output("software_repositories")
+        except KeyError:
+            self.logger.info(
+                "There's no software repositories information in the current SRE"
+            )
```
JimMadge, line 61: Is this a bug fix, can we put it in a different PR?
We shouldn't use exception handling for flow control if we can avoid it. Hopefully we can do a `if "software_repositories" in ...` here.

data_safe_haven/healthcheck/healthcheck_plugin.py
```diff
@@ -0,0 +1,25 @@
...
+    def project_manager(self) -> SREProjectManager:
+        return self._sre_project_manager
+
+    @property
+    def subscription_name(self) -> AzureSubscriptionName:
+        return self._subscription_name
+
+    @property
+    def sre_config(self) -> SREConfig:
+        return self._sre_config
```
JimMadge, line 25: What is this class for? It is just aliasing three variables.

data_safe_haven/healthcheck/healthcheck_utils.py
```diff
@@ -0,0 +1,66 @@
...
+from .healthcheck_plugin import SREHeathCheckPlugin
+
+
+class HealthCheckError(Exception):
+    pass
+
+
+class HealthCheckTest(Protocol):
+
+    def check(self, plugin: SREHeathCheckPlugin) -> str: ...
```
JimMadge, line 16: I don't like the implicit way classes are assigned. I think it would be better to make an ABC for health checks

data_safe_haven/healthcheck/healthcheck_utils.py
```diff
@@ -0,0 +1,66 @@
...
+
+from data_safe_haven.external import AzureContainerInstance
+from data_safe_haven.infrastructure import SREProjectManager
+from data_safe_haven.types import AzureSubscriptionName
+
+from .healthcheck_plugin import SREHeathCheckPlugin
+
+
+class HealthCheckError(Exception):
+    pass
```
JimMadge, line 11: We collect our exceptions in the `exceptions` module.

data_safe_haven/healthcheck/sre_healthcheck_runner.py
```diff
@@ -0,0 +1,52 @@
...
+from .check_container_instance import (
+    CheckContainerInstance,
+    CheckSoftwareRepositoriesContainer,
+)
+from .healthcheck_plugin import SREHeathCheckPlugin
+from .healthcheck_utils import HealthCheckError, HealthCheckTest
+
+
+@define
+class SREHealthCheckRunner:
```
JimMadge, line 18: Is there a benefit to this being a class? Feels like it is just a function.

data_safe_haven/healthcheck/sre_healthcheck_runner.py
```diff
@@ -0,0 +1,52 @@
...
+                "sre_identity",
+            ]
+        ]
+
+        for health_check in test_classes:
+            try:
+                success_message = health_check.check(health_check_plugin)
+                self._logger.info(f"\u2705 {success_message}")
+            except HealthCheckError as error:
+                self._logger.info(f"\u274c {error.args[0]}")
```
JimMadge, line 52: Rich has support for emoji codes that are more [human readable](https://rich.readthedocs.io/en/stable/markup.html#emoji).

JimMadge (changes requested):
I think this is looking good. I'm finding the structure quite complex and maybe over engineered though.

Couldn't we just have a `HealthCheck` ABC, and sublcass it as we need, maybe like `ContainerHealthCheck(HeathCheck)` `PackageContainerHealthCheck(ContainerHealthCheck)`. Which is, more or less what is here, but I don't like the implicit way `typing.Protocol` works.

We could also make factories, but I think classes make sense because we are likely to want helper methods as well as the `run` method.

For executing, I have used decorators before to add functions to a list. So you could have something like,

```python
health_check_instances = []

class MyCheck(HealthCheck):
    def run(self, ...):
        ...

@health_check_instance
MyCheck(case1)

@health_check_instance
MyCheck(case2)


for instances in health_check_instances:
    result = instance()
    handle_result(result)
```

How does that sound to you?

PR #2601, opened 2026-05-03
Support time-based credential confirmation

### LINKED ISSUE #2525: Drop requirement for repeated confirmation of account when running dsh

<!--
Before reporting a problem please check the following. Replace the empty checkboxes [ ] below with checked ones [x] accordingly.
-->

<!--
What would you like to see changed?
-->

Currently when running `dsh` commands the user is repeatedly requested to re-confirm that the correct Azure user is logged in:

```sh
$ dsh users list bridgeai_dsg_jan26
        user: David Llewellyn-Jones (16aa4559-233f-4726-851f-c17725d29ee3)                                                                                                                                                                      
        tenant: turing.ac.uk (4395f4a7-e455-4f95-8a9f-1fbaef6384f9)                                                                                                                                                                             
You are logged into the Microsoft Graph API as:                                                                                                                                                                                                 
        user: entra.admin.david.llewellyn-jones@turingprodtre.onmicrosoft.com (bf755730-a847-453a-bd47-2ef15aa01f7f)                                                                                                                            
        tenant: turingprodtre.onmicrosoft.com (c2284f3e-f7d9-4716-a0ea-49a502f30783)                                                                                                                                                            
Loaded stack shm-prod5-sre-bridgeai-dsg-jan26.                                                                                                                                                                                                  
[...]
```

It can sometimes take 10 seconds or so before each confirmation appears and the tool is quite slow in general, so this can often mean waiting for 30 seconds before being able to context switch to some other task.

It's not clear there's a real benefit from asking for this confirmation multiple times, except as an artefact of using the underlying Azure tools. It feels like it shouldn't be necessary and the user experience would be improved if it could be dropped.

<!--
Can you suggest how to implement the desired behaviour?
-->

Addressing this would require two changes:
1. Remove the confirmation for Azure CLI, the Graph API and anything else that requests it.
2. Implement a command to log out. For example: `dsh auth drop`. It's possible this could just be left to `az login`.

llewelld:

Ag
N/A.

Adds a local settings file to support per-user configurations.

Adds a setting to allow the credential confirmation to be skipped if it was previously confirmed within a given time period. The period defaults to eight hours but can be set in the configuration file.

Closes #2525

### :microscope: Tests

I chose this because it doesn't require a full deployment for testing. I checked by running the following commands using various configuration options set in the `~/.config/data_safe_haven/local.yaml` file:
```sh
# Clear out the existing settings file to reset to default
rm ~/.config/data_safe_haven/local.yaml
# List users to ensure a config file is written out
dsh users list <sre-name>
# Default period is set to 8 hours, so there should be no confirmation requested on the second run
dsh users list <sre-name>
# Set the timeout period to 10 seconds
sed -ie 's/timeout: [0-9]\+$/timeout: 10/g' ~/.config/data_safe_haven/local.yaml 
# Wait 10 seconds
sleep 10
# Now the timeout period has expired so confirmation should be requested
dsh users list <sre-name>
# Reset the settings back to their default values by removing the settings file
rm ~/.config/data_safe_haven/local.yaml
```
I've also added a couple of unit tests that uses the `freezegun` module to test out whether confirmation is requested or not after various periods of time.

### DIFF
--- a/data_safe_haven/config/config_sections.py
+++ b/data_safe_haven/config/config_sections.py
@@ -141,10 +141,3 @@ class ConfigSectionSRE(BaseModel, validate_assignment=True):
             msg = "When `allow_workspace_internet` is `true`, `software_packages` must be `any`"
             raise ValueError(msg)
         return self
-
-
-class ConfigSectionAccountConfirm(BaseModel, validate_assignment=True):
-    cache: bool = True
-    confirmed: dict[str, int] = {}
-    # Default to 8 hours timeout on credential reconfirmation
-    timeout: int = 8 * 60 * 60
--- a/data_safe_haven/config/config_sections.py
+++ b/data_safe_haven/config/config_sections.py
@@ -141,3 +141,10 @@ class ConfigSectionSRE(BaseModel, validate_assignment=True):
             msg = "When `allow_workspace_internet` is `true`, `software_packages` must be `any`"
             raise ValueError(msg)
         return self
+
+
+class ConfigSectionAccountConfirm(BaseModel, validate_assignment=True):
+    cache: bool = True
+    confirmed: dict[str, int] = {}
+    # Default to 8 hours timeout on credential reconfirmation
+    timeout: int = 8 * 60 * 60
--- a/data_safe_haven/external/api/credentials.py
+++ b/data_safe_haven/external/api/credentials.py
@@ -16,6 +16,8 @@ from azure.identity import (
     TokenCachePersistenceOptions,
 )
 
+# Import full config module to accommodate a circular reference
+import data_safe_haven.config
 from data_safe_haven import console
 from data_safe_haven.directories import config_dir
 from data_safe_haven.exceptions import (
@@ -87,12 +89,21 @@ class DeferredCredential(TokenCredential):
         if (user_id, tenant_id) in DeferredCredential.cache_:
             return True
 
+        localconfig = data_safe_haven.config.LocalConfig.getinstance()
+        skip = localconfig.skip_confirmation(self.name)
+
         DeferredCredential.cache_.add((user_id, tenant_id))
         self.logger.info(f"You are logged into the [blue]{self.name}[/] as:")
         self.logger.info(f"\tuser: [green]{user_name}[/] ({user_id})")
         self.logger.info(f"\ttenant: [green]{tenant_name}[/] ({tenant_id})")
 
-        return console.confirm("Are these details correct?", default_to_yes=True)
+        confirm = skip or console.confirm(
+            "Are these details correct?", default_to_yes=True
+        )
+        if not confirm:
+            localconfig.clear_confirm(self.name)
+        localconfig.write()
+        return confirm
 
     def get_token(
         self,
--- a/data_safe_haven/external/api/credentials.py
+++ b/data_safe_haven/external/api/credentials.py
@@ -89,8 +89,8 @@ class DeferredCredential(TokenCredential):
         if (user_id, tenant_id) in DeferredCredential.cache_:
             return True
 
-        localconfig = data_safe_haven.config.LocalConfig.getinstance()
-        skip = localconfig.skip_confirmation(self.name)
+        localconfig = data_safe_haven.config.LocalConfigManager.getinstance()
+        skip = localconfig.accountconfirm.skip_confirmation(self.name)
 
         DeferredCredential.cache_.add((user_id, tenant_id))
         self.logger.info(f"You are logged into the [blue]{self.name}[/] as:")
@@ -101,7 +101,7 @@ class DeferredCredential(TokenCredential):
             "Are these details correct?", default_to_yes=True
         )
         if not confirm:
-            localconfig.clear_confirm(self.name)
+            localconfig.accountconfirm.clear_confirm(self.name)
         localconfig.write()
         return confirm
 
--- a/data_safe_haven/external/api/credentials.py
+++ b/data_safe_haven/external/api/credentials.py
@@ -90,14 +90,16 @@ class DeferredCredential(TokenCredential):
             return True
 
         localconfig = data_safe_haven.config.LocalConfigManager.getinstance()
-        skip = localconfig.accountconfirm.skip_confirmation(self.name)
+        confirmation_still_active = (
+            localconfig.accountconfirm.confirmation_still_active(self.name)
+        )
 
         DeferredCredential.cache_.add((user_id, tenant_id))
         self.logger.info(f"You are logged into the [blue]{self.name}[/] as:")
         self.logger.info(f"\tuser: [green]{user_name}[/] ({user_id})")
         self.logger.info(f"\ttenant: [green]{tenant_name}[/] ({tenant_id})")
 
-        confirm = skip or console.confirm(
+        confirm = confirmation_still_active or console.confirm(
             "Are these details correct?", default_to_yes=True
         )
         if not confirm:
--- a/data_safe_haven/serialisers/yaml_serialisable_model.py
+++ b/data_safe_haven/serialisers/yaml_serialisable_model.py
@@ -22,12 +22,17 @@ class YAMLSerialisableModel(BaseModel, validate_assignment=True):
     config_type: ClassVar[str] = "YAMLSerialisableModel"
 
     @classmethod
+    def from_filepath_raw(cls: type[T], config_file_path: PathType) -> T:
+        """Construct a YAMLSerialisableModel from a YAML file"""
+        with open(Path(config_file_path), encoding="utf-8") as f_yaml:
+            settings_yaml = f_yaml.read()
+        return cls.from_yaml(settings_yaml)
+
+    @classmethod
     def from_filepath(cls: type[T], config_file_path: PathType) -> T:
         """Construct a YAMLSerialisableModel from a YAML file"""
         try:
-            with open(Path(config_file_path), encoding="utf-8") as f_yaml:
-                settings_yaml = f_yaml.read()
-            return cls.from_yaml(settings_yaml)
+            return cls.from_filepath_raw(config_file_path)
         except FileNotFoundError as exc:
             msg = f"Could not find file {config_file_path}."
             raise DataSafeHavenConfigError(msg) from exc
--- a/tests/external/api/test_credentials.py
+++ b/tests/external/api/test_credentials.py
@@ -6,7 +6,7 @@ from azure.identity import (
 from freezegun import freeze_time
 from pytest_mock import MockerFixture
 
-from data_safe_haven.config import LocalConfig
+from data_safe_haven.config import LocalConfigManager
 from data_safe_haven.directories import config_dir
 from data_safe_haven.exceptions import (
     DataSafeHavenAzureError,
@@ -27,7 +27,7 @@ def example_config_nocache_yaml():
             "  cache: false",
             "  confirmed:",
             "    Test: 0",
-            "  timeout: 7200",
+            "  timeout_in_milliseconds: 7200",
         ]
     )
 
@@ -40,7 +40,7 @@ def example_config_cache_yaml():
             "  cache: true",
             "  confirmed:",
             "    Test: 0",
-            "  timeout: 7200",
+            "  timeout_in_milliseconds: 7200",
         ]
     )
 
@@ -48,12 +48,14 @@ def example_config_cache_yaml():
 @pytest.fixture
 def mock_local_config_path(mocker: MockerFixture, tmp_path) -> None:
     mocker.patch.object(
-        LocalConfig, "default_config_file_path", return_value=tmp_path / "local.yaml"
+        LocalConfigManager,
+        "default_config_file_path",
+        return_value=tmp_path / "local.yaml",
     )
 
 
 def clear_local_config(mock_confirm_yes):
-    LocalConfig._instance = None
+    LocalConfigManager._instance = None
     mock_confirm_yes.reset_mock()
     DeferredCredential.cache_ = set()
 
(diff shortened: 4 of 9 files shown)


data_safe_haven/config/config_sections.py
```diff
@@ -141,3 +141,10 @@ def validate_internet_and_packages(self) -> ConfigSectionSRE:
             msg = "When `allow_workspace_internet` is `true`, `software_packages` must be `any`"
             raise ValueError(msg)
         return self
+
+
+class ConfigSectionAccountConfirm(BaseModel, validate_assignment=True):
+    cache: bool = True
+    confirmed: dict[str, int] = {}
+    # Default to 8 hours timeout on credential reconfirmation
+    timeout: int = 8 * 60 * 60
```
cptanalatriste: Can you add the units in the variable name? i.e `timeout_in_milliseconds`?

data_safe_haven/config/config_sections.py
```diff
@@ -141,3 +141,10 @@ def validate_internet_and_packages(self) -> ConfigSectionSRE:
             msg = "When `allow_workspace_internet` is `true`, `software_packages` must be `any`"
             raise ValueError(msg)
         return self
+
+
+class ConfigSectionAccountConfirm(BaseModel, validate_assignment=True):
+    cache: bool = True
+    confirmed: dict[str, int] = {}
```
cptanalatriste: What's stored in this dictionary?

data_safe_haven/external/api/credentials.py
```diff
@@ -87,12 +89,21 @@ def confirm_credentials_interactive(
...
         DeferredCredential.cache_.add((user_id, tenant_id))
         self.logger.info(f"You are logged into the [blue]{self.name}[/] as:")
         self.logger.info(f"\tuser: [green]{user_name}[/] ({user_id})")
         self.logger.info(f"\ttenant: [green]{tenant_name}[/] ({tenant_id})")
 
-        return console.confirm("Are these details correct?", default_to_yes=True)
+        confirm = skip or console.confirm(
+            "Are these details correct?", default_to_yes=True
+        )
+        if not confirm:
```
cptanalatriste, line 105: So .... let's say I forgot to switch contexts, and I'm seeing the wrong credentials. Does this mean I lose my caching config?

data_safe_haven/config/config_sections.py
```diff
@@ -141,3 +141,10 @@ def validate_internet_and_packages(self) -> ConfigSectionSRE:
             msg = "When `allow_workspace_internet` is `true`, `software_packages` must be `any`"
             raise ValueError(msg)
         return self
+
+
+class ConfigSectionAccountConfirm(BaseModel, validate_assignment=True):
```
cptanalatriste: As discussed in the morning, please consider implementing this by context

data_safe_haven/serialisers/yaml_serialisable_model.py
```diff
@@ -21,13 +21,18 @@ class YAMLSerialisableModel(BaseModel, validate_assignment=True):
...
 
+    @classmethod
+    def from_filepath_raw(cls: type[T], config_file_path: PathType) -> T:
+        """Construct a YAMLSerialisableModel from a YAML file"""
+        with open(Path(config_file_path), encoding="utf-8") as f_yaml:
+            settings_yaml = f_yaml.read()
+        return cls.from_yaml(settings_yaml)
+
     @classmethod
     def from_filepath(cls: type[T], config_file_path: PathType) -> T:
```
cptanalatriste, line 32: Nit: To avoid modifying this file, I would just called `from_filepath()` and when the exception is thrown I would create a new file

data_safe_haven/external/api/credentials.py
cptanalatriste, line 87: It appears there's already a mechanism in place for skipping confirmation. Maybe you can piggybank on it?

cptanalatriste:
Hi @llewelld ! Excellent work, I really see this feature adding value to the development workflow. Besides what we talked today and the comments I dropped in the code, here are some high-level suggestions, mostly related to piggy-backing on existing DSH functionality.

* There's already a way in DSH for handling local config files, like the context YAML file. Ideally, your implementation would follow the same structure. You should have a module for defining it's structure (like `context.py`), a manager file for modifying its content (like `context_manager.py`) , and use the manager module wherever you need to use the config file context (for example, in `sre.py`).
* Also, there seems to be already a mechanism for skipping credential confirmation: the `skip_confirmation` attribute of the `DeferredCredential` class. You can modify it if it doesn't suit your needs, and assign its value according to the YAML file values.

cptanalatriste (changes requested):

data_safe_haven/config/config_sections.py
llewelld: Thanks for this suggestion. I've made this change.

data_safe_haven/config/config_sections.py
llewelld: This stores the last confirmation time for each of the accounts. I've changed the name of the variable and added a comment to try to make this clearer, so it now looks like this:
```python
    # A dictionary for storing the time confirmation was last given
    # Keyed by the credential type (e.g. Azure; Graph API)
    time_confirmed: dict[str, int] = {}
```

data_safe_haven/external/api/credentials.py
llewelld: The cached credentials aren't affected., only the time of the last confirmation is cleared. In effect, this ensures that if the user selects "No; these aren't the correct credentials" then they'll be asked to confirm on the next run as well.

data_safe_haven/config/config_sections.py
llewelld: I thought hard about this and ultimately decided that storing per context is probably overkill and what I've ended up implementing is code that clears the cached confirmation if the user switches context.

This way, immediately after switching context, the user is guaranteed to be asked to check the credentials are the correct ones, which if I recall our discussion is the behaviour we're after.

data_safe_haven/serialisers/yaml_serialisable_model.py
llewelld: This was what I first tried, but it turns out that the `DataSafeHavenConfigError` exception prints an error to the screen, even if it's caught (as a consequence of the error logging on [line 18](https://github.com/alan-turing-institute/data-safe-haven/blob/13d97bf429a70e5c5966b93e826b6e37e9a33448/data_safe_haven/exceptions/__init__.py#L18) of `exceptions/__init__.py`). This isn't great for the user, because it looks like an error has occurred when it hasn't.

I felt removing the error logging from the exception probably isn't what we want, so splitting this function in two to avoid the exception would be the better option. What do you think?

data_safe_haven/external/api/credentials.py
llewelld: Yes you're right, there is. Currently though this mechanism is only used for internal flow (e.g. for disabling confirmation during certain execution paths) and I think overloading it with this new functionality would be liable to cause confusion.

Perhaps it would make sense to change the name of one or other of the functionalities, to make it clearer that they're different?

llewelld:
Thanks for your really helpful suggestions @cptanalatriste. I've made changes to the code to address them, with just a few exceptions, in which case I've explained my reasoning in a comment.

> * There's already a way in DSH for handling local config files, like the context YAML file. Ideally, your implementation would follow the same structure. You should have a module for defining it's structure (like `context.py`), a manager file for modifying its content (like `context_manager.py`) , and use the manager module wherever you need to use the config file context (for example, in `sre.py`).

I've reworked the file structure and class naming to try to better reflect this. There's a nuance, because in practice the `ContextManager` is storing multiple contexts and keeping track of the active one. In contrast we have only a single local config. Conceptually then, the new structure has the `LocalConfigManager` designed to be a base for extensions with different configuration capabilities, with those capabilities managed by distinct classes. I think this is a closer match than my original version.

> * Also, there seems to be already a mechanism for skipping credential confirmation: the `skip_confirmation` attribute of the `DeferredCredential` class. You can modify it if it doesn't suit your needs, and assign its value according to the YAML file values.

I've added a comment elsewhere about this, but from trying to change the code, I'd be inclined to keep these separate. The existing `skip_confirmation` seems to me to mean "should we skip confirmation for this execution flow" whereas the new functionality is "given a bunch of conditions, can we skip confirmation *on this occasion*?".  The two are ORed together to make a final decision. This also relates to an issue raised by @la-sts during the discussion we had (which I'm paraphrasing):

> "Can we do this completely differently? Do we actually need to ask for confirmation at all on most occasions and if not, do we need to cache the confirmation at all?"

I do agree with this, I think we could skip the check entirely for many of the flows where it's currently used (in particular, it probably only needs to be checked when deploying, tearing down or registering users with an SRE). But my feeling is that it'd be better to tackle that separately and that the existing `skip_confirmation` flag would be the way to handle this (i.e. set this flag for more of the execution flows than is currently the case). I'd prefer to make these changes incrementally as it becomes clearer to me where confirmation is needed and where it's not.

data_safe_haven/serialisers/yaml_serialisable_model.py
cptanalatriste: Mmm... I really don't like that behaviour: That makes catching Exceptions pointless. But in any way, I consider that out of scope of this PR

data_safe_haven/external/api/credentials.py
cptanalatriste: Can you elaborate on how they're different? It seems to me that there's a lot of overlap (And if they're indeed for different purposes, different descriptive names would definitely help)

data_safe_haven/serialisers/yaml_serialisable_model.py
llewelld: Thanks Carlos, I agree, it's something that it'd be nice to fix at a more fundamental level. I've created an issue to keep track of it: https://github.com/alan-turing-institute/data-safe-haven/issues/2607

tests/external/api/test_credentials.py
```diff
@@ -48,6 +100,113 @@ def test_confirm_credentials_interactive_fail(
         out, _ = capsys.readouterr()
         assert "Please authenticate with Azure: run 'az login'" in out
 
+    def test_confirm_credentials_interactive_file_cache(
+        self,
+        tmp_path,
+        example_config_cache_yaml,
+        mock_confirm_yes,
+        mock_azureclicredential_get_token,  # noqa: ARG002
```
llewelld, line 108: As we discussed, I double-checked whether this is being used.

It turns out calling `get_credential()` does result in a call to `get_token()`. We can see this on [line 138 of external/api/credentials.py](https://github.com/alan-turing-institute/data-safe-haven/blob/13d97bf429a70e5c5966b93e826b6e37e9a33448/data_safe_haven/external/api/credentials.py#L138).

I was also able to confirm this by switching `mock_azureclicredential_get_token` here and in the test below for `mock_azureclicredential_get_token_invalid`. With this change the tests fail.

tests/external/api/test_credentials.py
```diff
@@ -48,6 +100,113 @@ def test_confirm_credentials_interactive_fail(
...
+        capsys,
+        mock_local_config_path,  # noqa: ARG002
```
llewelld, line 110: I checked this one too, by creating a file that can't be read to or written to by nonroot users:

```sh
$ ls -l $HOME/temp
total 0
-rw------- 1 root root 0 May 28 11:25 local.yaml
```

Then pointing the mocked path to this inaccessible file:

```python
@pytest.fixture
def mock_local_config_path(mocker: MockerFixture, tmp_path) -> None:
    mocker.patch.object(
        LocalConfigManager,
        "default_config_file_path",
        return_value="$HOME/temp" / "local.yaml",
    )
```

With this change the three tests that consume this mock all fail. Without it the user's actual configuration file would be read from/written to.

There may be another way to include these mocks that doesn't require the `noqa` exceptions, in which case I'd be happy to look at implementing it differently.

data_safe_haven/external/api/credentials.py
llewelld: Thanks for flagging this. The two scenarios are:
1. The account is being used mid-flow, for example during the Pulumi deploy, to avoid the user being queried multiple times during the same execution.
2. The user recently confirmed during a previous execution and doesn't want to be hassled again.

The first case currently happens in two places:
1. When `AzureSdk` is used by `SSLCertificateProvider` (see [`ssl_certificate.py`](https://github.com/alan-turing-institute/data-safe-haven/blob/13d97bf429a70e5c5966b93e826b6e37e9a33448/data_safe_haven/infrastructure/components/dynamic/ssl_certificate.py#L66) where `disable_logging` sets the flag).
2. When `AzureSdk` is used by `BlobContainerAclProvider` (see [blob_container_acl.py](https://github.com/alan-turing-institute/data-safe-haven/blob/13d97bf429a70e5c5966b93e826b6e37e9a33448/data_safe_haven/infrastructure/components/dynamic/blob_container_acl.py#L58) where `disable_logging` sets the flag).

Both of these are Pulumi resources. The confirmations are skipped because it doesn't make sense to ask the user midway through a Pulumi deploy, say.

The second case is the subject of this PR. To make the distinction clearer, I've now updated this PR to avoid the `skip_confirmation` wording and changed it to `confirmation_still_active` instead. Hopefully that makes clear that one is preventing the confirmation happening in all situations,whereas the other is just a temporary respite from being hassled.

Hopefully that makes things clearer? What do you think?

llewelld:
Thanks for the additional comments @cptanalatriste and @la-sts. I've made some changes following what we discussed and added comments related to some of the additional queries. I'd be really grateful if you could please help with another review round.

tests/external/api/test_credentials.py
cptanalatriste: The other option is to do the mocking inside each method... but it's not the DSH way.

cptanalatriste (approved):
Looks good to me!

Issue #2610, opened 2026-06-02
Combine Azure Database for PostgreSQL instances


llewelld:
## :white_check_mark: Checklist

- [x] I have searched open and closed issues for duplicates.
- [x] This is a request for a new feature in the Data Safe Haven or an upgrade to an existing feature.
- [x] The feature is still missing in the [latest version](https://github.com/alan-turing-institute/data-safe-haven/releases).
- [x] I have read through the [documentation](https://alan-turing-institute.github.io/data-safe-haven/).
- [x] This isn't an open-ended question (open a [discussion](https://github.com/alan-turing-institute/data-safe-haven/discussions) if it is).

## :strawberry: Suggested change

User story: *"As a project owner running a TRE, I'd like to have as few database instances running as possible, so that I can minimise TRE running consts"*.

Currently there are six Azure Database for PostgreSQL instances:
1. shm-\<name\>-sre-\<name\>-db-server-gitea
2. shm-\<name\>-sre-\<name\>-db-server-gitea-mirror
3. shm-\<name\>-sre-\<name\>-db-server-guacamole
4. shm-\<name\>-sre-\<name\>-db-server-hedgedoc
5. shm-\<name\>-sre-\<name\>-db-server-postgresql
6. shm-\<name\>-sre-\<name\>-db-server-software-repositories

Each costs approximately £1.48 per day to run. Combining these into a single database instance could result in a daily saving of up to £7.40.

<p align="center">
<img width="80%" maxwidth="916" alt="Image" src="https://github.com/user-attachments/assets/b19ca508-fb1d-4a47-b530-e79e3c7ed068"/>
</p>

## :steam_locomotive: How could this be done?

Details copied here from a [different (non-public) ticket](https://github.com/alan-turing-institute/cvd-net/issues/107#issue-4335310784),since I'm planning to continue documenting development on this ticket.

Each is an instance of the `infrastructure.components.composite.PostgresqlDatabaseComponent` class, which [extends](https://www.pulumi.com/docs/iac/concepts/components/) the Pulumi `ComponentResource` class. There are six such classes created in the code during deployment:
1. `SREGiteaServerComponent`: "gitea", infrastructure/programs/sre/gitea_server.py:200
2. `SREGiteaMirrorManagerComponent`: "giteamirror", infrastructure/programs/sre/gitea_mirror_manager.py:165
[...]

PR #2612, opened 2026-06-03
Move multiple databases to single shared server

llewelld:
User story: *"As a project manager running an SRE, I'd like to reduce the number of database servers required, in order to reduce the overall costs of running the SRE.**

This PR includes code changes that move several databases so that they are shared from a single PostgreSQL database server, rather than each having their own individual server.

N/A.

As discussed in Issue #2610, the currently implementation deploys separate Azure PostgreSQL server instances for each database. We therefore have several such servers running:
1. shm-\<name\>-sre-\<name\>-db-server-gitea
2. shm-\<name\>-sre-\<name\>-db-server-gitea-mirror
3. shm-\<name\>-sre-\<name\>-db-server-guacamole
4. shm-\<name\>-sre-\<name\>-db-server-hedgedoc
5. shm-\<name\>-sre-\<name\>-db-server-postgresql
6. shm-\<name\>-sre-\<name\>-db-server-software-repositories

In practice PostgreSQL is capable of running multiple database instances from a single server. Several of these servers are also hosted in the same subnet and so it could make sense to move them to be served from a single PosgreSQL server instance.

This PR therefore moves the following database to be shared from a single server:
1. shm-\<name\>-sre-\<name\>-db-server-gitea
2. shm-\<name\>-sre-\<name\>-db-server-gitea-mirror
3. shm-\<name\>-sre-\<name\>-db-server-hedgedoc

Benefits:
1. Lower costs. This is the main benefit as this change will reduce costs by around £3 per day.
2. Simpler Pulumi deployment with less scope for interacting components to fail.

Drawbacks:
1. Less layering that could have a small impact on security: if the database is compromised it could result in an attacker gaining access to all three databases, rather than just one.
2. No automated migration route.

### :mag: Detail of changes

Although the diff gives the impression of a large change, the difference is actually smaller than it looks for two reasons:
1. There are quite a few changes to the unit tests which can be considered separately from the substantive changes. These have been included in a separate commit.
2. Most of the changes involve moving code rather then creation of new code.

There's also a pattern to the changes. The previous arrangement saw three Pulumi sub components (`SREGiteaServer`, `SREGiteaMirrorManager` and `SREHedgeDocServer`) created inside the `SREUserServicesComponent` component. Each of these three sub components created a `PostgresqlDatabaseComponent` which was directly populated with the component's database, hence we ended up with three databases running on three servers.

This change creates a new `PostgresqlDatabaseComponent` instance called `db_server_shared` owned by the `SREUserServicesComponent` component and which is passed to each of the three sub components as an argument. The three sub components then create the database they require on this shared `PostgresqlDatabaseComponent` resource, rather than creating a new `PostgresqlDatabaseComponent` resource to host it.

We end up with a situation where each sub component installs a database on the shared PostgreSQL database server, rather than them each creating one of their own.

The following discusses the changes made on a file-by-file basis. Where I write of "old lines" and "new lines" I'm referring to line numbers from the diff (depending on whether they relate to the old or new versions of the file).

#### `user_services.py`

At the top of the file we see changes to the `SREUserServicesProps` properties class. Instead of passing in three separate passwords (old lines 38-40) for each of the sub components we instead pass in just a single password (new lines 62-63) for use with the shared database server.

These details are used to create the `db_server_shared` shared database server (new lines 135-151).

Details of the server are then passed to each of the sub components:
1. `SREGiteaServerComponent` Gitea component, new lines 171-174.
2. `SREGiteaMirrorManagerProps` Gitea Mirror component, new lines 199-202.
3. `SREHedgeDocServerComponent` Hedgedoc component, new lines 227-230.

The changes from `user_services.py` cascade into each of the separate components.

#### `gitea_server.py`

In this file we see the details for the `SREGiteaServerComponent` which handles running of the in-SRE Gitea instance. On old lines 30-67 we see the changes to the input parameters (details of the shared database are passed in, rather than details for a new database).

Old lines 198-214 show the switch from creating a `PostgresqlDatabaseComponent` database server to creating a `dbforpostgresql.Database` database on the existing server. This is the substantive change for the component.

The code for deploying the database mirrors code that can be found in the implementation of `PostgresqlDatabaseComponent` on [lines 126-134](https://github.com/alan-turing-institute/data-safe-haven/blob/b884af4a400593ba7cc82607b5806cd9162601aa/data_safe_haven/infrastructure/components/composite/postgresql_database.py#L126) of `postgresql_database.py`.

Finally the changes on old lines 265-277 are to update the environment variables to use the credentials for the shared server for access.

#### `gitea_mirror_manager.py`

The changes in this file mirror those found in the `gitea_server.py` file, the difference being that here they're applied to the `SREGiteaMirrorManagerComponent`. I'll go through them in full, but you'll notice they're very similar to the changes in the `gitea_server.py` file.

Old lines 31-67 show the changes to the input parameters again. These are to ensure the details for the shared database server are used, rather than those needed for creating a new database server.

Old lines 163-179 capture the substantive change. Here a `PostgresqlDatabaseComponent` database server instance is no longer created. Instead a `dbforpostgresql.Database` database is created on the shared PostgreSQL server.

Once again, the changes on old lines 260-272 reflect the updated environment variable values needed to connect to the shared server.

#### `hedgedoc_server.py`

The changes to this file once again reflect a similar pattern. I'll go through them in full, but again, they're very similar to the changes in the previous two files.

Old lines 31-68 once again reflect changes to the input parameters needed to make use of the shared database, rather than those needed for creating a fresh database.

The new input parameters are used on old lines 120-136 where the switch is once again made from deploying a new `PostgresqlDatabaseComponent` database server instances to instead deploy a `dbforpostgresql.Database` database instance on the shared server instead.

Finally on old lines 188-201 the environment variables needed to access the database server are adjusted to reflect the new requirements.

#### `declarative_sre.py`

The `SREUserServices` component is created on old lines 374-409 of this file, inside the `DeclarativeSRE` class. The changes here reflect the input parameter changes needed to create the updated version of the `SREUserServices` component.

These changes are essentially to pass in just a single password for the shared database, rather than passwords for each of the three databases previously. These passwords need to be kept at this level since they end up in the key vault (see the `data.py` details below).

#### `data.py`

These changes update the key vault to reflect the fact we have a single shared database username and password instead of three separate usernames and passwords.

On new lines 279-281 we see the new shared database password being randomly generated.

Old lines 286-328 were for generating and storing the passwords for the three separate database server instances. These are no longer needed.

New lines 288-206 see the new details being stored for the shared server.

Old lines 851-859 which previously captured the passwords for the three servers are now reduced to new lines 809-811 for a single password.

#### `conftest.py` and `test_userservices.py`

See the Tests section below.

1. [Issue #2610](https://github.com/alan-turing-institute/data-safe-haven/issues/2610) details some of the background and motivations for the change.
4. [PR #1](https://github.com/llewelld/data-safe-haven/pull/1) extends this approach to also cover the software repositories database as well. I've separated this out as the software repositories database is in a different subnet, so has a larger impact.

### :microscope: Tests

I've performed several manual deployments of an SRE with these changes. I checked that the databases were still correctly deployed and performed some basic tests to ensure that the services were still working correctly inside the SRE. See my [comment below](https://github.com/alan-turing-institute/data-safe-haven/pull/2612#issuecomment-4634396887) for further details of these checks.

I've also updated the unit tests to perform a mock Pulumi deploy of the `SREUserServicesComponent`. The following covers the details:

#### `conftest;py`

To this file I've added a collection of new fixtures that are needed in order to perform the mock Pulumi deploy, covering the inputs needed to `SREUserServicesComponent`. Inclusion in the `conftest.py` file ensures these will be available for any new tests added in the future.

None of these fixtures are executed inside the `conftest.py` file. Instead they're used in the `test_userservices.py` file.

#### `test_userservices.py`

The majority of this file is for setting up and configuring the mock Pulumi deployment. The code added on new lines 35-71 is needed to work around cases where the mock deploy causes incomplete configuration data that would trigger errors unless patched around. In general the actual values are arbitrary: we just need for them not to be null.

Similarly new lines 74-98 are workarounds to allow the mock deploy to go through. For example the `patch_config()` function is used to ensure `"example.com"` is returned when the `"dnsZone"` configuration value is queried. The value isn't actually used, but if the value returned is null, the mock deploy fails with an error.

New lines 101-106 ensure Pulumi performs a mock deployment, rather than a real deployment.

New lines 109-168 provide the test fixture for the `SREUserServicesComponent` component.

The tests are defined from new line 171 onwards:
1. `test_user_service_props_creation` a simple test to ensure the `SREUserServicesProps` properties can be mock-deployed without error.
2. `test_user_service_creation` a simple test to ensure the `SREUserServicesComponent` component can be mock-deployed without error.
3. `test_shared_db_usernames` tests to ensure the usernames used by the three sub components match the shared database username.
4. `test_shared_db_passwords` tests to ensure the passwords used by the three sub components match the shared database password.
5. `test_shared_db_subnet_gitea` checks that the Gitea component is situated in the same subnet as the shared database.
6. `test_shared_db_subnet_hedgedoc` checks that the Hedgedoc component is situated in the same subnet as the shared database.
7. `test_shared_db_subnet_gitea_mirror` checks that the Gitea Mirror component is situated in the same subnet as the shared database.


llewelld:
After deploying an SRE with this PR integrated, Gitea, Gitea mirroring and Hedgedoc all seem to be working correctly. In order to access the shared database I first had to punch a hole through the firewall in order to allow my local machine to connect. Connecting also required me to enter the `postgresadmin` account password.

Here we see the repository mirrored in the configuration:

<p align="center">
Image
</p>

From the SRE configuration:
```yaml
  gitea_mirror:
    repositories:
    - repository_auth_token: github_pat_XXXXXXXXXXXXXXXXXXXXXXXXX
      repository_name: alanturinginstitute-templates
      repository_url: https://github.com/llewelld/alanturinginstitute-templates
````

The entry showing in the Gitea mirror database:
```text
$ psql -h shm-cvdnetdev-sre-dev04-db-server-shared.postgres.database.azure.com -U postgresadmin -p 5432 -d giteamirror
giteamirror=> select * from mirror;
 id | repo_id |   interval   | enable_prune | updated_unix | next_update_unix | lfs_enabled | lfs_endpoint |                        remote_address                         
----+---------+--------------+--------------+--------------+------------------+-------------+--------------+---------------------------------------------------------------
  1 |       1 | 600000000000 | t            |   1780682959 |       1780683559 | f           |              | https://github.com/llewelld/alanturinginstitute-templates.git
(1 row)
```

The entry showing in the Gitea database:
```text
$ psql -h shm-cvdnetdev-sre-dev04-db-server-shared.postgres.database.azure.com -U postgresadmin -p 5432 -d gitea
gitea=> select * from mirror;
 id | repo_id |   interval   | enable_prune | updated_unix | next_update_unix | lfs_enabled | lfs_endpoint |                                                   remote_address                                                   
----+---------+--------------+--------------+--------------+------------------+-------------+--------------+--------------------------------------------------------------------------------------------------------------------
  1 |       1 | 600000000000 | t            |   1780683076 |       1780683676 | f           |              | http://giteamirror.dev04.cvdnetdev.develop.turingsafehaven.ac.uk:3000/mirroruser/alanturinginstitute-templates.git
(1 row)
```

Wait until the next update is due to ensure it's being written to correctly by the SRE:
```
$ date +%s
1780683477
$ sleep $((1780683559 - $(date +%s)))
$ date +%s
1780684353
$ psql -h shm-cvdnetdev-sre-dev04-db-server-shared.postgres.database.azure.com -U postgresadmin -p 5432 -d giteamirror
giteamirror=> Select * from "mirror";
 id | repo_id |   interval   | enable_prune | updated_unix | next_update_unix | lfs_enabled | lfs_endpoint |                        remote_address                         
----+---------+--------------+--------------+--------------+------------------+-------------+--------------+---------------------------------------------------------------
  1 |       1 | 600000000000 | t            |   1780684159 |       1780684759 | f           |              | https://github.com/llewelld/alanturinginstitute-templates.git
(1 row)

```

For the Hedgedoc service:

<p align="center">
Image
</p>

Checking that this has made it into the database:

```text
$ psql -h shm-cvdnetdev-sre-dev04-db-server-shared.postgres.database.azure.com -U postgresadmin -p 5432 -d hedgedoc
hedgedoc=> Select content from "Notes";
                             content                             
-----------------------------------------------------------------
 
 # Database Testing                                             +
                                                                +
 This is a testing note to see whether it lands in the database.+
 
(2 rows)
```

llewelld:
These changes are extended by https://github.com/llewelld/data-safe-haven/pull/1 . Once this PR has been merged, I'll look further to see whether it's worth exploring extending the approach to cover the nexus software repositories database as well. That lives in a different subnet, so configuring it to use the shared database isn't so straightforward.

llewelld:
There are a couple of errors from the CI pipeline, but these seem unrelated to this PR:
1. Error in ./docs/build/html/overview/using_dsh.html: [404] https://satre-specification.readthedocs.io/en/stable/evaluations/alan_turing_institute.html | Rejected status code (this depends on your "accept" configuration): Not Found
2. Several [MD040](https://github.com/markdownlint/markdownlint/blob/main/docs/RULES.md#md040---fenced-code-blocks-should-have-a-language-specified) errors, meaning "Fenced code blocks should have a language specified", all coming from the `docs` subdirectory.

cptanalatriste:
@llewelld have you tested migrating from the previous Pulumi version to this PR? Pulumi is not that friendly with refactoring, and many times we need to tweak (not trivially) the state to make this possible.

data_safe_haven/infrastructure/programs/sre/gitea_mirror_manager.py
```diff
@@ -43,13 +40,11 @@ def __init__(
         storage_account_name: Input[str],
         workspace_username: str,
         workspace_password: Input[str],
-        database_username: Input[str] | None = None,
+        db_server_shared: Input[PostgresqlDatabaseComponent],
```
cptanalatriste: Nit: Is DSH convention that parameter/attributes are sorted alphabetically

data_safe_haven/infrastructure/programs/sre/gitea_mirror_manager.py
```diff
@@ -160,22 +158,17 @@ def __init__(
...
-            PostgresqlDatabaseProps(
-                database_names=[db_gitea_repository_name],
-                database_password=props.database_password,
-                database_resource_group_name=props.resource_group_name,
-                database_server_name=f"{stack_name}-db-server-gitea-mirror",
-                database_subnet_id=props.database_subnet_id,
-                database_username=props.database_username,
-                disable_secure_transport=False,
-                location=props.location,
+        dbforpostgresql.Database(
```
cptanalatriste: This is really cool! It's very nice that you can do this directly with Pulumi, without DSH "glue code"

data_safe_haven/infrastructure/programs/sre/gitea_server.py
```diff
@@ -42,15 +39,12 @@ def __init__(
         sre_fqdn: Input[str],
         storage_account_key: Input[str],
         storage_account_name: Input[str],
-        database_username: Input[str] | None = None,
+        db_server_shared: Input[PostgresqlDatabaseComponent],
+        db_server_shared_username: Input[str],
+        db_server_shared_password: Input[str],
+        db_server_shared_resource_group_name: Input[str],
```
cptanalatriste: Nit:  You can get the resource group name directly from the PostgresqlDatabaseComponent, without needing to pass it as a pareameter

data_safe_haven/infrastructure/programs/sre/gitea_server.py
cptanalatriste, line 383: @llewelld is it possible the container doesn't wait for the DB to be provisioned to start?  Maybe we can check the resource dependency tree (I think we can do this without deployment) and if there's no dependency, we can make it explicit here.

data_safe_haven/infrastructure/programs/sre/user_services.py
```diff
@@ -129,14 +132,29 @@ def __init__(
...
 
+        # Deploy the shared PostgreSQL database
+        self.db_server_shared = PostgresqlDatabaseComponent(
+            f"{self._name}_db_server_shared",
+            PostgresqlDatabaseProps(
+                database_names=[],
+                database_password=props.db_server_shared_password,
+                database_resource_group_name=props.resource_group_name,
+                database_server_name=f"{stack_name}-db-server-shared",
+                database_subnet_id=props.subnet_containers_support_id,
```
cptanalatriste, line 142: Nit: All the attributes have the `db-server-shared` preffix/suffix, except the `database_subnet_id ` (and am I correct to assume it only contains this DB now?).  But beware! Refactoring/changing the subnet structure can be problementic, specially with TRE migrations...

tests/infrastructure/programs/sre/conftest.py
```diff
@@ -1,20 +1,54 @@
...
+
+    def new_resource(
+        self, args: pulumi.runtime.MockResourceArgs
+    ) -> tuple[str | None, dict[Any, Any]]:
+        resources = (args.name + "_id", args.inputs)
         return resources
 
-    def call(self, _: pulumi.runtime.MockCallArgs):
-        return {}
+    def call(
```
cptanalatriste, line 48: Nice! Most of our tests lack type annotations

tests/infrastructure/programs/sre/test_userservices.py
```diff
@@ -0,0 +1,295 @@
...
+    SREUserServicesComponent,
+    SREUserServicesProps,
+)
+from data_safe_haven.types import (
+    DatabaseSystem,
+    SoftwarePackageCategory,
+)
+
+
+class DataSafeHavenMocks(pulumi.runtime.Mocks):
```
cptanalatriste: Nit: Maybe you can rename this? It clashes with the type in `conftest.py`

tests/infrastructure/programs/sre/test_userservices.py
```diff
@@ -0,0 +1,295 @@
...
+
+
+# The test suite
+class TestSREUserServicesProps:
+    @pulumi.runtime.test  # type: ignore
+    def test_user_service_props_creation(
+        self, user_services_props: SREUserServicesProps
+    ) -> None:
+        """Basic test to ensure properties are being created correctly"""
+        assert isinstance(user_services_props, SREUserServicesProps)
```
cptanalatriste: Not sure I see the purpose of this test... are you checking mock creation is done properly?

tests/infrastructure/programs/sre/test_userservices.py
```diff
@@ -0,0 +1,295 @@
...
+
+    @pulumi.runtime.test  # type: ignore
+    def test_user_service_creation(
+        self, user_services_component: SREUserServicesComponent
+    ) -> None:
+        """Basic test to ensure components are being created correctly"""
+        assert isinstance(user_services_component, SREUserServicesComponent)
```
cptanalatriste, line 178: Idem

cptanalatriste (changes requested):
Great work @llewelld , specially with the test suite, our codebase needs more of those :) I left some minor comments/suggestions, but my main concern/question is how this changes affect migrating a TRE to a new version. Can you please have a look?

data_safe_haven/infrastructure/programs/sre/gitea_mirror_manager.py
llewelld: I've moved the `db_server_shared_password` password parameter up to sort it alphabetically, but because the `db_server_shared_username` has a default value it will need to stay at the end.

I'm happy for it to go like this, but it does mean separating the two associated variables. Is that okay?

data_safe_haven/infrastructure/programs/sre/gitea_server.py
llewelld: I can get the username this way, but I'm not sure about the resource group name. I've made an adjustment so that now the resource group is stored explicitly when the database server resource is created. Is this what you were thinking or is there a different route to me extracting the resource group name?

data_safe_haven/infrastructure/programs/sre/user_services.py
llewelld: Yes, you're correct that it now only contains the single database, but I was thinking it might make sense for other shared support resources to live there in the future. Do you think I should rename it (to `db_server_shared_subnet`, say)? My general feeling is that it may be better to leave it as it is.

tests/infrastructure/programs/sre/test_userservices.py
llewelld: Thanks for flagging this. I've renamed it.

tests/infrastructure/programs/sre/test_userservices.py
llewelld: These tests essentially perform a check that the mock deployment completes successfully; all of the fixtures are set up correctly; etc. I've generally found these quite useful during development, because I can just run them to test small changes.

data_safe_haven/infrastructure/programs/sre/gitea_server.py
llewelld: Oh, this is a really good point. Nicely spotted; I'll check.

data_safe_haven/infrastructure/programs/sre/gitea_mirror_manager.py
llewelld: Discussed this during an over-the-shoulder review and @cptanalatriste confirmed that this (keep optional variables at the end even if it means splitting conceptual blocks) is the correct DSH approach; so I'll mark this conversation as resolved.

data_safe_haven/infrastructure/programs/sre/gitea_mirror_manager.py
llewelld: Thanks; and thanks also for clarifying your thoughts on this during the over-the-shoulder review. I'll mark this conversation as resolved.

data_safe_haven/infrastructure/programs/sre/gitea_server.py
llewelld: I've looked in to this further and generated some dependency graphs. Pulumi had correctly inferred that the container groups were dependent on the database server (through the `containers` parameter, which references the shared database server). The database instances were also (unexceptionally) dependent on the database server. However Pulumi hadn't identified dependences between the deployed databases and the container groups.

Here's the dependency diagram generated from the Pulumi graph. For ref:
1. Green lines: "is a dependency of" (line labels indicate the reason for the dependency).
2. Brown lines: "is a child of".

Dependency graph for Pulumi prior to adding explicit dependency connections.

Deployments were going through fine, but as you rightly say, this could leave the possibility of the containers being deployed before the databases, potentially leaving the containers in a non-working state. I've therefore added explicit dependencies in the code between the containers and the databases. Here's the updated dependency graph from Pulumi:

dev04-db-20260624

This looks better; right? Can you think of any other dependencies that might need to be made explicit here?

tests/infrastructure/programs/sre/test_userservices.py
llewelld: Discussed during the over-the-shoulder review; marking conversation as resolved.

tests/infrastructure/programs/sre/conftest.py
llewelld: Thanks :smile:

data_safe_haven/infrastructure/programs/sre/gitea_server.py
llewelld: After our discussion about this yesterday I dug into it a bit further. You're correct that, [according to the documentation](https://www.pulumi.com/registry/packages/azure-native/api-docs/dbforpostgresql/server/#outputs), this Input parameter should be mirrored as an Output. And the `db_server` instance *does* expose a `resource_group_name` member. But for reasons that are unclear, the Output is producing the value `None` even after the future resolves.

Here's the test code, added to `SREGiteaServerComponent`:
```python
props.db_server_shared.db_server.resource_group_name.apply(
    lambda val: print("Resource Group Name: {}".format(val))
)
print("DB Server variables: {}".format(vars(props.db_server_shared.db_server)))
```
And the resulting output (abridged and formatted for clarity):
```yaml
DB Server variables: {
    '_transformations': [],
    '_name': 'sre_user_services_db_server_shared_server',
    '_type': 'azure-native:dbforpostgresql:Server',
[...]
    'resource_group_name': <pulumi.output.Output object at 0x7f8dfbdc2090>,
    'server_name': <pulumi.output.Output object at 0x7f8dfbdc2180>,
[...]
}

Resource Group Name: None
```
Given this, I've left things with the `database_resource_group_name` being explicitly captured by the `PostgresqlDatabaseComponent` class, with this then used for creating the `dbforpostgresql.Database` instances.

tests/infrastructure/programs/sre/test_userservices.py
llewelld: Following our discussion I've removed this test; marking as resolved!

tests/infrastructure/programs/sre/test_userservices.py
llewelld: As discussed yesterday, I've left this in as I do think it's performing a useful check with the Mock deploy. Technically the other tests rely on a successful deploy as well, but I think it makes sense to separate it out explicitly.

llewelld:
> [M]y main concern/question is how this changes affect migrating a TRE to a new version. Can you please have a look?

Thanks for flagging this. I looked into it and you're correct: Pulumi failed during an upgrade from previous versions. I've therefore made further changes to try to address this, which you can see in commit [a3598e845642e](https://github.com/alan-turing-institute/data-safe-haven/pull/2612/changes/a3598e845642ebef4df1d662467010ded4a9e444) (now part of this pull request).

The main failure that prevents the upgrade is that the upgrade involves adding the new "Shared" database into the `UserServicesContainersSupportSubnet`. Pulumi deploys this with a /29 netmask, meaning 8 available addresses. Five of these are in use by Azure and three are in use by the SRE, leaving no space for the new database to be deployed.

Pulumi can't grow the subnet (it doesn't know how to, even if it were possible) and so the deployment fails.

I tried several tactics to get around this (e.g. renaming the subnet to force a redeployment), but none succeeded. Setting up a dependency (e.g. using a [replaceWith](https://www.pulumi.com/docs/iac/concepts/resources/options/replacewith/) relationship) also failed (it looks like there may be a way to get this to work, but it would have to be done in two stages, which doesn't fit the current flow). Seemingly the reason is that Azure deploys a NIC with each database and these NICs must all be deleted before the subnet can be deleted. I couldn't see a way to force Pulumi to respect this without performing two separate deploys.

The solution I've come up with is to add an `Upgrade` process which allows targetted interventions on specific upgrade paths in advance of the Pulumi deploy. This is handled by the `Upgrade` class in `upgrade.py` and called during the deployment flow in `project_manager.py`.

The upgrade process introduces a general framework for handling upgrades. It checks whether there's a mismatch between the DSH and SRE versions and then:
1. If they're the same, just run a normal deployment.
2. If the DSH version is lower than the SRE version, abort.
3. If the DSH version is higher than the SRE version, check for any special upgrade requirements.

In the case of 3 a warning is presented to the user, explaining the consequence of performing the upgrade (if any) and asking for consent.

In terms of implementation, this is split into two pieces:
1. `upgrade.can_proceed()` is called to check whether it's okay to proceed with the upgrade. In some flows, this involves confirmation from the user.
2. `upgrade.prepare()` triggers the calls to other methods for targetted changes for specific version changes.

Currently there's a trigger for upgrades from anything below 5.7.2 to anything above 5.7.2 (which I'm expecting will be the next release). In this case the user is warned about database deletion:

```md
You're using DSH version 5.7.2 but your SRE was deployed with version 5.7.1. Deployment will
therefore trigger an upgrade.
Upgrading to version 5.7.2 requires reprovisioning of several databases. All of the data on
the following databases will therefore be lost during the upgrade:
1. Gitea
2. Gitea mirror
3. Hedgedoc
If you have non-duplicate data stored on any of these you should back them up before proceeding.
Are you sure you which to proceed with the upgrade? [y/n] (n):
```
(we could also consider adding in a comment about the datadrive here as well, following the merging of #2615).

In this case the upgrade process performs the following steps:
1. Checks whether the NICs and subnet exist.
2. If any of the three NICs that block removal of the subnet exist, delete them. 
3. If the subnet exists, deletes it.

These steps are all performed using the Azure API; there are some new helper functions in `azure_sdk.py` to support this.

If any of the resources have already been removed, the upgrade continues.

My hope is that this upgrade code may be useful for future upgrades as well.

As an aside, I've been performing tests using Pulumi 2.32.0 and haven't experienced any regressions, so have also included this change (but I can pull that into a separate PR if you prefer?).

There are currently no unit tests for this upgrade code. If it's agreed to go down this route, then I can add some, but I didn't want to spend time on them prematurely.

data_safe_haven/external/api/azure_sdk.py
```diff
@@ -542,6 +545,98 @@ def ensure_managed_identity(
...
+            self.credential(), self.subscription_id
+        ) as resource_client:
+            if len(resource_ids) > 0:
+                self.logger.info(f"Deleting {len(resource_ids)} resources:")
+                for index, resource_id in enumerate(resource_ids):
+                    self.logger.info(f"  {index + 1}. {resource_id}")
+            else:
+                self.logger.info("No resources to delete")
+
+            # Unfortunately Azure doesn't like parallel requests, generating
```
cptanalatriste, line 631: I wonder how Pulumi accomplishes this... it launches several processes when deploying TREs...

data_safe_haven/infrastructure/project_manager.py
```diff
@@ -256,10 +261,11 @@ def deploy(
             if force:
                 self.cancel()
             self.refresh(run_program)
+            self.upgrade(run_program=run_program)
             self.preview(disable_diff)
             self.update()
         except Exception as exc:
-            msg = "Pulumi deployment failed."
+            msg = f"Pulumi deployment failed: {exc}"
```
cptanalatriste, line 268: I like this! We need better error handling

cptanalatriste:
> > [M]y main concern/question is how this changes affect migrating a TRE to a new version. Can you please have a look?
> 
> Thanks for flagging this. I looked into it and you're correct: Pulumi failed during an upgrade from previous versions. I've therefore made further changes to try to address this, which you can see in commit [a3598e845642e](https://github.com/alan-turing-institute/data-safe-haven/pull/2612/changes/a3598e845642ebef4df1d662467010ded4a9e444) (now part of this pull request).
> 
> The main failure that prevents the upgrade is that the upgrade involves adding the new "Shared" database into the `UserServicesContainersSupportSubnet`. Pulumi deploys this with a /29 netmask, meaning 8 available addresses. Five of these are in use by Azure and three are in use by the SRE, leaving no space for the new database to be deployed.
> 
> Pulumi can't grow the subnet (it doesn't know how to, even if it were possible) and so the deployment fails.
> 
> I tried several tactics to get around this (e.g. renaming the subnet to force a redeployment), but none succeeded. Setting up a dependency (e.g. using a [replaceWith](https://www.pulumi.com/docs/iac/concepts/resources/options/replacewith/) relationship) also failed (it looks like there may be a way to get this to work, but it would have to be done in two stages, which doesn't fit the current flow). Seemingly the reason is that Azure deploys a NIC with each database and these NICs must all be deleted before the subnet can be deleted. I couldn't see a way to force Pulumi to respect this without performing two separate deploys.
> 
> The solution I've come up with is to add an `Upgrade` process which allows targetted interventions on specific upgrade paths in advance of the Pulumi deploy. This is handled by the `Upgrade` class in `upgrade.py` and called during the deployment flow in `project_manager.py`.
> 
> The upgrade process introduces a general framework for handling upgrades. It checks whether there's a mismatch between the DSH and SRE versions and then:
> 
> 1. If they're the same, just run a normal deployment.
> 2. If the DSH version is lower than the SRE version, abort.
> 3. If the DSH version is higher than the SRE version, check for any special upgrade requirements.
> 
> In the case of 3 a warning is presented to the user, explaining the consequence of performing the upgrade (if any) and asking for consent.
> 
> In terms of implementation, this is split into two pieces:
> 
> 1. `upgrade.can_proceed()` is called to check whether it's okay to proceed with the upgrade. In some flows, this involves confirmation from the user.
> 2. `upgrade.prepare()` triggers the calls to other methods for targetted changes for specific version changes.
> 
> Currently there's a trigger for upgrades from anything below 5.7.2 to anything above 5.7.2 (which I'm expecting will be the next release). In this case the user is warned about database deletion:
> 
> ```
> You're using DSH version 5.7.2 but your SRE was deployed with version 5.7.1. Deployment will
> therefore trigger an upgrade.
> Upgrading to version 5.7.2 requires reprovisioning of several databases. All of the data on
> the following databases will therefore be lost during the upgrade:
> 1. Gitea
> 2. Gitea mirror
> 3. Hedgedoc
> If you have non-duplicate data stored on any of these you should back them up before proceeding.
> Are you sure you which to proceed with the upgrade? [y/n] (n):
> ```
> 
> (we could also consider adding in a comment about the datadrive here as well, following the merging of #2615).
> 
> In this case the upgrade process performs the following steps:
> 
> 1. Checks whether the NICs and subnet exist.
> 2. If any of the three NICs that block removal of the subnet exist, delete them.
> 3. If the subnet exists, deletes it.
> 
> These steps are all performed using the Azure API; there are some new helper functions in `azure_sdk.py` to support this.
> 
> If any of the resources have already been removed, the upgrade continues.
> 
> My hope is that this upgrade code may be useful for future upgrades as well.
> 
> As an aside, I've been performing tests using Pulumi 2.32.0 and haven't experienced any regressions, so have also included this change (but I can pull that into a separate PR if you prefer?).
> 
> There are currently no unit tests for this upgrade code. If it's agreed to go down this route, then I can add some, but I didn't want to spend time on them prematurely.

This al seem very useful!  I do have a minor question. For some old versions of DSH, there won't be tags specifying the version. What's the behaviour in those cases?

data_safe_haven/upgrade/upgrade.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+class UpgradeAbortedError(Exception):
+    def __init__(self, message: str | None = None) -> None:
+        if not message:
+            message = "Deployment aborted."
+        super().__init__(message)
+
+
+class Version:
```
cptanalatriste: I find it hard to believe that we have nothing similar in `pydantic` ... but a quick Google search produced nothing :/

data_safe_haven/upgrade/upgrade.py
```diff
@@ -0,0 +1,213 @@
...
+    major = 0
+    minor = 0
+    patch = 0
+
+    def __init__(self, version: str, *args: Any, **kwargs: Any):
```
cptanalatriste: FYI, we have some version comparison  logic at `.github/scripts/update_docker_versions.py`. But it doesn't seem that portable

data_safe_haven/upgrade/upgrade.py
```diff
@@ -0,0 +1,213 @@
...
+                self.logger.info(
+                    f"You're using DSH version {self.dsh_version} but your SRE was deployed with a more recent version, {self.sre_version}. Deployment will be aborted."
+                )
+                self.proceed = False
+            elif self.dsh_version > self.sre_version:
+                self.logger.info(
+                    f"You're using DSH version {self.dsh_version} but your SRE was deployed with version {self.sre_version}. Deployment will therefore trigger an upgrade."
+                )
+                if self.dsh_version == Version("5.7.2"):
+                    self.logger.info(
```
cptanalatriste, line 80: I think this is fine for this initial attempt, but maybe in following changes we can make this a bit more robust. Ideally, if you want to develop a custom update process you should just: 1) Create a new Python Module, 2) Optional: Edit a config file, without touching much of the existing code.

data_safe_haven/upgrade/upgrade.py
```diff
@@ -0,0 +1,213 @@
...
+    def prepare(self) -> bool:
+        changes = False
+
+        if not self.proceed:
+            raise UpgradeFailedError
+
+        if self.fresh_deployment:
+            changes = False
+        elif self.dsh_version > self.sre_version:
+            if self.dsh_version == Version("5.7.2"):
```
cptanalatriste: Idem here: Ideally, all of the logic of a custom update should be in the same symbol (module, Class, method?). But that's just nitpicking :)

data_safe_haven/upgrade/upgrade.py
```diff
@@ -0,0 +1,213 @@
...
+                changes = self.upgrade_to_5_7_2()
+
+        return changes
+
+    def upgrade_to_5_7_2(self) -> bool:
```
cptanalatriste, line 113: Idem: I envision this as an "Update Manager" that coordinate specific updates modules. But that's for later

cptanalatriste: More wishful thinking, way beyond the scope of this PR: We can extend the DSH CLI to support custom update module execution, to fix pulumi state when things go awry, instead of dealing with portal/azure CLI. And we can have this "Fix" module in source control, for audit purposes

data_safe_haven/external/api/azure_sdk.py
llewelld: I think the problem that I experienced may have been a very specific one: during testing I was removing multiple private endpoints attached to the same resource; when attempting to do this in parallel the Azure SDK generated an error.

It's possible this wouldn't happen in practice and so parallel execution would be safe here, but given that this is a one-off execution deleting only a few resources, it didn't feel worth taking the risk (or the time to investigate in depth).

data_safe_haven/infrastructure/project_manager.py
llewelld: Thanks :smile: It's definitely easier during development to immediately know the reason for the failure.

data_safe_haven/upgrade/upgrade.py
llewelld: Thanks for highlighting this. Following our discussion last week I've switched to using `packaging.version.Version` instead, which will be more flexible in how it interprets version numbers.

data_safe_haven/upgrade/upgrade.py
llewelld: There you use `packaging.version.Version`, which is a nice approach and I've changed the code here to do the same.

data_safe_haven/upgrade/upgrade.py
llewelld: Yes, I totally agree and I like the approach you're suggesting here.

As you know, the real change in this PR is the database consolidation, so I wanted to get this through. But putting in place some more robust upgrade infrastructure feels like an important goal.

data_safe_haven/upgrade/upgrade.py
llewelld: Yes, I agree. Following our discussion last week, I agree we should think about this, but that it's best left for a future pull request.

data_safe_haven/upgrade/upgrade.py
llewelld: Yeah, this would be really great and a big help form the perspective of someone managing a DSH.

llewelld:
> > > [M]y main concern/question is how this changes affect migrating a TRE to a new version. Can you please have a look?

[snip]
> > There are currently no unit tests for this upgrade code. If it's agreed to go down this route, then I can add some, but I didn't want to spend time on them prematurely.
> 
> This al seem very useful! I do have a minor question. For some old versions of DSH, there won't be tags specifying the version. What's the behaviour in those cases?

I hadn't appreciated that there may be versions of the DSH deployed that don't have a `version` tag. Upgrades would have failed on these deployments, so that's helpful to know.

To address this I've amended the code so that if there's no `version` tag a version number of `0.0.0` is assumed. I think this is fine, given that at the moment the only upgrade step where there's a change in behaviour is from below `5.7.2` to `5.7.2` or above. So anything without a version tag using this version of `dsh` should have the upgrade applied.

I've also now added unit tests for the code and have rebased on latest develop. Thanks for all of your help with and input into this @cptanalatriste.

In case it's helpful to know, most of the changes in the latest batch can be found in the following files, all for the purposes of adding upgrade-related unit tests:
1. `tests/conftest.py`
2. `tests/upgrade/test_upgrade.py`
3. `tests/infrastructure/test_project_manager.py`
4. `tests/external/api/azure_sdk.py`

Plus the removal of the3 `Version` class from `data_safe_haven/upgrade/Upgrade.py`.

data_safe_haven/external/api/azure_sdk.py
```diff
@@ -542,6 +545,101 @@ def ensure_managed_identity(
...
+                rg
+                for rg in cast(
+                    list[ResourceGroup], resource_client.resource_groups.list()
+                )
+                if rg.name == resource_group_name
+            ]
+            try:
+                version = str(resource_groups[0].tags["version"])
+            except KeyError:
+                version = "0.0.0"
```
llewelld, line 567: This comes into effect if there's no `version` tag on the SRE resource group.

data_safe_haven/infrastructure/programs/sre/user_services.py
cptanalatriste: If it's easier, leave it as it is

tests/external/api/test_azure_sdk.py
```diff
@@ -404,3 +405,76 @@ def test_storage_exists(
         sdk = AzureSdk("subscription name")
 
         assert sdk.storage_exists(storage_account_name) == exists
+
+    # Check that the AzureSdk code for getting a resource is acting sensibly
```
cptanalatriste: Nit: Since this is a method-level comment, maybe it's better to add it as a docstring

tests/external/api/test_azure_sdk.py
```diff
@@ -404,3 +405,76 @@ def test_storage_exists(
...
+        resource_type = "ResourceType"
+        resource_name = "ResourceName"
+        resource_id = f"/subscriptions/{subscription_id}/resourceGroups/{resource_group}/providers/{provider_namespace}/{resource_type}/{resource_name}"
+        azure_sdk = AzureSdk(sre_project_manager.context.subscription_name)
+        resource = azure_sdk.get_resource(
+            resource_group, provider_namespace, resource_type, resource_name
+        )
+        assert resource.id == resource_id
+
+    # Check that the AzureSdk code for getting a private endpoint is acting sensibly
```
cptanalatriste: Idem

tests/external/api/test_azure_sdk.py
```diff
@@ -404,3 +405,76 @@ def test_storage_exists(
...
+        resource_group = sre_project_manager.output("sre_resource_group")
+        provider_namespace = "Microsoft.Network"
+        resource_type = "privateEndpoints"
+        resource_name = "endpoint"
+        resource_id = f"/subscriptions/{subscription_id}/resourceGroups/{resource_group}/providers/{provider_namespace}/{resource_type}/{resource_name}"
+        azure_sdk = AzureSdk(sre_project_manager.context.subscription_name)
+        resource = azure_sdk.get_private_endpoint(resource_group, resource_name)
+        assert resource.id == resource_id
+
+    # Check that the AzureSdk code for getting a subnet is acting sensibly
```
cptanalatriste: Idem... you know what I mean :)

tests/infrastructure/programs/sre/conftest.py
```diff
@@ -182,3 +251,123 @@ def subnet_dns_sidecar() -> network.GetSubnetResult:
...
+def dockerhub_credentials() -> DockerHubCredentials:
+    return DockerHubCredentials(
+        access_token="docker_token",
+        server="docker_server",
+        username="docker_username",
+    )
+
+
+@fixture
+def ldap_user_filter(ldap_group_search_base: str) -> str:
```
cptanalatriste, line 266: Are these tests related to the current PR?

tests/infrastructure/programs/sre/test_userservices.py
```diff
@@ -0,0 +1,288 @@
...
+            ]
+        ).apply(check)
+
+    @pulumi.runtime.test  # type: ignore
+    def test_shared_db_passwords(
+        self,
+        user_services_component: SREUserServicesProps,
+        user_services_props: SREUserServicesProps,
+    ) -> None:
+        """Check that the password for the shared database is shared correctly across components"""
```
cptanalatriste, line 213: I see you're using docstrings here. Maybe it's better to use them in all the tests/fixtures

tests/upgrade/test_upgrade.py
```diff
@@ -0,0 +1,227 @@
...
+
+    # Check that downgrading is not allowed
+    def test_user_checks_sre_older_deny(
+        self,
+        sre_project_manager: SREProjectManager,
+        mock_sre_project_manager_output: None,  # noqa: ARG002
+        mock_confirm_no: None,  # noqa: ARG002
+        capsys: CaptureFixture[str],
+    ) -> None:
+        with mock.patch.object(AzureSdk, "get_version", return_value="5.7.0"):
```
cptanalatriste, line 73: Nit: Food for thought: We might need to implement downgrading in case we REALLY mess up some of our releases. But that's definitely outside the scope of the current PR

cptanalatriste (approved):
Looks good to me!

tests/external/api/test_azure_sdk.py
llewelld: Yes, thanks for flagging this. I've updated all of the strings to docstring format.

tests/external/api/test_azure_sdk.py
llewelld: Similarly.

tests/external/api/test_azure_sdk.py
llewelld: I do :smile: I've updated to use docstring format.

tests/infrastructure/programs/sre/conftest.py
llewelld: Yes, they look unrelated but all of these new fixtures are needed for the Pulumi stack tests. They fill in blanks that otherwise would fail on account of the fact it's not a real deployment.

tests/infrastructure/programs/sre/test_userservices.py
llewelld: Yes; I've changed them all. Thanks!

tests/upgrade/test_upgrade.py
llewelld: Noted! All create a ticket about upgrades/downgrades to try to capture this.

PR #2644, opened 2026-08-14
Upgrade cvdupgrade to 1.2.0

llewelld:
## :white_check_mark: Checklist

- [x] You have given your pull request a meaningful title (_e.g._ `Enable foobar integration` rather than `515 foobar`).
- [x] You are targeting the appropriate branch. If you're not certain which one this is, it should be **`develop`**.
- [x] Your branch is up-to-date with the **target branch** (it probably was when you started, but it may have changed since then).

### :vertical_traffic_light: Depends on

Depends on PRs #2642 and #2643 which should be merged first.

### :arrow_heading_up: Summary

The main change here is to upgrade cvdupgrade to version 1.2.0. The older version we were using no longer works on systems that are FIPS compliant, because it relies on MD5 checksums, which are no longer considered to be secure.

The source of the problem is that the clamav-mirror Docker image we use to run cvdupgrade is no longer being maintained and the cvdupdate version it contains has therefore been stuck on version 0.3.0 for several years.

This change therefore relies on a forked version of the Docker image that we'll need to maintain (it's very lightweight), but which now contains the updated version of the software we need.

See the forked image repository (and in particular the 5.8.0 releas) for more details:

https://github.com/alan-turing-institute/docker-clamav-mirror

### :closed_umbrella: Related issues

Fixes #2620.
Contributes to #2622 (Release 5.8.0).
PR #2624 build on these changes.

### :microscope: Tests

Combined with the changes in Pull Requests #2642, #2643 and #2624 I've successfully performed several deployments of SREs at Tier 2 and Tier 3. With the combined changes the unit tests all pass.

PR #2643, opened 2026-08-14
Upgrade Nexus container to version 3.94.1

llewelld:
## :white_check_mark: Checklist

- [x] You have given your pull request a meaningful title (_e.g._ `Enable foobar integration` rather than `515 foobar`).
- [x] You are targeting the appropriate branch. If you're not certain which one this is, it should be **`develop`**.
- [x] Your branch is up-to-date with the **target branch** (it probably was when you started, but it may have changed since then).

### :vertical_traffic_light: Depends on

Depends on PR #2642.

### :arrow_heading_up: Summary

Upgrades the Nexus container to verson 3.94.1. In order to get the container to work the PostgreSQL Trigram Module must be enabled for the default public schema on the Nexus database instance.

### :closed_umbrella: Related issues

Fixes #2557.
Contributes to #2622 (Release 5.8.0).
PRs #2644 and #2624 build on these changes.

### :microscope: Tests

Combined with the changes in Pull Requests #2642, #2644 and #2624 I've successfully performed several deployments of SREs at Tier 2 and Tier 3. With the combined changes the unit tests all pass.

PR #2624, opened 2026-07-22
Release v5.8.0

llewelld:
Preparations for DSH release v5.8.0.

## :white_check_mark: Checklist

- [x] You have given your pull request a meaningful title (_e.g._ `Enable foobar integration` rather than `515 foobar`).
- [x] You are targeting the appropriate branch. If you're not certain which one this is, it should be **`develop`**.
- [x] Your branch is up-to-date with the **target branch** (it probably was when you started, but it may have changed since then).

### :vertical_traffic_light: Depends on

Depends on PRs #2642, #2643 and #2644 which should be merged first.

### :arrow_heading_up: Summary

This PR is for collecting together changes needed for release of version 5.8.0.

### :microscope: Tests

Combined with the changes in Pull Requests #2642, #2643 and #2644 I've successfully performed several deployments of SREs at Tier 2 and Tier 3. With the combined changes the unit tests all pass.

More details will be provided in the comments.

Issue #2622, opened 2026-07-20
Release: 5.8.0


llewelld:
## :white_check_mark: Checklist

Refer to the [Deployment](https://data-safe-haven.readthedocs.io/en/latest/deployment) section of our documentation when completing these steps.

- [x] Consult the `data-safe-haven/VERSIONING.md` guide and determine the version number of the new release. Record it in the title of this issue
- [x] Create a release branch called e.g. `release-v0.0.1`
    - If this is a hotfix release then this branch should be based off `latest`
    - In all other cases it should be based off `develop`
- [x] Draft a changelog for the release similar to our [previous releases](https://github.com/alan-turing-institute/data-safe-haven/releases)

### For patch releases only

~~- [ ] Confirm that the resource to which the patch applies can be successfully deployed~~

### For minor releases and above

- [x] Deploy an SHM from this branch and save a transcript of the deployment logs
- [x] Deploy a tier 2 SRE from this branch and save the transcript of the deployment logs
- [x] Deploy a tier 3 SRE from this branch and save the transcript of the deployment logs
- [x] Complete the [Security evaluation checklist](https://data-safe-haven.readthedocs.io/en/latest/deployment/security_checklist.html) from the deployment documentation

### For major releases only

~~- [ ] Confirm that a third party has carried out a full penetration test evaluating:
    1. external attack surface
    1. ability to exfiltrate data from the system
    1. ability to transfer data between SREs
    1. ability to escalate privileges on the SRD.~~

### Update documentation

- [x] Update supported versions in `SECURITY.md`
- [x] Update pen test results in `VERSIONING.md`

### Making the release

- [x] Merge release branch into `latest`
- [x] Create a tag of the form `v0.0.1` pointing to the most recent commit on `latest` (the merge that you just made)
- [x] Publish your draft GitHub release using this tag
- [x] Ensure docs for the latest version are built and deployed on ReadTheDocs
- [x] Push a build to PyPI
- [x] Announce release on communications channels
- [x] Create a PR from `latest` into `develop` to ensure that release-specific changes are not lost

## :deciduous_tree: Deployment problems

During deployment hit the following issues:
1. #2638
2. #2639
3. #2620
4. #2640

llewelld:
See #2560 for an example.

llewelld:
PR #2624 is for collecting together the changes needed for this release, on branch `release-v5.8.0`.

llewelld:
[...]

PR #2645, opened 2026-08-14
Updates to adjust to changes in the AzureSdk API

llewelld:

Nothing.

After updating the AzureSdk various changes have been needed to align with changes to the API:

1. The Subscription object in our version of the AzureSdk doesn't expose a tenant_id member, so this is now read from the active user account instead. We are using 3.1.1 of the Subscription module. See the link below.

2. The Permissions class constructor "keys" parameter now takes the name "keys_property". We are on version 14.0.1 and this change was made in version 13.0.0. See the link below.

The Subscription module Changelog:
https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/subscription/azure-mgmt-subscription/CHANGELOG.md#320b1-2022-12-27

The KeyVault module Changelog:
https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/keyvault/azure-mgmt-keyvault/CHANGELOG.md#1300-2025-12-11

### :mag: Detailed explanation of changes

Below is a file-by-file breakdown of the changes with explanations.

#### `data_safe_haven/commands/shm.py`

During testing I deployed an SHM and was a bit confused that there was no indication at the end that everything had been deployed successfully. I added a couple of new output lines which I though would be helpful for the user.

These changes aren't critical for this PR and can be removed if you feel they make things messy.

#### `data_safe_haven/external/api/azure_sdk.py`

The changes here and in `credentials.py` are the substantive changes of the PR. There are two things going on:

1. The `Permissions.keys` parameter is now called `Permissions.keys_property`. This change is detailed in the [Azure SDK ChangeLog](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/keyvault/azure-mgmt-keyvault/CHANGELOG.md#1300-2025-12-11) and has a simple fix: I just update the name.
2. The `Subscription` class no longer has a `tenant_id` property. This is [recorded in the AzureSdk ChangeLog](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/subscription/azure-mgmt-subscription/CHANGELOG.md#320b1-2022-12-27) has having been added in a version that we're not yet using, so practically speaking I'm not really sure why this has happened. Nevertheless, we can't use the `tenant_id` property with our version of the SDK. The change I've implemented to work around this is to get the tenant ID from the user's credential instead. This doesn't feel wrong to me, but you may feel differently.

#### `data_safe_haven/external/api/credentials.py`

We have our own `AzureSdkCredential` wrapper around the `TokenCredential` class from the AzureSdk, so some changes are needed here to expose the `tenant_id` property. The class hierarchy makes this a little tricky, but to ensure backwards compatibility I've turned the `tenant_id` member variable into a `property` which gets automatically populated in the background.

In the previous implementation `tenant_id` was never actually being populated. It now gets populated automatically when it's first queried.

I accept that the logic is a little convoluted here. My main aim was to ensure backwards compatibility, because although I don't think we were using the `tenant_id` member variable, I didn't want to risk making the assumption.

#### `data_safe_haven/external/api/graph_api.py`

I experienced some errors when using this method, added some debugging output and felt it might be helpful to keep it in. It's not critical for this PR, so I can remove it if you feel it's unhelpful.

#### `tests/external/api/test_azure_sdk.py`

Since the tenant ID is now being handled by the credentials rather than the subscription, the tenant Id test needs updating to use a different mock.

#### `tests/conftest.py`

There are two change here.

I've removed `tenant_id` from the `Subscription` mock so that it reflects the AzureSdk API.

I've also updated the `Credential` mock so that it handles the `tenant_id` property correctly.

The first change isn't necessary, but I think it makes sense. The second is necessary in order to get the tests to pass, but I think it's fair to say that it's not a cludge to get the tests through: it's a valid reflection of the AzureSdk API.

This contributes to issue #2622 (Release 5.8.0).
PRs #2643, #2644 and #2624 build on these changes.

### :microscope: Tests

Combined with the changes in Pull Requests #2643, #2644 and #2624 I've successfully performed several deployments of SREs at Tier 2 and Tier 3. With the combined changes the unit tests all pass.

### DIFF
--- a/data_safe_haven/commands/shm.py
+++ b/data_safe_haven/commands/shm.py
@@ -121,7 +121,9 @@ def deploy(
         logger.critical(msg)
         raise typer.Exit(1) from exc
     # Upload config file to blob storage
+    logger.info("Uploading config files")
     config.upload(context)
+    logger.info(f"Successfully deployed SHM {fqdn}")
 
 
 @shm_command_group.command()
--- a/data_safe_haven/external/api/azure_sdk.py
+++ b/data_safe_haven/external/api/azure_sdk.py
@@ -101,9 +101,7 @@ class AzureSdk:
     @property
     def tenant_id(self) -> str:
         if not self.tenant_id_:
-            self.tenant_id_ = str(
-                self.get_subscription(self.subscription_name).tenant_id
-            )
+            self.tenant_id_ = str(self.credential().tenant_id)
         return self.tenant_id_
 
     def blob_client_(
@@ -434,7 +432,7 @@ class AzureSdk:
                                 tenant_id=tenant_id,
                                 object_id=admin_group_id,
                                 permissions=Permissions(
-                                    keys=[
+                                    keys_property=[
                                         "GET",
                                         "LIST",
                                         "CREATE",
@@ -508,6 +506,7 @@ class AzureSdk:
             )
             return key
         except AzureError as exc:
+            self.logger.error(f"Error: {exc}")
             msg = f"Failed to create key '{key_name}' in KeyVault '{key_vault_name}'."
             raise DataSafeHavenAzureError(msg) from exc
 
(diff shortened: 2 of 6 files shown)


cptanalatriste (approved):
Looks good to me!

PR #2642, opened 2026-08-14
Updates to adjust to changes in the AzureSdk API

llewelld:
## :white_check_mark: Checklist

- [x] You have given your pull request a meaningful title (_e.g._ `Enable foobar integration` rather than `515 foobar`).
- [x] You are targeting the appropriate branch. If you're not certain which one this is, it should be **`develop`**.
- [x] Your branch is up-to-date with the **target branch** (it probably was when you started, but it may have changed since then).

### :vertical_traffic_light: Depends on

Nothing.

### :arrow_heading_up: Summary

After updating the AzureSdk various changes have been needed to align with changes to the API:

1. The Subscription object in our version of the AzureSdk doesn't expose a tenant_id member, so this is now read from the active user account instead. We are using 3.1.1 of the Subscription module. See the link below.

2. The Permissions class constructor "keys" parameter now takes the name "keys_property". We are on version 14.0.1 and this change was made in version 13.0.0. See the link below.

The Subscription module Changelog:
https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/subscription/azure-mgmt-subscription/CHANGELOG.md#320b1-2022-12-27

The KeyVault module Changelog:
https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/keyvault/azure-mgmt-keyvault/CHANGELOG.md#1300-2025-12-11

### :closed_umbrella: Related issues

This contributes to issue #2622 (Release 5.8.0).
PRs #2643, #2644 and #2624 build on these changes.

### :microscope: Tests

Combined with the changes in Pull Requests #2643, #2644 and #2624 I've successfully performed several deployments of SREs at Tier 2 and Tier 3. With the combined changes the unit tests all pass.

PR #2648, opened 2026-08-14
Bump version to 5.8.0

llewelld:
Preparations for DSH release v5.8.0.

Depends on PRs #2642, #2643 and #2644 which should be merged first.

This PR is for collecting together changes needed for release of version 5.8.0.

### :mag: Detailed explanation of changes

Below is a file-by-file breakdown of the changes with explanations.

#### `data_safe_haven/upgrade/upgrade.py`

There are a few minor changes here:
1. Some changes to the wording presented to the user during the upgrade for accuracy.
2. Previously I'd planned for this to be version 5.7.2, but at some point decided 5.8.0 was more appropriate. The numbers have been altered to address this.
3. There was a bug that meant the upgrade would be performed every time in the future, which is a mistake: it should only be performed if transitioning over the 5.8.0 boundary. The logic has been amended to fix this.

#### `data_safe_haven/version.py`

This bumps up the version number. There was never a 5.8.2 release; as I mentioned above, I'd originally planned for this to be 5.7.2 but since changed my mind.

#### `tests/upgrade/test_upgrade.py`

The tests needed updating to reflect changes following from the new version number. I also added to the tests to ensure the new logic (only upgrade if transitioning over the boundary) is adhered to.

#### `SECURITY.md`

Only the latest version is officially supported.

#### `VERSIONING.md`

Some info about previous DSGs was missing, so I've added it in.

### :microscope: Tests

Combined with the changes in Pull Requests #2642, #2643 and #2644 I've successfully performed several deployments of SREs at Tier 2 and Tier 3. With the combined changes the unit tests all pass.

I ran through the security checklist for both Tier 2 and Tier 3 SREs. I had to apply the nexus fix to get them to pass, but having done so everything worked as expected. The results can be seen on the Release Issue #2622.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -7,8 +7,8 @@ All organisations using an earlier version in production should update to the la
 
 | Version                                                                               | Supported          |
 | ------------------------------------------------------------------------------------- | ------------------ |
-| [5.7.1](https://github.com/alan-turing-institute/data-safe-haven/releases/tag/v5.7.1) | :white_check_mark: |
-| < 5.7.1                                                                               | :x:                |
+| [5.8.0](https://github.com/alan-turing-institute/data-safe-haven/releases/tag/v5.8.0) | :white_check_mark: |
+| < 5.8.0                                                                               | :x:                |
 
 ## Reporting a Vulnerability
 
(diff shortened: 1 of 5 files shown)


cptanalatriste (approved):
Looks good to me!

PR #2657, opened 2026-08-21
Pass api_version as named rather than positional argument

llewelld:

No dependencies.

However, this should be part of the 5.8.0 release, so I'd be inclined to re-write the history of `develop` so that this commit ends up earlier than the "Bump version to 5.8.0" commit 3aa09bdfed09fe. I'd be interested in opinions on this.

The Python AzureSdk API changed in version 26.0.0 so that various ResourcesOperations calls change the api_version argument from being positional or keyword to keyword only.

Since we were calling ResourcesOperations.get_id() and ResourcesOperations.begin_delete_by_id() with it as a positional argument, this has needed to be changed now that we're using 26.0.0 so it's passed as a named argument instead.

See the [ResourceOperations API is documented](https://learn.microsoft.com/en-us/python/api/azure-mgmt-resource/azure.mgmt.resource.resources.operations.resourcesoperations?view=azure-python) for info about the specific calls and their current interface.

See the following from the azure-mgmt-resource_26.0.0 ChangeLog:

> Method ResourcesOperations.begin_create_or_update/
> begin_create_or_update_by_id/ begin_delete/begin_delete_by_id/ begin_update/
> begin_update_by_id/ check_existence/ check_existence_by_id/ get/ get_by_id
> changed its parameter api_version from positional_or_keyword to keyword_only

https://github.com/Azure/azure-sdk-for-python/releases/tag/azure-mgmt-resource_26.0.0

## Detailed notes

File-by-file breakdown of changes.

### `data_safe_haven/external/api/azure_sdk.py`

This is the material change. The calls to `get_by_id()` and `begin_delete_by_id()` have been amended so that the `api-version` argument is passed as a named argument.

The only other changes here involve reformatting to appease the linter.

### `tests/conftest.py`

The mock resource manager client also has to be updated to accept `api_version` as a named argument. The change here both adds it as a possibility and makes it a requirement.

### :microscope: Tests

The issue was triggered during an upgrade from 5.7.1 to 5.8.0. After making the change I've performed a full upgrade which passed successfully.

### DIFF
--- a/data_safe_haven/external/api/azure_sdk.py
+++ b/data_safe_haven/external/api/azure_sdk.py
@@ -583,7 +583,9 @@ class AzureSdk:
             self.credential(), self.subscription_id
         ) as resource_client:
             azure_id = f"/subscriptions/{self.subscription_id}/resourceGroups/{resource_group_name}/providers/{provider_namespace}/{resource_type}/{resource_name}"
-            resource = resource_client.resources.get_by_id(azure_id, "2026-01-01")
+            resource = resource_client.resources.get_by_id(
+                azure_id, api_version="2026-01-01"
+            )
 
         return resource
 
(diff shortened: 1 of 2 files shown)


cptanalatriste (approved):
Looks good to me! And do not think it's mandatory to re-write the Git history. We can be honest about how we found it out later.

Other option is to add this changes directly to the branch's release, that it always get's merged back into develop (doing fixes during the release process is somewhat common in DSH)

llewelld:
Thanks for the review and for explaining about the history. I'm going to merge straight into `devel` since I think it makes little difference at this stage, but will target my changes at a release branch in future.

Issue #2585, opened 2026-03-18
Log analytics are 15% of the TRE operation costs


cptanalatriste:
## :white_check_mark: Checklist

- [x] I have searched open and closed issues for duplicates.
- [x] This is a request for a new feature in the Data Safe Haven or an upgrade to an existing feature.
- [x] The feature is still missing in the [latest version](https://github.com/alan-turing-institute/data-safe-haven/releases).
- [x] I have read through the [documentation](https://alan-turing-institute.github.io/data-safe-haven/).
- [x] This isn't an open-ended question (open a [discussion](https://github.com/alan-turing-institute/data-safe-haven/discussions) if it is).

## :strawberry: Suggested change

Log analytics can represent 15% (or more) of the cost of operating a DSH TRE:

<img width="1135" height="543" alt="Image" src="https://github.com/user-attachments/assets/07b139f6-3d1e-4872-b410-85015b675f5a" />

The log analytics costs are in purple. For example, in a day of operations that cost 90GBP, 14GBP are for log analytics. We believe this can be improved.

## :steam_locomotive: How could this be done?

There are many [Azure Monitor configuration options that can reduce the costs](https://learn.microsoft.com/en-us/azure/azure-monitor/fundamentals/best-practices-cost). Maybe we can expose those options in the configuration file, so Admins can tailor them accoding to their needs.

llewelld:
According to the [linked document ](https://learn.microsoft.com/en-us/azure/azure-monitor/fundamentals/best-practices-cost), Log Analytics costs can come from one of four services:
1. Azure Monitor
2. Log Analytics
3. Insight and Analytics
4. Application Insights

Downloading the costs through the Azure Cost Management interface as suggested shows that in our case, all of the costs fall under the Log Analytics service:

Date | Charge | Service Name | Meter | Cost | Currency
:- | :- | :- | :- | -: | :-
2026-05-01 | Usage | Log Analytics | Analytics Logs Data Ingestion | 8.87017227162598 | GBP
2026-05-02 | Usage | Log Analytics | Analytics Logs Data Ingestion | 13.2375896956989 | GBP
2026-05-03 | Usage | Log Analytics | Analytics Logs Data Ingestion | 13.249600158856 | GBP
2026-05-04 | Usage | Log Analytics | Analytics Logs Data Ingestion | 13.2664094746934 | GBP
[...]

PR #2666, opened 2026-09-17
Allow monitoring and logging to be configured

llewelld:

N/A.

Adds three new SRE configuration options:
1. monitoring.log_level
2. monitoring.retention_period
3. monitoring.sampling_interval

Controlling log and monitoring levels helps balance the verbosity and cost of logs with the level of granularity required for safe operation.

#2585


sylvesterkaczmarek (commented):
The new numeric monitoring settings are plain ints, so values like `retention_period: 0` or `sampling_interval: -1` pass config validation and are handed straight to the Azure workspace/perf-counter fields. That turns a config typo into a deployment-time failure. Could these use `PositiveInt` (plus any Azure-specific bounds) and get 0/negative config tests?

sylvesterkaczmarek (commented):
PositiveInt fixes the zero/negative case I raised, thanks. There is still one validation gap on the new retention cap: ensure_retention_period is only decorated with @classmethod, not @field_validator("retention_period"), so the 730-day limit is currently dead code and values such as 731 still pass model validation. Could you register that validator and add a >730 regression? The lower-bound concern is otherwise resolved.

llewelld:
> PositiveInt fixes the zero/negative case I raised, thanks. There is still one validation gap on the new retention cap: ensure_retention_period is only decorated with https://github.com/classmethod, not @field_validator("retention_period")...

Thank you @sylvesterkaczmarek for your suggestions: generally speaking your input is valuable and appreciated. Adding validation during configuration upload was a good call and makes for an improved user experience. I also appreciate the numerous valuable PR contributions you've been making. I apologise that I've not had time to look at many of them yet.

However, perhaps this is just my working style, but as a general rule I'd prefer for suggestions to be deferred to the review process (unless there's a problem with the entire premise of a change that needs discussion). Otherwise it can feel like backseat driving, which I find disruptive. This is especially the case while a PR is marked as being in Draft.

I recently moved PR #2669 out of draft and if you have comments there, I'd be very happy to receive them. Please don't take this the wrong way and I hope you'll appreciate me being upfront about it: it's just a matter of working style preference and clarifying the signals coming off, for example, my use of the "Draft" label.

sylvesterkaczmarek:
Understood, and thanks for being explicit about the working preference. I’ll treat Draft PRs as hands-off for review comments unless you ask for input or there is a premise-level issue, and wait until they are marked ready for review otherwise.

llewelld:
Thanks! I appreciate your understanding.

